PR #5, opened 2016-11-22
Create Test Infrastructure

### LINKED ISSUE #4: Allow Tests as in JUnit Jupiter
In JUnit Jupiter many tests extend from [AbstractJupiterTestEngineTests](https://github.com/junit-team/junit5/blob/master/junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/AbstractJupiterTestEngineTests.java). To allow tests with the same rigor, the same approach should work in Io.

nicolaiparlog:
This PR includes AssertJ and Mockito, draws in classes from JUnit 5 so tests can extend `AbstractJupiterTestEngineTests` and finally uses the Gradle plugin to run tests.

Fixes #4.

Regarding `AbstractJupiterTestEngineTests`: Unfortunately, JUnit 5 does currently not release test artifacts, (see junit-team/junit5#572), which means that the classes contained in this commit are only available

1. in source form from the JUnit code base
2. in an artifact if said code base is built

Building JUnit as part of Io would add much complexity and overhead. It would technically be possible to use Git to include a part of JUnit's source tree here (by an amalgam of submodules, sparse checkouts, and symlinks) but that would also add much complexity.

Lacking more ideas, I opted for the simplest approach and just copy-pasted the classes into Io.

---

### DIFF
--- a/src/test/java/org/junit/jupiter/engine/AbstractJupiterTestEngineTests.java
+++ b/src/test/java/org/junit/jupiter/engine/AbstractJupiterTestEngineTests.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2015-2016 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v1.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+
+package org.junit.jupiter.engine;
+
+import static org.junit.platform.engine.discovery.DiscoverySelectors.selectClass;
+import static org.junit.platform.launcher.core.LauncherDiscoveryRequestBuilder.request;
+
+import org.junit.platform.engine.ExecutionRequest;
+import org.junit.platform.engine.TestDescriptor;
+import org.junit.platform.engine.UniqueId;
+import org.junit.platform.engine.test.event.ExecutionEventRecorder;
+import org.junit.platform.launcher.LauncherDiscoveryRequest;
+
+/**
+ * Abstract base class for tests involving the {@link JupiterTestEngine}.
+ *
+ * @since 5.0
+ */
+public abstract class AbstractJupiterTestEngineTests {
+
+	private final JupiterTestEngine engine = new JupiterTestEngine();
+
+	protected ExecutionEventRecorder executeTestsForClass(Class<?> testClass) {
+		return executeTests(request().selectors(selectClass(testClass)).build());
+	}
+
+	protected ExecutionEventRecorder executeTests(LauncherDiscoveryRequest request) {
+		TestDescriptor testDescriptor = discoverTests(request);
+		ExecutionEventRecorder eventRecorder = new ExecutionEventRecorder();
+		engine.execute(new ExecutionRequest(testDescriptor, eventRecorder, request.getConfigurationParameters()));
+		return eventRecorder;
+	}
+
+	protected TestDescriptor discoverTests(LauncherDiscoveryRequest request) {
+		return engine.discover(request, UniqueId.forEngine(engine.getId()));
+	}
+
+}
--- a/src/test/java/org/junit/platform/engine/test/TestDescriptorStub.java
+++ b/src/test/java/org/junit/platform/engine/test/TestDescriptorStub.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2015-2016 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v1.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+
+package org.junit.platform.engine.test;
+
+import org.junit.platform.engine.UniqueId;
+import org.junit.platform.engine.support.descriptor.AbstractTestDescriptor;
+
+/**
+ * @since 1.0
+ */
+public class TestDescriptorStub extends AbstractTestDescriptor {
+
+	public TestDescriptorStub(UniqueId uniqueId, String displayName) {
+		super(uniqueId, displayName);
+	}
+
+	@Override
+	public boolean isTest() {
+		return getChildren().isEmpty();
+	}
+
+	@Override
+	public boolean isContainer() {
+		return !isTest();
+	}
+
+}
--- a/src/test/java/org/junit/platform/engine/test/TestEngineSpy.java
+++ b/src/test/java/org/junit/platform/engine/test/TestEngineSpy.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2015-2016 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v1.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+
+package org.junit.platform.engine.test;
+
+import org.junit.platform.engine.EngineDiscoveryRequest;
+import org.junit.platform.engine.ExecutionRequest;
+import org.junit.platform.engine.TestDescriptor;
+import org.junit.platform.engine.TestEngine;
+import org.junit.platform.engine.UniqueId;
+
+/**
+ * @since 1.0
+ */
+public class TestEngineSpy implements TestEngine {
+
+	private static final String ID = TestEngineSpy.class.getSimpleName();
+
+	public EngineDiscoveryRequest discoveryRequestForDiscovery;
+	public UniqueId uniqueIdForDiscovery;
+	public ExecutionRequest requestForExecution;
+
+	@Override
+	public String getId() {
+		return ID;
+	}
+
+	@Override
+	public TestDescriptor discover(EngineDiscoveryRequest discoveryRequest, UniqueId uniqueId) {
+		this.discoveryRequestForDiscovery = discoveryRequest;
+		this.uniqueIdForDiscovery = uniqueId;
+
+		UniqueId engineUniqueId = UniqueId.forEngine(ID);
+		TestDescriptorStub engineDescriptor = new TestDescriptorStub(engineUniqueId, ID);
+		TestDescriptorStub testDescriptor = new TestDescriptorStub(engineUniqueId.append("test", "test"), "test");
+		engineDescriptor.addChild(testDescriptor);
+		return engineDescriptor;
+	}
+
+	@Override
+	public void execute(ExecutionRequest request) {
+		this.requestForExecution = request;
+	}
+}
--- a/src/test/java/org/junit/platform/engine/test/TestEngineStub.java
+++ b/src/test/java/org/junit/platform/engine/test/TestEngineStub.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2015-2016 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v1.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+
+package org.junit.platform.engine.test;
+
+import org.junit.platform.engine.EngineDiscoveryRequest;
+import org.junit.platform.engine.ExecutionRequest;
+import org.junit.platform.engine.TestDescriptor;
+import org.junit.platform.engine.TestEngine;
+import org.junit.platform.engine.UniqueId;
+
+/**
+ * @since 1.0
+ */
+public class TestEngineStub implements TestEngine {
+
+	private final String id;
+
+	public TestEngineStub() {
+		this(TestEngineStub.class.getSimpleName());
+	}
+
+	public TestEngineStub(String id) {
+		this.id = id;
+	}
+
+	@Override
+	public String getId() {
+		return this.id;
+	}
+
+	@Override
+	public TestDescriptor discover(EngineDiscoveryRequest discoveryRequest, UniqueId uniqueId) {
+		return new TestDescriptorStub(UniqueId.forEngine(getId()), getId());
+	}
+
+	@Override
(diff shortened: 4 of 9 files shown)


src/test/java/org/junit/platform/engine/test/event/ExecutionEventRecorder.java
```diff
@@ -0,0 +1,175 @@
...
+	public static List<ExecutionEvent> execute(TestEngine testEngine, EngineDiscoveryRequest discoveryRequest) {
+		TestDescriptor engineTestDescriptor = testEngine.discover(discoveryRequest,
+			UniqueId.forEngine(testEngine.getId()));
+		ExecutionEventRecorder listener = new ExecutionEventRecorder();
+		testEngine.execute(
+			new ExecutionRequest(engineTestDescriptor, listener, discoveryRequest.getConfigurationParameters()));
+		return listener.getExecutionEvents();
+	}
+
+	public final List<ExecutionEvent> executionEvents = new CopyOnWriteArrayList<>();
```
jbduncan: Have you considered using a functional persistent list here, like PCollections' `TreePVector` or Javaslang's `javaslang.collection.List`?

src/test/java/org/junit/platform/engine/test/event/ExecutionEventRecorder.java
```diff
@@ -0,0 +1,175 @@
...
+	public long getContainerStartedCount() {
+		return containerEventsByType(STARTED).count();
+	}
+
+	public long getContainerFinishedCount() {
+		return containerEventsByType(FINISHED).count();
+	}
+
+	public List<ExecutionEvent> getSkippedTestEvents() {
+		return testEventsByType(Type.SKIPPED).collect(toList());
```
jbduncan: It's not clear to me here if you want the list returned here to be mutable or immutable (the javadoc for `toList()` makes no guarantees on this, so the mutability of the list may change from JVM to JVM).

I'd strongly suggest changing it to either `.collect(toCollection(ArrayList::new))` or `.collect(collectingAndThen(toList(), Collections::unmodifiableList))`.

src/test/java/org/junit/platform/engine/test/event/ExecutionEventRecorder.java
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	public List<ExecutionEvent> getSuccessfulTestFinishedEvents() {
+		return testFinishedEvents(Status.SUCCESSFUL).collect(toList());
```
jbduncan: Ditto.

src/test/java/org/junit/platform/engine/test/event/ExecutionEventRecorder.java
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	public List<ExecutionEvent> getFailedTestFinishedEvents() {
+		return testFinishedEvents(Status.FAILED).collect(toList());
```
jbduncan: Ditto.

src/test/java/org/junit/platform/engine/test/event/ExecutionEventRecorder.java
```diff
@@ -0,0 +1,175 @@
...
+	private Stream<ExecutionEvent> eventsByTypeAndTestDescriptor(Type type,
+			Predicate<? super TestDescriptor> predicate) {
+		return eventStream().filter(byType(type).and(byTestDescriptor(predicate)));
+	}
+
+	private void addEvent(ExecutionEvent event) {
+		executionEvents.add(event);
+	}
+
+}
```
jbduncan: I think it would be worth making the constructor `private` and allowing instances to be created with `ExecutionEventRecorder.create()`.

src/test/java/org/junit/platform/engine/test/event/ExecutionEventRecorder.java
```diff
@@ -0,0 +1,175 @@
...
+
+	public List<ExecutionEvent> getExecutionEvents() {
+		return executionEvents;
+	}
+
+	public Stream<ExecutionEvent> eventStream() {
+		return getExecutionEvents().stream();
+	}
+
+	public long getTestSkippedCount() {
```
jbduncan: The underlying stream is created from `executionEvents`, which being a `List` cannot have more than `Integer.MAX_VALUE` elements.

Have you considered casting the result of `.count()` to an `int` and making this method and all the other `long`-returning methods below return `int` instead?

src/test/java/org/junit/platform/engine/test/TestDescriptorStub.java
```diff
@@ -0,0 +1,35 @@
...
+
+import org.junit.platform.engine.UniqueId;
+import org.junit.platform.engine.support.descriptor.AbstractTestDescriptor;
+
+/**
+ * @since 1.0
+ */
+public class TestDescriptorStub extends AbstractTestDescriptor {
+
+	public TestDescriptorStub(UniqueId uniqueId, String displayName) {
```
jbduncan: Does the super constructor already check the nullness of these parameters?

If not, have you considered using `Objects.requireNonNull` or Guava's `Preconditions.checkNotNull`?

src/test/java/org/junit/platform/engine/test/event/ExecutionEventRecorder.java
jbduncan: Have you considered checking the nullness of the parameters of all the methods in this class, using `Objects.requireNonNull` or Guava's `Preconditions.checkNotNull`?

src/test/java/org/junit/platform/engine/test/event/TestExecutionResultConditions.java
```diff
@@ -0,0 +1,59 @@
...
+
+	}
+
+	public static Condition<TestExecutionResult> cause(Condition<? super Throwable> condition) {
+		return new Condition<TestExecutionResult>(where(TestExecutionResult::getThrowable, throwable -> {
+			return throwable.isPresent() && condition.matches(throwable.get());
+		}), "cause where %s", condition);
+	}
+
+}
```
jbduncan: Have you considered checking the nullness of the parameters of all the methods in this class, using `Objects.requireNonNull` or Guava's `Preconditions.checkNotNull`?

src/test/java/org/junit/jupiter/engine/AbstractJupiterTestEngineTests.java
```diff
@@ -0,0 +1,46 @@
...
+		ExecutionEventRecorder eventRecorder = new ExecutionEventRecorder();
+		engine.execute(new ExecutionRequest(testDescriptor, eventRecorder, request.getConfigurationParameters()));
+		return eventRecorder;
+	}
+
+	protected TestDescriptor discoverTests(LauncherDiscoveryRequest request) {
+		return engine.discover(request, UniqueId.forEngine(engine.getId()));
+	}
+
+}
```
jbduncan: Have you considered checking the nullness of the parameters of all the methods in this class, using `Objects.requireNonNull` or Guava's `Preconditions.checkNotNull`?

src/test/java/org/junit/platform/engine/test/TestEngineSpy.java
```diff
@@ -0,0 +1,51 @@
...
+		TestDescriptorStub testDescriptor = new TestDescriptorStub(engineUniqueId.append("test", "test"), "test");
+		engineDescriptor.addChild(testDescriptor);
+		return engineDescriptor;
+	}
+
+	@Override
+	public void execute(ExecutionRequest request) {
+		this.requestForExecution = request;
+	}
+}
```
jbduncan: Have you considered checking the nullness of the parameters of all the methods in this class, using `Objects.requireNonNull` or Guava's `Preconditions.checkNotNull`?

src/test/java/org/junit/platform/engine/test/TestEngineStub.java
```diff
@@ -0,0 +1,48 @@
...
+	@Override
+	public TestDescriptor discover(EngineDiscoveryRequest discoveryRequest, UniqueId uniqueId) {
+		return new TestDescriptorStub(UniqueId.forEngine(getId()), getId());
+	}
+
+	@Override
+	public void execute(ExecutionRequest request) {
+	}
+
+}
```
jbduncan: Have you considered checking the nullness of the parameters of all the methods in this class, using `Objects.requireNonNull` or Guava's `Preconditions.checkNotNull`?

src/test/java/org/junit/platform/engine/test/event/ExecutionEvent.java
```diff
@@ -0,0 +1,106 @@
...
+		// @formatter:off
+		return new ToStringBuilder(this)
+				.append("type", type)
+				.append("testDescriptor", testDescriptor)
+				.append("payload", payload)
+				.toString();
+		// @formatter:on
+	}
+
+}
```
jbduncan: Have you considered checking the nullness of the parameters of all the methods in this class, using `Objects.requireNonNull` or Guava's `Preconditions.checkNotNull`?

src/test/java/org/junit/platform/engine/test/event/ExecutionEventConditions.java
```diff
@@ -0,0 +1,156 @@
...
+		return new Condition<>(byPayload(TestExecutionResult.class, condition::matches), "event with result where %s",
+			condition);
+	}
+
+	public static Condition<ExecutionEvent> reason(String expectedReason) {
+		return new Condition<>(byPayload(String.class, isEqual(expectedReason)), "event with reason '%s'",
+			expectedReason);
+	}
+
+}
```
jbduncan: Have you considered checking the nullness of the parameters of all the methods in this class, using `Objects.requireNonNull` or Guava's `Preconditions.checkNotNull`?

src/test/java/org/junit/platform/engine/test/event/ExecutionEventConditions.java
```diff
@@ -0,0 +1,156 @@
...
+
+/**
+ * Collection of AssertJ conditions for {@link ExecutionEvent}.
+ *
+ * @since 1.0
+ */
+public class ExecutionEventConditions {
+
+	@SafeVarargs
+	public static void assertRecordedExecutionEventsContainsExactly(List<ExecutionEvent> executionEvents,
```
jbduncan: It's not clear to me if it makes sense for either `executionEvents` or `conditions` to be empty.

If they're not meant to be empty, I'd suggest throwing an `IllegalArgumentException` or similar if `executionsEvents.isEmpty()`, and changing

```
Condition<? super ExecutionEvent>... conditions
```

to

```
Condition<? super ExecutionEvent> firstCondition, Condition<? super ExecutionEvent>... rest
```

jbduncan: Does `executionEvents` need to be a `List`, or could it be more flexible and be an `Iterable`?

src/test/java/org/junit/platform/engine/test/event/ExecutionEventConditions.java
```diff
@@ -0,0 +1,156 @@
...
+		softly.assertThat(executionEvents).hasSize(conditions.length);
+		for (int i = 0; i < conditions.length; i++) {
+			softly.assertThat(executionEvents).has(conditions[i], atIndex(i));
+		}
+		softly.assertAll();
+	}
+
+	@SafeVarargs
+	@SuppressWarnings("varargs")
+	public static Condition<ExecutionEvent> event(Condition<? super ExecutionEvent>... conditions) {
```
jbduncan: Does it make sense for `conditions` to be empty?

nicolaiparlog:
Wow, @jbduncan! I had no idea this would attract any attention and so much work. I'm flattered as well as impressed. 😳 But except for the comment about the build file (true that!), I'm afraid I will not act on your recommendations.

I simply copy-pasted this code from JUnit 5 - take a look at [a07b65f](https://github.com/CodeFX-org/junit-io/pull/5/commits/a07b65f221f56375aaad95587a154bad878f5435) for the reasons behind that decision. I am currently testing out whether the classes work reliably (looks like it) and will continue to use and update them until they are published as part of some test artifact (see junit-team/junit5#572). Because of that I deliberately ignored their content and will not change them unless absolutely necessary.

But gleaming at your comments, they seem to be worth considering. If someone from the @junit-team would take a look and agree, why not make the changes you propose and open a PR over at JUnit 5?

Finally, I really appreciate the time you invested and have been following your comments over at the JUnit 5 repo. If you ever have an idea for an extension or any other feature that can be build from outside Jupiter and that would not be a good fit for the core project, please remember *Io* and open an issue or a PR. Thanks again! ✨

jbduncan:
You're very welcome @nicolaiparlog! I realise that I made many, many comments in my review, so thank you for taking the time to have a look, even if it was cursory, and for considering them.

Indeed, I'd be more than happy to open an issue or PR on the JUnit 5 repo if someone from the @junit-team were to have a look and think my comments are worth applying.

I'm honoured that you've been following my comments over at JUnit 5, so indeed if I ever have an idea for a JUnit 5 extension or something similar in the future, I'll consider making a proposal on JUnit Io.

Thanks and you're very welcome again. :sparkles:

nicolaiparlog:
I decided to play around with Git. That wasn't a good idea because it screwed up this PR.

@jbduncan: Now all those great comments are gone. 😭  Well, not gone but hidden behind _Show outdated:...

jbduncan:
Ah okay, I've only just realised this PR is not only copy-pasted code from JUnit 5 but test code as well, so I'm starting to think most of my comments are probably not applicable. :stuck_out_tongue_closed_eyes:

sbrannen:
@jbduncan, yep.... this is all internal testing code from JUnit 5. 

Sooooo, sorry you spent so much time reviewing it. 😱 

In any case, I would hold off any kind of review until the JUnit team has actually put forth an effort to _shape_ the code for public consumption.

Regards,

Sam

nicolaiparlog:
What's up, Sonar bot? You bored, so you build 40+ months old PRs? :smile:

PR #8, opened 2016-11-25
Implement Vintage @Test

### LINKED ISSUE #7: Recreate JUnit 4's @Test
In case JUnit 5 users would like to migrate JUnit 4 tests to Jupiter they might be interested in using an `@Test` annotation that behaves as it did in JUnit 4. This means there must be `expected` and `timeout` arguments with the same behavior.

nicolaiparlog:
Create an annotation `@Test` that has parameters `expected` and `timeout`. The former does what is expected (excuse the pun), the latter not exactly. Because Jupiter has no extension point to influence threading (yet?) the test can not actually be aborted if it runs too long. Instead, the test is failed when it ends and took too much time.

Fixes #7.

---

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,6 +15,7 @@ repositories {
 
 dependencies {
     compile group: 'org.junit.jupiter', name: 'junit-jupiter-api', version: '5.0.0-SNAPSHOT'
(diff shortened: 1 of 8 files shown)


nicolaiparlog:
This looks about done...

nicolaiparlog:
This looks about done.

PR #9, opened 2016-11-25
Implement @DisabledOnOs

### LINKED ISSUE #2: DisabledOnOs
Create annotations `@DisabledOnOs` and `@EnabledOnOs` that disable a container or test on the specified operating system or on all except the specified one, respectively.

Spawned from #1, also see junit-team/junit5#219, and this [proof of concept](https://github.com/CodeFX-org/demo-junit-5/blob/master/src/main/java/org/codefx/demo/junit5/OsCondition.java).

nicolaiparlog:
Implements annotations `@DisabledOnOs` and `@EnabledOnOs`, which pretty much do what the name says.

Fixes #2.

---

### DIFF
--- a/src/main/java/org/codefx/junit/io/extension/DisabledOnOs.java
+++ b/src/main/java/org/codefx/junit/io/extension/DisabledOnOs.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2015-2016 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v1.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+
+package org.codefx.junit.io.extension;
+
+import java.lang.annotation.ElementType;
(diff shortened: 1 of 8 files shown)


nicolaiparlog:
This is almost done, open issues:

* `OS` needs to be implemented properly
* there is only Javadoc so far but until a further documentation mechanism is decided upon, that suffices

src/main/java/org/codefx/junit/io/extension/DisabledOnOs.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+@Target({ ElementType.TYPE, ElementType.METHOD })
+@Retention(RetentionPolicy.RUNTIME)
+@ExtendWith(OsCondition.class)
+public @interface DisabledOnOs {
+
+	/**
+	 * Operating systems on which the test or test container is disabled.
+	 */
+	OS[] value() default {};
```
sbrannen: Don't provide a default value for an annotation attribute if its required.

In other words, declaring `@DisabledOnOs` without specifying and OS is pointless. 😉 

Same for `@EnabledOnOs`.

PR #45, opened 2017-05-12
Rename packages

### LINKED ISSUE #24: Rename packages to match JUnit Pioneer
Current package prefix is `org.codefx.junit.io`. The new prefix is not decided upon yet but the domain `junit-pioneer.org` belongs to the project and might be used.

nicolaiparlog:
Closes #24 by renaming the package from `org.codefx.junit.io` to `org.junit$pioneer`. (Yes, I chose `$` 😛.)

---


nicolaiparlog:
Proposed commit message:

```
Update package names to new project name (#24, #45)

With the project being renamed from "Junit Io" to "JUnit Pioneer" the
old top level package name `org.codefx.junit.io` had to be changed.
Because the project owns the domain junit-pioneer.org, it was obvious
to use it as a package name according to the establishes reverse DNS
naming strategy. In that case the dash (-) has to be replaced with a
valid Java package identifier character. In a bold move (we're
pioneers, after all!) the dollar sign ($) was chosen. :)
```

nicolaiparlog:
Weird that the Spotless check fails only now...

jbduncan:
Weird, indeed...

I have no theories as to this behaviour's cause. (Bad Spotless!) :stuck_out_tongue_winking_eye:

nicolaiparlog:
@smoyer64 Since this is an important decision can you :+1: if you're ok with it?

nicolaiparlog:
D'oh, there are no notifications for Emoji reactions.

PR #47, opened 2017-05-21
Use Gradle properties

### LINKED ISSUE #46: Use gradle.properties for login credentials
Currently `build.gradle` defines the user's password for the Maven snapshot repository as follows:

```
ext {
    // the password needs to be specified via command line
    snapshotRepoPass = project.hasProperty('snapshotRepoPass') ? project.getProperty('snapshotRepoPass') : ''
}
```

This requires the user to enter the password on the command line, which commits it to CLI history. A more sensible way to configure this would be a user-local `gradle.properties` file [as explained here](https://docs.gradle.org/current/userguide/build_environment.html#sec:gradle_properties_and_system_properties).

The opportunity should be used to make the username configurable as well and additionally use a more sensible property name.

* [ ] configure `build.gradle` to use properties defined in `gradle.properties`
* [ ] update `.travisyml` and Travis configuration to use the new mechanism
* [ ] use the opportunity to document the publication process

nicolaiparlog:
Maven user name and password can now be defined as `gradle.properties` _or_ as command line properties.

Closes #46.

---

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ script:
 deploy:
   skip_cleanup: true
   provider: script
-  script: ./gradlew publish -P snapshotRepoPass=$SNAPSHOT_REPO_PASS --stacktrace
+  script: ./gradlew publish -P mavenUserName=$MAVEN_USER_NAME -P mavenPassword=$MAVEN_PASSWORD --stacktrace
   on:
     branch: master
     jdk: oraclejdk8
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -87,3 +87,24 @@ the detailed history, which will be more coarse than with merge commits
 or fast-forward merges. This was deemed acceptable in order to achieve
 the other points, particularly the last one.
 ```
+
+
+## Publishing
+
+### Snapshots
+
+To publish snapshots to Maven Central you need to execute `gradle publish` after defining the properties `mavenUserName` and `mavenPassword`.
+
+One way to do the latter are [Gradle properties](https://docs.gradle.org/current/userguide/build_environment.html#sec:gradle_properties_and_system_properties).
+For that approach, create a file `gradle.properties` in `GRADLE_USER_HOME` (which defaults to `USER_HOME/.gradle`) with the following content:
+
+```
+mavenUserName=...
+mavenPassword=...
+```
+
+Another way are command line flags (but note that these add sensitive information to your terminal history):
+
+```
+gradle publish -PmavenUserName=... -PmavenPassword=...
+```
--- a/README.md
+++ b/README.md
@@ -15,4 +15,24 @@ A melting pot for all kinds of extensions to
 ## Contributing
 
 We welcome contributions of all kinds and shapes!
-We are still building up infrastructure and what exactly we need help on is not easy to say, but we already settled and [what we want contributions to look like](CONTRIBUTING.md).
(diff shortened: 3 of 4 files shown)


nicolaiparlog:
@smoyer64 Can you give this branch a try to publish to Maven Central? Besides checking whether it works, you can also check whether the documentation suffices - see [updated `README`](https://github.com/junit-pioneer/junit-pioneer/pull/47/files#diff-04c6e90faac2675aa89e2176d2eec7d8R22) for that.

nicolaiparlog:
Argh, CI builds fail because the servers do not define the command line arguments, which means they also fail for anyone else who doesn't have. I have to fix that but you can still try out whether publication works.

mkobit:
You can use `project.findProperty('propName')` and it will be null for local builds and whatever was specified on the CI System when it runs there.

nicolaiparlog:
@mkobit Yes, I did that earlier for the command line flag but for some reason forgot I had to do it here, too. Just rediscovered it. :wink:

nicolaiparlog:
Oh, I did `hasProperty`/`getProperty` - `findProperty` is much better, thanks for the hint.

nicolaiparlog:
Proposed commit message:

```
Read Maven Central credentials from Gradle properties (#46 / #47)

To allow all maintainers to publish to Maven Central the credentials
are read from the Gradle properties file. This did not actually
require a technical change as the existing mechanism already did that
implicitly (and without my knowledge). New is that the user name as
well as the password is read from properties and that CONTRIBUTING.md
describes the process.
```

nicolaiparlog:
@smoyer64 Can you give this a try, so I can merge and continue working on release pipeline?

smoyer64:
@nicolaiparlog I tried to publish a both a snapshot and a release this morning.  When publishing the snapshot, it failed with the message:

```
 * What went wrong:
Execution failed for task ':publishCompletePublicationToMavenSnapshotRepository'.
> Failed to publish publication 'complete' to repository 'MavenSnapshot'
   > Failed to deploy artifacts: Could not transfer artifact org.junit-pioneer:junit-pioneer:jar:0.001-20170623.111606-1 from/to remote (https://oss.sonatype.org/content/repositories/snapshots/): Could not write to resource 'org/junit-pioneer/junit-pioneer/0.001-SNAPSHOT/junit-pioneer-0.001-20170623.111606-1.jar'
```

When trying to publish the release, it failed to sign the artifacts since I don't have the signing key on my machine.  Since we're "always" using branches and PRs, should we have the triggered CI build on master perform the releaseSnapshot task?

I've added you on Keybase if you need to share anything securely.

nicolaiparlog:
Now that everything's merged, can you have a look at the [docs on publishing](https://github.com/junit-pioneer/junit-pioneer/blob/master/CONTRIBUTING.md#publishing) and proceed as they describe? You should be able to publish snapshots and releases, using your credentials for each. You can use any PGP key want as long as it's published on a public key server. It may also have to be bound to the email address you registered on Sonatype (it would make sense and is the case for me but I don't know whether Sonatype checks).

If you have problems, can you run your gradle command with `--stacktrace` and open an issue for it?

smoyer64:
I was using the publishing instructions in CONTRIBUTING.md - it turns out that my problem was that ```mavenUsername != mavenUserName```.  I hadn't thought about publishing a release with my PGP key but that should also work fine.  When we publish artifacts at Penn State, we use a key that was created specifically for our CI server so that all our releases are signed with the same key.

PR #52, opened 2017-06-13
Release to Maven Central

### LINKED ISSUE #17: Enable Publication To Maven Central
Once the first release rolls around, `build.gradle` needs to be expanded so that it can sign artifacts and upload them to Maven Central.

nicolaiparlog:
First work on using Gradle and [maven-publish](https://docs.gradle.org/3.5/userguide/publishing_maven.html) to publish to Maven Central - closes #17. ToDos:

* technical:
    * [x] stage JAR on Maven central
    * [x] stage POM, sources, and JavaDoc
    * [x] sign all artifacts and stage signatures
* other:
    * [x] document release workflow
    * [x] write commit message
    * [x] create follow-up issue to further automate release ~> #53

---

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ script:
 deploy:
   skip_cleanup: true
   provider: script
-  script: ./gradlew publishCompletePublicationToMavenSnapshotRepository -P mavenUserName=$MAVEN_USER_NAME -P mavenPassword=$MAVEN_PASSWORD --stacktrace
+  script: ./gradlew publishSnapshot -P mavenUserName=$MAVEN_USER_NAME -P mavenPassword=$MAVEN_PASSWORD --stacktrace
   on:
     branch: master
     jdk: oraclejdk8
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ script:
 deploy:
   skip_cleanup: true
   provider: script
-  script: ./gradlew publish -P mavenUserName=$MAVEN_USER_NAME -P mavenPassword=$MAVEN_PASSWORD --stacktrace
+  script: ./gradlew publishCompletePublicationToMavenSnapshotRepository -P mavenUserName=$MAVEN_USER_NAME -P mavenPassword=$MAVEN_PASSWORD --stacktrace
   on:
     branch: master
     jdk: oraclejdk8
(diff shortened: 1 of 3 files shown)


nicolaiparlog:
Proposed commit message:

```
Enable publication to Maven Central (#17 / #52)

Since snapshot releases were already possible, "only" artifact
signing had to be activated for publications to Maven Central.
This turned out to be a little tricky with the maven-publish
plugin but it eventually worked. Some hurdles to overcome:

* the default way to configure the signing plugin
  (`sign configurations.archives`) does not work for the POM, so
  a specific task had to be created for that
* the POM-creating can not be used as direct input for the signing
  task, so instead the created file was referenced directly, which
  may make the configuration brittle
* the signing tasks are part of the "assemble" task but because
  they fail on machines without a signatory they had to be
  deactivated except when actually needed

The maven-publish plugin also needed some massaging:

* the "publish" task wants to execute all publications, so I
  deactivated it
* the individual snapshot and release tasks got ridiculously long
  names so I created shorter aliases
* I'm convinced publication should depend on "check"

This change really only covers the pure release process - preparatory
steps (see `CONTRIBUTING.md`) are still manual. Their automation
might be implemented in the future (see #53).
```

nicolaiparlog:
Ready to merged...

smoyer64:
I'm not sure if there's a better way of signing the POM - I've only published from Maven builds and I'm still a Gradle newb.  It would have taken me a lot longer to get this done!  In any case, if we find a better way in the future we can always put in another PR.

nicolaiparlog:
@smoyer64 Remember that [we want to squash and merge](https://github.com/junit-pioneer/junit-pioneer/blob/master/CONTRIBUTING.md#merging) with the commit message provided in the PR. I just did that manually and force-pushed to master.

smoyer64:
I see that ... sorry to make you rewrite history ;)

I think you can also ignore my comment about having the CI server push commits to master up to oss.sonatype.org (it seems to already be doing that).

nicolaiparlog:
> sorry to make you rewrite history

I'm not the one who has to trouble through a `git pull`, so I'm fine with that. :wink: It was just ~40 minutes so hopefully nobody has that trouble.

nicolaiparlog:
Hah, and I found a setting that removes the options to use merge commits or fast-rebase/forward to close branched. You should be safe from repeating that mistake. :smiley:

smoyer64:
I think I did better on the PR I just merged ;)

I also noticed that you were following the published branching strategy from forked branches.  The contributor's guide doesn't state a preference for forked branches over local branches but it makes sense since contributors won't have a choice.

PR #56, opened 2017-08-15
Document dependencies

### LINKED ISSUE #33: Decide on what dependencies we should allow
LIke the JUnit 5 project, I think limiting the dependencies in the core project is a worthwhile goal.  At this point I'd really only argue strongly for JSR-305 annotations (current GAV is com.google.code.findbugs:jsr305:3.0.1).

For tests and infrastructure, I don't tend to be as militant ... should we allow the following?

-   AssertJ
-   Mockito
-   JaCoCo

(I'm sure this list will grow as we do!)

nicolaiparlog:
Document dependencies. Closes #33.

---


nicolaiparlog:
Proposed commit message:

```
Document allowed dependencies (#33, #56)

JUnit Pioneer, like JUnit 5, strives for ease of use in all kinds of
projects, which would be complicated by taking on runtime
dependencies. That's why we don't do it.

Compile and test dependencies should also not go overboard, so we
settled on a set of required dependencies (for now).
```

nicolaiparlog:
Ignore the failing build, it has nothing to do with this PR - will fix it separately.

PR #57, opened 2017-08-15
Document use of Optional

### LINKED ISSUE #51: Create code style for Optional vs null
How should the project handle `null` references and which ones should be replaced by `Optional`? Possible distinctions are:

* method return values
* method parameters
* fields

Possible variants could be decided for:

* `public` vs `default` vs `private`
* `static` vs instance
* ... ?

nicolaiparlog:
Document use of `, closes #51.

---


nicolaiparlog:
Proposed commit message:

```
Document null-handling with Optional (#51/ #57)

JUnit Pioneer does not use `null` to encode legal values or state
unless that is required for calling into JUnit API. Instead,
maintainers aim to reduce the prevalence of optionality with good
design and, where no other options are left, use `Optional`. This
explicitly includes places, where use is generally discouraged, like
arguments and fields.

Read these posts for background on the general philosophy:

* https://blog.codefx.org/techniques/intention-revealing-code-java-8-optional/
* https://blog.codefx.org/java/stephen-colebourne-optional-a-strict-approach/
```

Also, ignore the damn build. 😃

PR #61, opened 2017-08-17
Add Automatic-Module-Name manifest entry

### LINKED ISSUE #60: Provide automatic-module-name in JAR manifest
How about `org.junit-pioneer` ... is minus `-` a valid character in module names?!

Like
```groovy
jar {
	manifest {
		attributes(
			'Automatic-Module-Name': 'org.junit$pioneer'
		)
	}
}
```

nicolaiparlog:
Closes #60 

---

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -104,6 +104,15 @@ check.dependsOn javadoc
 // it should ran last, though
 javadoc.shouldRunAfter test
 
+// the manifest needs to declare the future module name
+jar {
+    manifest {
+        attributes(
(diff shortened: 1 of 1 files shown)


nicolaiparlog:
Proposed commit message:

```
Define automatic module name in manifest (#60 / #61)

With Java 9 on the doorstep, JUnit Pioneer needs to be prepared to be
used as a module. Until a proper modularization includes a module
descriptor in the JAR, it is important to define its future module
name. This way, when used as an automatic module that other modules
depend on, releasing a modular JAR does not break those modules.
```

PR #63, opened 2017-09-21
Dependency updates

mkobit:
---


mkobit:
I didnt realize `$pioneer` was intentional from reading https://github.com/junit-pioneer/junit-pioneer/issues/60, it seems like an odd choice for the package namespace. I'm still not sure I understand the reasoning.

mkobit:
Just to throw it out there underscore is an option in the package namespace as well.

build.gradle
```diff
@@ -1,7 +1,32 @@
+buildscript {
```
nicolaiparlog: I don't like the reordering - I preferred to start the file with the definition of the project's group and version.

build.gradle
```diff
@@ -1,7 +1,32 @@
...
+    dependencies {
+        classpath group: 'com.diffplug.gradle.spotless', name: 'spotless', version: '2.4.1'
+        classpath group: 'org.junit.platform', name: 'junit-platform-gradle-plugin', version: '1.0.0'
+        classpath group: 'com.squareup.okhttp3', name: 'okhttp', version: '3.9.0'
+        classpath group: 'org.ajoberstar', name: 'grgit', version: '2.0.1'
+    }
+}
+
+plugins {
+    id 'com.github.ben-manes.versions' version '0.15.0'
```
nicolaiparlog: What do we need this for?

build.gradle
```diff
@@ -28,35 +53,14 @@ repositories {
 }
 
 dependencies {
-    compile group: 'org.junit.jupiter', name: 'junit-jupiter-api', version: '5.0.0-SNAPSHOT'
-    compile group: 'org.opentest4j', name: 'opentest4j', version: '1.0.0-M1'
-
-    testCompile group: 'org.junit.jupiter', name: 'junit-jupiter-engine', version: '5.0.0-SNAPSHOT'
-    testCompile group: 'org.junit.platform', name: 'junit-platform-launcher', version: '1.0.0-SNAPSHOT'
+    api group: 'org.junit.jupiter', name: 'junit-jupiter-api', version: '5.0.0'
```
nicolaiparlog: We've never discussed versions so far, but my goal was to continuously build against the latest snapshot and only pick a concrete version when creating a release branch. For now, can you please use the latest snapshots (5.1... I assume)?

build.gradle
```diff
@@ -28,35 +53,14 @@ repositories {
...
+    testImplementation group: 'org.assertj', name: 'assertj-core', version: '3.8.0'
+    testImplementation group: 'org.mockito', name: 'mockito-core', version: '2.10.0'
 
-    dependencies {
-        classpath group: 'com.diffplug.gradle.spotless', name: 'spotless', version: '2.2.0'
-        classpath group: 'org.junit.platform', name: 'junit-platform-gradle-plugin', version: '1.0.0-SNAPSHOT'
-        classpath group: 'com.squareup.okhttp3', name: 'okhttp', version: '3.8.1'
-        classpath group: 'org.ajoberstar', name: 'grgit', version: '1.9.3'
-    }
+    testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-engine', version: '5.0.0'
```
nicolaiparlog: Is it necessary to repeat `api` grouped dependencies as `testImplementation`? Looks redundant to me.

nicolaiparlog:
Regarding the module/package names, see #24 - TL;DR: yes, `junit$pioneer` is on purpose. 😄 Can you reverse that change?

build.gradle
mkobit: Can get rid of this is we would like - it can be used manually like `./gradlew dependencyUpdates` to get a report of all out-of-date dependencies

build.gradle
mkobit: I _believe_ that `buildscript` must come before `plugins {}` block because of how the script gets processed, but I can take another crack at it

build.gradle
mkobit: Might not be a bad idea to have the CI run builds against both 'latest' and static versions. I changed them to the release because it could be error prone if accidentally published and because when I pulled this project up on the train it tried to refresh those dependencies :neutral_face: 

I'll switch it back to the snapshot for this PR

build.gradle
mkobit: Shouldn't have to repeat them, was a mistake by me. I will fix it.

mkobit:
I'll have another go at this tonight.

build.gradle
nicolaiparlog: I don't know much about Gradle, but I'm under the impression it worked well before. :wink:

build.gradle
nicolaiparlog: Oh, that's actually pretty nice. Leave it in then, and I will give it a go when I build this branch locally.

build.gradle
nicolaiparlog: Just created #64 to discuss this.

> Might not be a bad idea to have the CI run builds against both 'latest' and static versions.

Yes, that's a good idea.

> it could be error prone if accidentally published

True, would be good to build something into the Gradle script that checks this.

> when I pulled this project up on the train it tried to refresh those dependencies

Didn't think about that, you're right it makes the project depend on being online - I actually like working in the train, so that's a serious downside.

build.gradle
mkobit: I just tested it with

```
group 'org.junit-pioneer'
version '0.1-SNAPSHOT'

buildscript {
    repositories {
        mavenLocal()
        maven {
            name "GradlePlugins"
            url 'https://plugins.gradle.org/m2/'
        }
        maven {
            name "MavenSnapshot"
            url 'https://oss.sonatype.org/content/repositories/snapshots/'
        }
    }

    dependencies {
        classpath group: 'com.diffplug.gradle.spotless', name: 'spotless', version: '2.4.1'
        classpath group: 'org.junit.platform', name: 'junit-platform-gradle-plugin', version: '1.0.0'
        classpath group: 'com.squareup.okhttp3', name: 'okhttp', version: '3.9.0'
        classpath group: 'org.ajoberstar', name: 'grgit', version: '2.0.1'
    }
}

plugins {
    id 'com.github.ben-manes.versions' version '0.15.0'
}
```

and it fails with 

```

* Where:                                             
Build file '/home/mkobit/Workspace/personal/junit-pioneer/junit-pioneer/build.gradle' line: 25            

* What went wrong:                                   
Could not compile build file '/home/mkobit/Workspace/personal/junit-pioneer/junit-pioneer/build.gradle'.  
> startup failed:                                    
  build file '/home/mkobit/Workspace/personal/junit-pioneer/junit-pioneer/build.gradle': 25: only buildscript {} and other plugins {} script blocks are allowed before plugins {} blocks, no other statements are allowed                                                
                                                     
  See https://docs.gradle.org/4.2/userguide/plugins.html#sec:plugins_block for information on the plugins {} block                                                                                                  
                                                     
   @ line 25, column 1.                              
     plugins {                                       
     ^                                               
                                                     
  1 error                                            


* Try:                                               
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.                                                                                                      

* Get more help at https://help.gradle.org   
```

build.gradle
mkobit: Just took another look at this:

`api` dependencies:

* `junit-jupiter-api
* `opentest4j`

Those make sense to me for limititing the transitive dependencies.

`testImplementation` dependencies

* `junit-jupiter-engine`
* `junit-platform-launcher`

Which I think are needed only in the tests because of usage of some pieces like https://github.com/junit-pioneer/junit-pioneer/tree/master/src/test/java/org/junit/platform/engine/test

mkobit:
I closed it as it is mostly stale now and haven't seen activity on here in a while.

jbduncan:
Fair enough, that makes sense to me. :)

Issue #64, opened 2017-09-26
Decide on which JUnit version to depend


nicolaiparlog:
So far, the project depends on JUnit 5's SNAPSHOT versions. This obviously has to be changed for releases, which begs the question whether we always want to use the latest version? Or depend on the minimum required to work? And what about during development - are snapshots the right choice there? Lots of questions... here are some thoughts:

Depending on snapshots during development...

* ... allows us to develop against unreleased features/APIs
* ... forces us to immediately react to upstream changes
* ... improves JUnit's stability by giving quick feedback from an outside system
* ... requires some work to figure out which version a Pioneer release depends on

Which version of JUnit should released Pioneer artifacts depend on?

* the latest, so we can do the most?
* only update to a newer version when required for important features?
* maybe we have several release lines with different target versions (sounds very complex for a project that doesn't even have any feature, let alone a release)

I have an idea how to go about this, but I'm interested to hear yours first.

nicolaiparlog:
Here are the ideas I alluded to over half a year ago.

## First try

The solution that I find the most compelling is this one:

* during development (i.e. `master` branch) depend on the most recent snapshots
* before each release, determine the minimum required JUnit 5 version and release against that

This gives us the chance to work with the latest and greatest and also gives JUnit 5 an additional round of feedback. There are side effects of course:

* Pioneer might be broken if JUnit 5 breaks ~> great, more feedback for 5!
* Pioneer might be broken until it adapts to breaking upstream changes ~> acceptable as that needs to be done at some point anyway
* new versions of Pioneer are very likely only usable with new versions of 5 ~> since 5 values backwards compatibility, this shouldn't be a big problem
* we might have to release a version (e.g. a bug fix) while we already have code depending on an unreleased version

That last one would hurt a lot because there's no really good fix. Unless we cut release branches. 

## Version proposal

Updated idea:

* `master` depends on the most recent snapshots
* each release line gets its own branch and when a new version gets to be released, we cherry-pick all features and fixes that are compatible with the JUnit version it depends on onto that branch

That begs the question how to cut release branches. This is a separate discussion, so when evaluating  the version proposal, don't conflate the two as all or nothing.

## Branching proposal

Here's my proposal for how to cut release branches:

* name minor releases so that they match the minor JUnit 5 version they depend on (e.g. `1.2.x` ~> `5.2.y`)
* create a branch for each such minor release line

[...]

PR #69, opened 2018-05-05
 Introduce TempDirectory Jupiter Extension

### LINKED ISSUE #39: Move TemporaryFolderExtension to pioneer

marcphilipp:
This PR introduces a Jupiter extension to create and cleanup per-test temporary directories. Resolves #39.

### Example

```java
@ExtendWith(TempDirectory.class)
class Tests {
	@Test
	void test(@TempDir Path tempDir) throws Exception {
		Path file = tempDir.resolve("my-file.txt");
		Files.write(file, "Hello World".getBytes());
	}
}
```

---

### DIFF
--- a/src/main/java/org/junit$pioneer/jupiter/TempDirectory.java
+++ b/src/main/java/org/junit$pioneer/jupiter/TempDirectory.java
@@ -24,9 +24,9 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.SimpleFileVisitor;
 import java.nio.file.attribute.BasicFileAttributes;
-import java.util.Optional;
 import java.util.concurrent.Callable;
 
+import org.junit.jupiter.api.extension.ExtensionConfigurationException;
 import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.jupiter.api.extension.ExtensionContext.Namespace;
 import org.junit.jupiter.api.extension.ExtensionContext.Store.CloseableResource;
@@ -79,6 +79,8 @@ import org.junit.jupiter.api.extension.ParameterResolver;
  */
 public class TempDirectory implements ParameterResolver {
 
+	private static final String TEMP_DIR_PREFIX = "junit";
+
 	/**
 	 * {@code TempDir} can be used to annotate a test or lifecycle method or
 	 * test class constructor parameter of type {@link Path} that should be
@@ -106,17 +108,35 @@ public class TempDirectory implements ParameterResolver {
 		 * Get the parent directory for all temporary directories created by the
 		 * {@link TempDirectory} extension this is used with.
 		 *
-		 * @return the {@link Optional optional} parent directory for all temporary
-		 * directories; must be {@link Optional#empty() empty} or contain an existing
-		 * directory.
+		 * @return the parent directory for all temporary directories
 		 */
-		Optional<Path> get(ParameterContext parameterContext, ExtensionContext extensionContext) throws Exception;
+		Path get(ParameterContext parameterContext, ExtensionContext extensionContext) throws Exception;
+	}
+
+	/**
+	 * {@code TempDirProvider} is used internally to define how the temporary
+	 * directory is created.
+	 *
+	 * <p>The temporary directory is by default created on the regular
+	 * file system, but the user can also provide a custom file system
+	 * by using the {@link ParentDirProvider}. An instance of
+	 * {@code TempDirProvider} executes these (and possibly other) strategies.
+	 *
+	 * @see TempDirectory.ParentDirProvider
+	 */
+	@FunctionalInterface
+	private interface TempDirProvider {
+		CloseablePath get(ParameterContext parameterContext, ExtensionContext extensionContext, String dirPrefix);
 	}
 
 	private static final Namespace NAMESPACE = Namespace.create(TempDirectory.class);
 	private static final String KEY = "temp.dir";
 
-	private final ParentDirProvider parentDirProvider;
+	private final TempDirProvider tempDirProvider;
+
+	public TempDirectory(TempDirProvider tempDirProvider) {
+		this.tempDirProvider = requireNonNull(tempDirProvider);
+	}
 
 	/**
 	 * Create a new {@code TempDirectory} extension that uses the default
@@ -128,27 +148,7 @@ public class TempDirectory implements ParameterResolver {
 	 * {@link org.junit.jupiter.api.extension.ExtendWith @ExtendWith}.
 	 */
 	public TempDirectory() {
-		this((parameterContext, extensionContext) -> Optional.empty());
-	}
-
-	/**
-	 * Create a new {@code TempDirectory} extension that uses the supplied
-	 * {@link Callable} to configure the parent directory for the temporary
-	 * directories created by this extension.
-	 *
-	 * <p>The {@link Optional optional} parent directory returned by the
-	 * supplied {@link Callable} must be {@link Optional#empty() empty} or
-	 * contain an existing directory.
-	 *
-	 * <p>You may use this constructor when registering this extension via
-	 * {@link org.junit.jupiter.api.extension.RegisterExtension @RegisterExtension}.
-	 *
-	 * @param parentDirProvider used to configure the parent directory for the
-	 * temporary directories created by this extension
-	 */
-	public TempDirectory(Callable<Optional<Path>> parentDirProvider) {
-		this((parameterContext, extensionContext) -> parentDirProvider.call());
-		requireNonNull(parentDirProvider);
+		this((__, ___, dirPrefix) -> createDefaultTempDir(dirPrefix));
 	}
 
 	/**
@@ -163,7 +163,27 @@ public class TempDirectory implements ParameterResolver {
 	 * temporary directories created by this extension
 	 */
 	public TempDirectory(ParentDirProvider parentDirProvider) {
-		this.parentDirProvider = requireNonNull(parentDirProvider);
+		// @formatter:off
+		this((parameterContext, extensionContext, dirPrefix) ->
+				createCustomTempDir(parentDirProvider, parameterContext, extensionContext, dirPrefix));
+		// @formatter:on
+		requireNonNull(parentDirProvider);
+	}
+
+	/**
+	 * Create a new {@code TempDirectory} extension that uses the supplied
+	 * {@link Callable} to configure the parent directory for the temporary
+	 * directories created by this extension.
+	 *
+	 * <p>You may use this constructor when registering this extension via
+	 * {@link org.junit.jupiter.api.extension.RegisterExtension @RegisterExtension}.
+	 *
+	 * @param parentDirProvider used to configure the parent directory for the
+	 * temporary directories created by this extension
+	 */
+	public TempDirectory(Callable<Path> parentDirProvider) {
+		this((parameterContext, extensionContext) -> parentDirProvider.call());
+		requireNonNull(parentDirProvider);
 	}
 
 	@Override
@@ -179,34 +199,35 @@ public class TempDirectory implements ParameterResolver {
 				"Can only resolve parameter of type " + Path.class.getName() + " but was: " + parameterType.getName());
 		}
 		return extensionContext.getStore(NAMESPACE) //
-				.getOrComputeIfAbsent(KEY, key -> createCloseablePath(parameterContext, extensionContext),
+				.getOrComputeIfAbsent(KEY,
+					key -> tempDirProvider.get(parameterContext, extensionContext, TEMP_DIR_PREFIX),
 					CloseablePath.class) //
 				.get();
 	}
 
-	private CloseablePath createCloseablePath(ParameterContext parameterContext, ExtensionContext extensionContext) {
-		Optional<Path> parentDir = getParentDirFromProvider(parameterContext, extensionContext);
+	private static CloseablePath createDefaultTempDir(String dirPrefix) {
 		try {
-			String prefix = "junit";
-			// @formatter:off
-			Path tempDirectory = parentDir.isPresent()
-					? Files.createTempDirectory(parentDir.get(), prefix)
-					: Files.createTempDirectory(prefix);
-			// @formatter:on
-			return new CloseablePath(tempDirectory);
+			return new CloseablePath(Files.createTempDirectory(dirPrefix));
 		}
-		catch (Exception e) {
-			throw new ParameterResolutionException("Failed to create temp directory", e);
+		catch (Exception ex) {
+			throw new ExtensionConfigurationException("Failed to create default temp directory", ex);
 		}
 	}
 
-	private Optional<Path> getParentDirFromProvider(ParameterContext parameterContext,
-			ExtensionContext extensionContext) {
+	private static CloseablePath createCustomTempDir(ParentDirProvider parentDirProvider,
+			ParameterContext parameterContext, ExtensionContext extensionContext, String dirPrefix) {
+		Path parentDir;
 		try {
-			return parentDirProvider.get(parameterContext, extensionContext);
+			parentDir = parentDirProvider.get(parameterContext, extensionContext);
 		}
 		catch (Exception ex) {
 			throw new ParameterResolutionException("Failed to get parent directory from provider", ex);
+		}
+		try {
+			return new CloseablePath(Files.createTempDirectory(parentDir, dirPrefix));
+		}
+		catch (Exception ex) {
+			throw new ParameterResolutionException("Failed to create custom temp directory", ex);
 		}
 	}
 
--- a/src/main/java/org/junit$pioneer/jupiter/TempDirectory.java
+++ b/src/main/java/org/junit$pioneer/jupiter/TempDirectory.java
@@ -235,6 +235,7 @@ public class TempDirectory implements ParameterResolver {
 		Path parentDir;
 		try {
 			parentDir = parentDirProvider.get(parameterContext, extensionContext);
+			requireNonNull(parentDir);
 		}
 		catch (Exception ex) {
 			throw new ParameterResolutionException("Failed to get parent directory from provider", ex);
--- a/src/main/java/org/junit$pioneer/jupiter/TempDirectory.java
+++ b/src/main/java/org/junit$pioneer/jupiter/TempDirectory.java
@@ -64,9 +64,9 @@ import org.junit.jupiter.api.extension.ParameterResolver;
  * <p>By default, this extension will use the default
  * {@link java.nio.file.FileSystem FileSystem} to create temporary directories
  * in the default location. However, you may instantiate this extension using
- * the {@link TempDirectory#TempDirectory(ParentDirProvider) TempDirectory(ParentDirProvider)}
- * or {@link TempDirectory#TempDirectory(Callable)} constructor and register it
- * via {@link org.junit.jupiter.api.extension.RegisterExtension @RegisterExtension}
+ * the {@link TempDirectory#createInCustomDirectory(ParentDirProvider)}
+ * or {@link TempDirectory#createInCustomDirectory(Callable)}} factory methods
+ * and register it via {@link org.junit.jupiter.api.extension.RegisterExtension @RegisterExtension}
  * to pass a custom provider to configure the parent directory for all temporary
  * directories created by this extension. This allows the use of this extension
  * with any third-party {@code FileSystem} implementation, e.g.
@@ -100,7 +100,7 @@ public class TempDirectory implements ParameterResolver {
 	 * {@link TempDirectory} extension this is used with.
 	 *
 	 * @see org.junit.jupiter.api.extension.RegisterExtension
-	 * @see TempDirectory#TempDirectory(ParentDirProvider)
+	 * @see TempDirectory#createInCustomDirectory(ParentDirProvider)
 	 */
 	@FunctionalInterface
 	public interface ParentDirProvider {
@@ -134,7 +134,7 @@ public class TempDirectory implements ParameterResolver {
 
 	private final TempDirProvider tempDirProvider;
 
-	public TempDirectory(TempDirProvider tempDirProvider) {
+	private TempDirectory(TempDirProvider tempDirProvider) {
 		this.tempDirProvider = requireNonNull(tempDirProvider);
 	}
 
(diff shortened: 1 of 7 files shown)


jbduncan:
@marcphilipp I would like to propose changing this PR to allow someone using this extension to be able to use a different underlying `FileSystem` implementation to the default on-disk one, like [Jimfs](https://github.com/google/jimfs)'s in-memory implementation.

I raise this proposal because I understand that people sometimes use Jimfs in place of the default `FileSystem` for testing.

But having said this, I'm not sure I fully understand why people use Jimfs for testing in the first place, and I also don't actually know how widespread the usage of alternative `FileSystem` implementations for testing is in general...

WDYT?

marcphilipp:
Interesting idea... I guess we could add a constructor to `TempDirectory` that takes a `FileSystem`. However, I'm not sure whether the extension should close it. In any case, that could be added later. 😉

marcphilipp:
Instead of a `FileSystem`, a `Path` representing the parent dir would work and communicate clearly that the extension will not close the `FileSystem`.

src/main/java/org/junit$pioneer/jupiter/TempDirectory.java
```diff
@@ -0,0 +1,124 @@
...
+ * Copyright 2015-2018 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v1.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+
+package org.junit$pioneer.jupiter;
```
sbrannen: Hmmmmm.... is the official package name really going to be `junit$pioneer` with a dollar sign?

src/main/java/org/junit$pioneer/jupiter/TempDirectory.java
```diff
@@ -0,0 +1,124 @@
...
+	/**
+	 * {@code Temp} can be used to annotate a test or lifecycle method parameter
+	 * of type {@link Path} that requires a temporary directory.
+	 *
+	 * @see TempDirectory
+	 */
+	@Target(ElementType.PARAMETER)
+	@Retention(RetentionPolicy.RUNTIME)
+	@Documented
+	public @interface Temp {
```
sbrannen: IMHO, `@Temp` is too generic. I'd prefer something more descriptive like `@TempDir`.

src/main/java/org/junit$pioneer/jupiter/TempDirectory.java
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	private static final Namespace NAMESPACE = Namespace.create(TempDirectory.class);
+	private static final String KEY = "dir";
```
sbrannen: I'd probably name the key `temp.dir`.

src/main/java/org/junit$pioneer/jupiter/TempDirectory.java
```diff
@@ -0,0 +1,124 @@
...
+
+	@Override
+	public boolean supportsParameter(ParameterContext parameterContext, ExtensionContext extensionContext) {
+		return parameterContext.isAnnotated(Temp.class);
+	}
+
+	@Override
+	public Object resolveParameter(ParameterContext parameterContext, ExtensionContext context) {
+		Class<?> parameterType = parameterContext.getParameter().getType();
+		if (!Path.class.equals(parameterType)) {
```
sbrannen: I think `!=` should suffice since you're comparing to a core Java type.

src/main/java/org/junit$pioneer/jupiter/TempDirectory.java
```diff
@@ -0,0 +1,124 @@
...
+			throw new ParameterResolutionException(
+				"Can only resolve parameters of type " + Path.class.getName() + " but was: " + parameterType.getName());
```
sbrannen: "parameter" should be singular.

sbrannen:
The documentation states that the scope is "per test"; however, it appears to me that the behavior would be different when used for constructor injection and with test instance per-class lifecycle.

So, if my assumption is correct, could you please update the implementation to take that into account?

sbrannen:
FYI: you might want to do something like I did in `org.junit.jupiter.engine.TrackLogRecords.Extension.supportsParameter(ParameterContext, ExtensionContext)` to limit usage to the test method level. 😉

src/main/java/org/junit$pioneer/jupiter/TempDirectory.java
marcphilipp: My reasoning was that it's `@Temp Path ...` so it should be clear what kind of temporary thing this is. But if another extension were to provide another temporary thing it would be confusing. So, you convinced me to change it.

src/main/java/org/junit$pioneer/jupiter/TempDirectory.java
marcphilipp: ok, will change that

src/main/java/org/junit$pioneer/jupiter/TempDirectory.java
marcphilipp: You got it.

marcphilipp:
@sbrannen @sormuras @jbduncan PTAL

src/main/java/org/junit$pioneer/jupiter/TempDirectory.java
sbrannen: 🤢

sbrannen: Thanks for the link. I'll follow up there...

src/main/java/org/junit$pioneer/jupiter/TempDirectory.java
```diff
@@ -0,0 +1,243 @@
...
+
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.extension.ExtensionContext.Namespace;
+import org.junit.jupiter.api.extension.ExtensionContext.Store.CloseableResource;
+import org.junit.jupiter.api.extension.ParameterContext;
+import org.junit.jupiter.api.extension.ParameterResolutionException;
+import org.junit.jupiter.api.extension.ParameterResolver;
+
+/**
+ * {@code TempDirectory} is a JUnit Jupiter extension to create and cleanup a
```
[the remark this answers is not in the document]
marcphilipp: D'oh.

src/main/java/org/junit$pioneer/jupiter/TempDirectory.java
```diff
@@ -0,0 +1,243 @@
...
+ *
+ * <p>The scope of the temporary directory depends on where the first
+ * {@link TempDir @TempDir} annotation is encountered when executing a test
+ * class. The temporary directory will be shared by all tests in a class when
+ * the annotation is present on a parameter of a
+ * {@link org.junit.jupiter.api.BeforeAll @BeforeAll} method or the test class
+ * constructor. Otherwise, e.g. when only used on test or
+ * {@link org.junit.jupiter.api.BeforeEach @BeforeEach} or
+ * {@link org.junit.jupiter.api.AfterEach @AfterEach} methods, each test will
+ * use its own temporary directory.
```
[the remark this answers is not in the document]
marcphilipp: There are tests for this, just not for `@TestInstance(PER_CLASS)`. I'll add some for the per class lifecycle.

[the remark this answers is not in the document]
sbrannen: Cool. Thanks! 👍

marcphilipp:
Thanks for the thorough review, @sbrannen!

I've addressed all your comments in the update commit. PTAL!

src/main/java/org/junit$pioneer/jupiter/TempDirectory.java
nicolaiparlog: @sbrannen I dare you. 😏

nicolaiparlog:
Before I get to the technical content, can we slim this PR down a bit? As I see it, does the following:

1. updates the copyright date (could be done in that upcoming PR I heard about where the license gets updated)
2. updates dependency on JUnit 5 from `5.0.0-SNAPSHOT` to `5.2.0` (we still didn't discuss what we want to build against - see #64)
3. adds logging infrastructure (can you give me a pointer why we need this?)
4. implements #39

Each of those should be its own issue and PR. Sorry for being a stickler. As I see it, this could be a quick way to go about it:

1. Leave that out for now as its not required for this PR.
2. Lets start discussing that immediately in #64.
3. Is that required for this PR?

marcphilipp:
> Each of those should be its own issue and PR.

These things are already separated into different commits (except the version update) that I intend to keep by using GitHub's "Rebase and merge" strategy. I can of course create multiple PRs for it but, in this case, that's just more work without any real benefit IMHO.

The extension implementation uses `ClosableResource` and thus needs at least 5.1.0. However, I don't see a reason why we shouldn't always use the latest version. I'll chime in on #64.

marcphilipp:
> adds logging infrastructure (can you give me a pointer why we need this?)

I configured the logging because failures (from tests) when closing a `CloseableResource` are currently logged by the Jupiter engine using the SEVERE log level.

marcphilipp:
I've downgraded to 5.1.1 which is the minimum we need because of https://github.com/junit-team/junit5/issues/1345.

nicolaiparlog:
> These things are already separated into different commits that I intend to keep by using GitHub's "Rebase and merge" strategy.

We purposely went out of your way to discourage that merging strategy because it requires us to quibble over commit messages (for which we have [high requirements](https://github.com/junit-pioneer/junit-pioneer/blob/master/CONTRIBUTING.md#merging)) while evaluating the PR. I know this is a little unusual, but I think [the results](https://github.com/junit-pioneer/junit-pioneer/commits/master) are worth it.

> I configured the logging because failures (from tests) when closing a `CloseableResource` are currently logged by the Jupiter engine using the SEVERE log level.

I would prefer to track such reasons in their own issue, which is then referenced in the final commit message.

Of the four commit, the copyright upgrade is the only one that can stand on its own. The other three should reference an issue and a PR and need a more detailed commit message.

marcphilipp:
I've now created separate issues for updating the copyright headers (#72), upgrading to 5.1.1 (#73), and configuring logging (#74). I consider this PR "frozen" until they have been resolved.

@nicolaiparlog @smoyer64 Please signal me whether you agree with the rationale behind each issue. If so, I will submit separate PRs to address them first before continuing to work on this PR.

src/main/java/org/junit$pioneer/jupiter/TempDirectory.java
smoyer64: I'm not going to comment on the $ in the package name but I am wondering about adding ``jupiter``.  Pioneer was supposed to imply that these were (all) jupiter extensions.  I think there would be some benefit in using the package to group together the "rule replacement" extensions though.

src/main/java/org/junit$pioneer/jupiter/TempDirectory.java
```diff
@@ -0,0 +1,245 @@
...
+	 * <p>The parent directory returned by the supplied {@link Callable} must
+	 * be {@code null} or an existing directory.
+	 *
+	 * <p>You may use this constructor when registering this extension via
+	 * {@link org.junit.jupiter.api.extension.RegisterExtension @RegisterExtension}.
+	 *
+	 * @param parentDirProvider used to configure the parent directory for the
+	 * temporary directories created by this extension; never {@code null}
+	 */
+	public TempDirectory(Callable<Path> parentDirProvider) {
```
smoyer64: I like this solution ... it's pretty agnostic and, in addition to supporting JimFS, it should work with the 15+ ``FileSystem`` types supported by [ApacheVFS](https://commons.apache.org/proper/commons-vfs/filesystems.html).

src/main/java/org/junit$pioneer/jupiter/TempDirectory.java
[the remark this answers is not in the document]
smoyer64: I'm going to need to learn the newer conventions for javadocs!

src/test/resources/log4j2-test.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
smoyer64: This probably needs to be a separate discussion but do we want to use log4j for test logging?  I would think if we're going to provide runtime logging for the extensions, we'd want to use a facade that didn't require an implementation.  For test logging should we  align to that?

sbrannen: I assume Marc used log4j 2 simply because that's what we do in JUnit 5, and I would have done the same.

Pioneer should use JUL (Java Util Logging) just like JUnit 5 does. Otherwise, Pioneer will require a dependency on a logging framework, and I think we should avoid that.

Note that we're only using log4j as the `LogManager` for JUL within our own test suite: that has no impact on users of Pioneer.

sbrannen: I agree with @smoyer64: we should introduce a `jupiter` subpackage for things specific to JUnit Jupiter.

src/main/java/org/junit$pioneer/jupiter/TempDirectory.java
marcphilipp: I was just following the lead of the existing code in this case. There's already a `org.junit$pioneer.vintage` package so a `org.junit$pioneer.jupiter` seemed only logical.

@smoyer64 Are you arguing for a subpackage?

It might be easiest for users to have all extensions in one package so they are easy to discover. On the other hand, a package per extension would allow to have package-private helper classes etc.

sbrannen: I agree with @marcphilipp: we should use a `jupiter` subpackage for things specific to JUnit Jupiter.

sbrannen: I think it's also worthwhile considering a package per extension since that will make it easier to maintain, especially given that we expect Pioneer to house multiple different types of extensions with varying levels of complexity (_read_: number of helper classes, annotations, interfaces, etc.).

src/test/resources/log4j2-test.xml
jbduncan: Would Google [Flogger](https://github.com/google/flogger) be worth considering? It has [many advantages](https://google.github.io/flogger/benefits) over other logging APIs, but I'm unsure if it's worth the extra dependency.

sbrannen: Ummm... you're missing the point. 😉 

Flogger wouldn't help: we don't want to use any logging _API_ other than JUL.

Rationale: we don't want JUnit or Pioneer to force users to add additional dependencies to their projects. JUL is always present in the JDK, and logging frameworks such as log4j provide "logger" _implementations_ for JUL.

jbduncan: > Ummm... you're missing the point. 😉

Flogger wouldn't help: we don't want to use any logging API other than JUL.

Rationale: we don't want JUnit or Pioneer to force users to add additional dependencies to their projects. JUL is always present in the JDK, and logging frameworks such as log4j provide "logger" implementations for JUL.

Oops, didn't quite realise that! Thanks for the clarification @sbrannen. 😅

smoyer64: > I assume Marc used log4j 2 simply because that's what we do in JUnit 5, and I would have done the same.

I just want to make sure we're all in agreement with that assumption - I don't have an issue with it.

> Pioneer should use JUL (Java Util Logging) just like JUnit 5 does. Otherwise, Pioneer will require a dependency on a logging framework, and I think we should avoid that.

Pioneer has the added caveat that we can use external dependencies that are "provided" scoped, so we could use another facade.  (I'm not suggesting this is one of those times and I don't know that Flogger would be my choice if we did).  I think during our telecon we all agreed that this was to be used very sparingly.

sbrannen: Right: JUnit and Pioneer should not log extensively.

And although the JUL _API_ isn't the nicest (maybe the worst?), that's really just a burden on the developers of JUnit and Pioneer. Users of the JUnit and Pioneer are then free to use whatever logging framework they like (and provide a bridge from JUL to their logging backend).

sbrannen: > Pioneer has the added caveat that we can use external dependencies that are "provided" scoped, so we could use another facade.

True. But I don't think logging would fall into that category of dependencies. As I recall, we agreed that Pioneer could have _optional_ dependencies on other third-party frameworks (e.g., database connectivity, server APIs, etc.).

Of course, if an optional third-party dependency brings in logging dependencies transitively, that would be a different story. 😉

smoyer64: Yes ... "optional" (I'd use "provided" scope in a Maven project).  Won't Gradle scope the transitive dependencies for an "optional" dependency as "optional" as well?

sbrannen: Gradle doesn't actually have an `optional` scope. That's more like `compileOnly` & Co.

But there's a plugin from the Spring team that _implements_ an `optional` scope in Gradle that gets exported as Maven's `optional` scope.

And... I _think_ an upcoming version of Gradle might support `optional` out of the box, though I imagine @marcphilipp can better answer that. 😉

marcphilipp:
So, finally, we're down to a single commit that "just" introduces the extension.

@nicolaiparlog Please take another look! 🙂

nicolaiparlog:
Great documentation (let's see whether it automagically appears as I planned :wink:), commit message, and all other fluffy stuff. To make this quick, I'll just trust you and your many reviewers on the code. :wink:

What about [the build error](https://ci.appveyor.com/project/nicolaiparlog/junit-pioneer/build/marc/issues/39-temp-directory-109#L447), though? Looks spurious and I'm fine with ignoring it. If you are to, go ahead and merge.

nicolaiparlog:
Re commit message, remember to add the PR:

```
Introduce TempDirectory Jupiter Extension (#39 / #69)

The `TempDirectory` extension can be used to create and clean up a
temporary directory for an individual test or all tests in a test class.
To use it, simply register the extension and add a parameter of type
`java.nio.file.Path` to your test or lifecycle method or constructor:

  @ExtendWith(TempDirectory.class)
  class Tests {
    @Test
    void test(@TempDir Path tempDir) throws Exception {
      Path file = tempDir.resolve("test.txt");
      writeFile(file);
      assertExpectedFileContent(file);
    }
  }

In addition to the default file system, the extension can be used with
any FileSystem implementation, e.g. Jimfs [1]. In order to use a custom
file system, simply register the extension programmatically and pass a
provider of a custom parent directory of type `Path` to its constructor.

[1] https://github.com/google/jimfs
```

src/test/java/org/junit$pioneer/jupiter/TempDirectoryTests.java
```diff
@@ -0,0 +1,484 @@
...
+		@Test
+		@DisplayName("from ParentDirProvider")
+		void resolvesTempDirWithCustomParentDirFromProvider() {
+			assertResolvesSeparateTempDirs(ParentDirFromProviderTestCase.class);
+		}
+	}
+
+	@Nested
+	@DisplayName("reports failure")
+	class Foo {
```
nicolaiparlog: Can you improve that name?

src/main/java/org/junit$pioneer/jupiter/TempDirectory.java
```diff
@@ -0,0 +1,245 @@
...
+	@FunctionalInterface
+	public interface ParentDirProvider {
+		/**
+		 * Get the parent directory for all temporary directories created by the
+		 * {@link TempDirectory} extension this is used with.
+		 *
+		 * @return the parent directory for all temporary directories; must be
+		 * {@code null} or an existing directory.
+		 */
+		Path get(ParameterContext parameterContext, ExtensionContext extensionContext) throws Exception;
```
nicolaiparlog: Any reason why this is not an `Optional`?

src/main/java/org/junit$pioneer/jupiter/TempDirectory.java
```diff
@@ -0,0 +1,245 @@
...
+		this((parameterContext, extensionContext) -> null);
+	}
+
+	/**
+	 * Create a new {@code TempDirectory} extension that uses the supplied
+	 * {@link Callable} to configure the parent directory for the temporary
+	 * directories created by this extension.
+	 *
+	 * <p>The parent directory returned by the supplied {@link Callable} must
+	 * be {@code null} or an existing directory.
```
nicolaiparlog: `Optional` instead of `null`.

src/main/java/org/junit$pioneer/jupiter/TempDirectory.java
```diff
@@ -0,0 +1,245 @@
...
+	 *
+	 * <p>You may use this constructor when registering this extension via
+	 * {@link org.junit.jupiter.api.extension.RegisterExtension @RegisterExtension}.
+	 *
+	 * @param parentDirProvider used to configure the parent directory for the
+	 * temporary directories created by this extension; never {@code null}
```
nicolaiparlog: You can remove `never {@code null}` - in Pioneer nothing is ever allowed to be `null`.

src/main/java/org/junit$pioneer/jupiter/TempDirectory.java
```diff
@@ -0,0 +1,245 @@
...
+	/**
+	 * Create a new {@code TempDirectory} extension that uses the supplied
+	 * {@link ParentDirProvider} to configure the parent directory for the
+	 * temporary directories created by this extension.
+	 *
+	 * <p>You may use this constructor when registering this extension via
+	 * {@link org.junit.jupiter.api.extension.RegisterExtension @RegisterExtension}.
+	 *
+	 * @param parentDirProvider used to configure the parent directory for the
+	 * temporary directories created by this extension; never {@code null}
```
nicolaiparlog: Remove `never {@code null}`.

src/main/java/org/junit$pioneer/jupiter/TempDirectory.java
```diff
@@ -0,0 +1,245 @@
...
+	 */
+	public TempDirectory(ParentDirProvider parentDirProvider) {
+		this.parentDirProvider = Objects.requireNonNull(parentDirProvider);
```
nicolaiparlog: I usually statically import `requireNonNull`.

src/test/java/org/junit$pioneer/jupiter/TempDirectoryTests.java
```diff
@@ -0,0 +1,484 @@
...
+import org.junit.platform.engine.test.event.ExecutionEvent;
+import org.junit.platform.engine.test.event.ExecutionEventRecorder;
+
+@DisplayName("TempDirectory extension")
+class TempDirectoryTests extends AbstractPioneerTestEngineTests {
+
+	@BeforeEach
+	@AfterEach
+	void resetStaticVariables() {
+		BaseSharedTempDirTestCase.tempDir = null;
```
nicolaiparlog: Replace `null` with `Optional`.

src/test/java/org/junit$pioneer/jupiter/TempDirectoryTests.java
```diff
@@ -0,0 +1,484 @@
...
+	}
+
+	@TestInstance(PER_CLASS)
+	static class AnnotationOnBeforeAllMethodParameterWithTestInstancePerClassTestCase
+			extends AnnotationOnBeforeAllMethodParameterTestCase {
+	}
+
+	@ExtendWith(TempDirectory.class)
+	static class AnnotationOnAfterAllMethodParameterTestCase {
+		static Path firstTempDir;
```
nicolaiparlog: Looks nullable...

nicolaiparlog:
I'm decided to work on the `null` issue myself, so my requirements don't cause you additional work. :)

nicolaiparlog:
Done, but not happy. While updating the code it felt like there's something hiding. As I see it, `ParentDirProvider` serves two masters:

* the user needs it to pass a recipe for a (custom) parent directory
* the extension needs it to create either the default or a custom parent directory

The difference between default and custom is expressed by a missing value, right?

If that's so, the user should never pass in a missing directory. I would hence prefer to make `ParentDirProvider`'s return value mandatory (non-optional/non-nullable) and have `TempDirectory`'s constructors create strategies for `createCloseablePath` based on their parameters.

If my analysis is correct and you agree with the course of action, I can make the changes myself.

marcphilipp:
Agreed, let’s make it mandatory! Let me know if you have time to make the changes. Otherwise, I can take care of it.

nicolaiparlog:
> Agreed, let’s make it mandatory! Let me know if you have time to make the changes.

I do and I already started. If today's ICE has a working WiFi, I will push some changes later today.

nicolaiparlog:
I pushed two commits:

* 4352267 removes `Optonal` and makes the returned parent directory compulsory
* 1f8bee1 replaces constructor calls with static factory methods

I generally prefer such methods over constructors because they can get proper names (in this case `createInDefaultDirectory` and `createInCustomDirectory`). Unfortunately in this case, they clash with the fact that Jupiter's extension model requires a public parameterless constructor, so one such constructor remains visible - that's a little weird. I still prefer having the factories but understand if others see it differently. WDYT?

(It is possible the JUnit 5 team already had that discussion, in which case we can cut this short and simply do the same thing here, possibly reverting/removing 1f8bee1).

sbrannen:
> I generally prefer such methods over constructors because they can get proper names (in this case `createInDefaultDirectory` and `createInCustomDirectory`). Unfortunately in this case, they clash with the fact that Jupiter's extension model requires a public parameterless constructor, so one such constructor remains visible - that's a little weird. I still prefer having the factories but understand if others see it differently. WDYT?

I don't think it's weird at all.

The default constructor is required for declarative extension registration via `@ExtendWith` and automatic extension registration via the `ServiceLoader` mechanism. There is literally no way around the latter, and there is currently no way around the former (and no proposed changes on the horizon).

On the other hand, the static factory methods are great for programmatic extension registration via `@RegisterExtension`.

So I think the status quo is perfectly fine _as is_.  👍 

> (It is possible the JUnit 5 team already had that discussion, in which case we can cut this short and simply do the same thing here, possibly reverting/removing 1f8bee1).

What do you mean by the "same thing"? 🤔

nicolaiparlog:
> What do you mean by the "same thing"?

Right, that's not clear.

If I use static factories, I make all constructors non-public, but this is of course not possible here (for the reasons you laid out). I find the resulting mixture of public constructors and factory methods a little unfortunate, but think all in all, factories are still the way to go. It's of course possible that others come to a different conclusion and want to remove the factories (which I'd be fine with).

I thought the JUnit 5 team might already have had that discussion for your own extensions, in which case I assume some conclusion has been reached. "The same thing" referenced that conclusion. If you decided that, for extensions, factories are the way to go or a thing to avoid, we could just have applied that conclusion here without further discussion.

sbrannen:
> the comments show up but only because they are on the commit, not as part of the typical PR comments. Just curious whether there's a reason for that.

I simply replied in context.

sbrannen:
Thanks to the discussion here, I realized that the _constructor requirements_ for extensions were not well documented. So I addressed that in https://github.com/junit-team/junit5/commit/83cd64f4b44f6196a266e5bfd7f0186c08d59c5a.

sbrannen:
> If you decided that, for extensions, factories are the way to go or a thing to avoid, we could just have applied that conclusion here without further discussion.

We decided that extension authors should do whatever makes sense with regard to how an extension is intended to be used/registered.

Most of our examples have default constructors; however, our examples for extensions registered via `@RegisterExtension` provide static factory methods and builder APIs.

nicolaiparlog:
@marcphilipp Do you have time to take a look at my changes?

marcphilipp:
@nicolaiparlog I pushed one small polishing commit. Otherwise, I think this PR is ready to be merged. Thanks for making the changes, looks much better now!

src/main/java/org/junit$pioneer/jupiter/TempDirectory.java
```diff
@@ -168,7 +167,7 @@ public static TempDirectory createInDefaultDirectory() {
 	}
 
 	/**
-	 * Returns a {@code TempDirectory} extension that uses the supplied
+	 * Create a new {@code TempDirectory} extension that uses the supplied
```
nicolaiparlog: Is there a good reason to promise a fresh instance? I prefer the contract of static factory methods to be purposely vague about what exactly is returned. This way there is more room to refactor.

marcphilipp: Agreed, I changed it back. Ready to be merged now IMHO. 🙂

sbrannen:
Yay....... a new feature! 😄

Issue #72, opened 2018-05-09
Change license to EPL v2.0


marcphilipp:
The JUnit team has released JUnit 5 under EPL v2.0 which came out just in time for the first GA release (cf. https://github.com/junit-team/junit5/issues/1045).

AFAIK Pioneer picked EPL v1.0 because of JUnit. Thus, it would make sense to follow JUnit's lead and "upgrade" to EPL v2.0. That would make it a no brainer to move code between the two projects.

@smoyer64 @nicolaiparlog If you agree, I can submit a PR in the next few days.

sbrannen:
> That would make it a no brainer to move code between the two projects.

I'd say that is actually essential to allow us to move code between the two projects.

In other words, having different licenses is in some ways a blocker for the success of Pioneer.

smoyer64:
> That would make it a no brainer to move code between the two projects.

Yes ... I think that's the intent.  Provide a place where community contributions could live and (possibly) mature into parts of the core JUnit 5 project.

nicolaiparlog:
I agree that the license should be upgraded. Does that change need explicit agreement by all committers? If so, judging by the fact that the opened the issue, I guess @marcphilipp is fine with it and I am, too. I will just assume @smoyer64 agrees as well (can you tell us explicitly, though?) and will prepare a PR soon.

smoyer64:
@nicolaiparlog I agree

nicolaiparlog:
> Does Spotless take care of putting the license at the top of each file? Or is there a Gradle equivalent to the Maven RAT plugin?

Yes, Spotless does that.

marcphilipp:
Done in #80.

PR #80, opened 2018-05-15
Change license to EPL v2.0

nicolaiparlog:
As discussed in #72, the project license is updated to EPL 2.0.

There are two commits:

* the first one hopefully updates all the infrastructure bits to the new license - it would be great if someone could give it a close examination
* the second one simply applies the change to all source files - under the assumption that Spotless works correctly, there is no reason to examine it

---

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,4 +1,4 @@
 
 ---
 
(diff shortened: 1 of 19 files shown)


nicolaiparlog:
Proposed commit message:

```
Update Eclipse Public License from 1.0 to 2.0 (#72 / #73)

EPL v1.0 was chosen as JUnit Pioneer's license because JUnit 5 used
it. No longer, though, because it was updated to 2.0 right before the
first GA release (see junit-team/junit5#1045).

This change realigns JUnit Pioneer with JUnit 5.
```

PR #84, opened 2018-05-15
Configure logging during test execution using log4j (#74)

### LINKED ISSUE #74: Configure logging
JUnit uses JUL as its logging framework. When testing extensions, we often need to check for error cases that sometimes produce quite verbose console output when running the Gradle build. Thus, we should add the necessary dependencies to redirect JUL to log4j2 and add a basic log4j config file.

marcphilipp:
To suppress unnecessary output from JUnit during test execution, e.g.
for error cases that are tested explicitly.

Resolves #74.


nicolaiparlog:
#74 mentions that `@TrackLogRecords` should be copied over. That was blocked by #72, which is no longer the case. Do you want to add that?

When you're done, can you combine issue and PR descriptions to propose a commit message? If I got this right, the main reason to add a configuration is to set the level to `WARN` and thus avoid spamming the log with messages that were provoked on purpose. If that's right, please mention it.

marcphilipp:
In my book, `@TrackLogRecords` is different from configuring logging for test execution. While the former is used to test that a piece of code logs something, the latter is used to configure what should be logged by JUnit when executing the tests. Thus, I'd rather create a separate issue/PR for that. Deal?

nicolaiparlog:
The issue description wasn't 100% clear to me (why _exactly_ are we making this change?) and thus the commit message isn't either - hence my request to clarify and propose a message here. In the future I'll make sure to "Request changes", so these things don't go under.

PR #88, opened 2018-05-19
Simplify Gradle publishing configuration

### LINKED ISSUE #83: Simplify Gradle publishing configuration
Currently, the build script defines two Maven repositories in the `publishing` block which leads to Gradle creating two publishing tasks:

https://github.com/junit-pioneer/junit-pioneer/blob/9ada30b5ce06d5dd9fc1793a11dd665aa80f8268/build.gradle#L260-L277

The build script then takes extra care to disable the `publish` lifecycle task (because it would attempt to publish to both) and only enable the snapshot publishing tasks for snapshots and the release publishing one for releases:

https://github.com/junit-pioneer/junit-pioneer/blob/9ada30b5ce06d5dd9fc1793a11dd665aa80f8268/build.gradle#L167-L194

Those decisions are entirely based on the `isReleaseVersion` property:

https://github.com/junit-pioneer/junit-pioneer/blob/9ada30b5ce06d5dd9fc1793a11dd665aa80f8268/build.gradle#L17

## Proposal

Instead, it would be much simpler if we only configured a single Maven repository and used `isReleaseVersion` to determine it's URL:

```gradle
publishing {
    repositories {
        maven {
            def releasesRepoUrl = 'https://oss.sonatype.org/service/local/staging/deploy/maven2/'
            def snapshotsRepoUrl = 'https://oss.sonatype.org/content/repositories/snapshots/'
            url isReleaseVersion ? releasesRepoUrl : snapshotsRepoUrl
            credentials {
                username mavenUserName
                password mavenPassword
            }
        }
    }
}
```

As a matter of fact, a very similar example has just been published as part of the [Gradle 4.8 RC1 User Guide chapter on the Maven Publish Plugin](https://docs.gradle.org/4.8-rc-1/userguide/publishing_maven.html#publishing_maven:snapshot_and_release_repositories). Disclaimer: I wrote it. 😉

marcphilipp:
- Let Gradle set up task dependencies for sourcesJar and javadocJar
   automatically
 - Upgrade to Gradle 4.8 RC1 to use built-in support for signing
   publications and configuring the POM
 - Use single publishing repo and remove custom configuration that
   ensured that only snapshots were published to the snapshot repo and
   releases to the staging repo

Resolves #83.

~This PR requires features introduced in Gradle 4.8 RC1. If that's too bleeding edge, we can wait for the GA release.~ Gradle 4.8 has now been released.

---


nicolaiparlog:
The changes look good. To test, I published a snapshot build and checked [its pom](https://oss.sonatype.org/content/repositories/snapshots/org/junit-pioneer/junit-pioneer/0.1-SNAPSHOT/junit-pioneer-0.1-20180522.082407-23.pom) - :+1:.

But I got this message on the command line:

> Deprecated Gradle features were used in this build, making it incompatible with Gradle 5.0.
> See https://docs.gradle.org/4.8-rc-1/userguide/command_line_interface.html#sec:command_line_warnings

Should we fix that as well?

> This PR requires features introduced in Gradle 4.8 RC1. If that's too bleeding edge, we can wait for the GA release.

Do you have an idea when GA can be expected? If it's another week or two, we can wait, but beyond that, I'm fine with the bleeding edge.

marcphilipp:
The deprecation warning was caused by the old version of Spotless. I've now merged the Spotless PR and rebased this PR onto master which made it go away.

Gradle 4.8 is scheduled for Tuesday but it looks like there might be an RC3 which would delay it by a few days. I'll let you make the call whether to wait or not as merging this PR isn't urgent.

nicolaiparlog:
> Gradle 4.8 is scheduled for Tuesday but it looks like there might be an RC3 which would delay it by a few days.

Then let's just wait for a few more days until 4.8 is GA.

nicolaiparlog:
Thanks Marc, this is a great simplification! :+1:

Issue #103, opened 2018-09-06
Setting Default Locale and TimeZone for a test


britter:
Hi,

I'm currently migrating the Apache Commons Lang project to JUnit 5. Since we have `Locale` and `TimeZone` dependent tests, we wrote a [custom JUnit 4 rule](https://github.com/apache/commons-lang/blob/master/src/test/java/org/apache/commons/lang3/test/SystemDefaultsSwitch.java) for that. Since I need to migrate this to JUnit 5 some time, I thought about bringing it here. The JUnit 4 API looks like this:

```java
   public class SystemDefaultsDependentTest {
  
        @Rule
       public SystemDefaultsSwitch locale = new SystemDefaultsSwitch();
  
        @Test
       public void testThatWillExecuteWithTheDefaultLocaleAndTimeZone() {
           // nothing to do, just implement the test
       }
  
        @Test
        @SystemDefaults(locale="zh_CN")
       public void testWithSimplifiedChinaDefaultLocale() {
           // Locale.getDefault() will return Locale.CHINA until the end of this test method
       }
  
        @Test
        @SystemDefaults(timezone="America/New_York")
       public void testWithNorthAmericaEasternTimeZone() {
           // TimeZone.getDefault() will equal TimeZone.getTimeZone("America/New_York") until the end of this method
       }
   }
```

Is this something you consider in scope of junit-pioneer? If that's the case I'm happy bring it here ported to JUnit 5.

marcphilipp:
I think that would be very useful!

sormuras:
_PPP_ 

Please Provide PR. ;-)

britter:
okay, going to work on this!

nicolaiparlog:
> Is this something you consider in scope of junit-pioneer?

Yes, absolutely! :1:

Issue #87, opened 2018-05-17
Use ShipKit to release continuously


marcphilipp:
The Mockito project has written [Shipkit](https://github.com/mockito/shipkit), a Gradle plugin to release with every commit to `master`.

From the readme:

> You will be more productive if your releases are fully automated and happen on every change. You will build great product if you can focus on code & features, but not on the release overhead. Shipkit will make it happen. Shipkit is a toolkit for shipping it.

I think we should seriously consider using it.

It looks like it currently only supports Bintray, though. @mockitoguy Is that correct?

Thoughts?

nicolaiparlog:
That sounds great (and would fix #53), but releasing every commit? Phew, I'm not sure whether I'm up to that. I mean, are we sure our code (and tests) are good enough, that every merged PR hits home? I know they technically _should_, but _will_ they?

We're still before `1.0`, so stability requirements are lower - we could just give it a try and observe. Although without users (that I know of), we will hardly know if we screw up.

mockitoguy:
>It looks like it currently only supports Bintray, though. @mockitoguy Is that correct?

Yes, correct. Is this a problem? Bintray is a great repo, very easy to use. We release to Maven Central via Bintray.

In Shipkit we want to support releasing directly to Sonatype. We haven't implemented it yet due to other priorities. It would be great if you tried out Shipkit!

>I mean, are we sure our code (and tests) are good enough, that every merged PR hits home? I know they technically should, but will they?

It's a great forcing function. Your tests will have to be great. I absolutely recommend this model.

nicolaiparlog:
I can see up- and downsides and end up with not having a strong opinion either way. Everybody else (who chimed in) seems to think this is worth a try, so if nobody speaks up against this soon, we should do it.

marcphilipp:
First step:
https://github.com/junit-pioneer/junit-pioneer/commit/64afd4fd50c9f26019f5d3f881d1c31720668b4f

### Next steps:

- [ ] [Setup Bintray](https://blog.bintray.com/2014/02/11/bintray-as-pain-free-gateway-to-maven-central/)
- [ ] [Resolve TODOs in `gradle/shipkit.gradle`](https://github.com/mockito/shipkit/blob/master/docs/getting-started.md#shipkitgradle-configuration-file)
- [ ] [Setup Travis to use Bintray](https://github.com/mockito/shipkit/blob/master/docs/getting-started.md#production-configuration)

[...]

PR #104, opened 2018-09-06
Setting Default Locale and TimeZone for a test (#103)

britter:
PoC for the feature described in #103. I'm not very fluent in writing JUnit 5 extensions so I'd like to get some early feedback, before I implement this. Is this going in the right direction?

Thank you!
---

### DIFF
--- a/docs/default-locale-timezone.adoc
+++ b/docs/default-locale-timezone.adoc
@@ -1,0 +1,93 @@
+:page-title: @DefaultLocale and @DefaultTimeZone
+:page-description: JUnit Jupiter extensions to change the values returned from Locale.getDefault() and TimeZone.getDefault()
+
+The `@DefaultLocale` and `@DefaultTimeZone` annotations can be used to change the values returned from `Locale.getDefault()` and `TimeZone.getDefault()`.
+Both annotations work on the test class level and on the test method level.
+After the annotated element has been executed the initial default value is restored.
+
+== `@DefaultLocale`
+
+The default `Locale` is specified according to the constructors of https://docs.oracle.com/javase/10/docs/api/java/util/Locale.html[Locale].
+
+[source,java]
+----
+@Test
+@DefaultLocale(language = "en")
+void test_with_lanuage() {
+    assertEquals(new Locale("en"), Locale.getDefault());
+}
+
+@Test
+@DefaultLocale(language = "en", country = "EN")
+void test_with_lanuage_and_country() {
+    assertEquals(new Locale("en", "EN"), Locale.getDefault());
+}
+
+@Test
+@DefaultLocale(language = "en", country = "EN", variant = "gb")
+void test_with_lanuage_and_country_and_vairant() {
+    assertEquals(new Locale("en", "EN", "gb"), Locale.getDefault());
+}
+----
+
+Note that variant can only be specified when the country is also specified.
+If a variant is specified without a country an `ExtentionConfigurationException` will be thrown.
+
+Any method level `@DefaultLocale` configurations will override class level configurations.
+
+[source,java]
+----
+@DefaultLocale(language = "fr")
+class MyLocaleTests {
+
+    @Test
+    void test_with_class_level_configuration() {
+        assertEquals(new Locale("fr"), Locale.getDefault());
+    }
+
+    @Test
+    @DefaultLocale(language = "en")
+    void test_with_method_level_configuration() {
+        assertEquals(new Locale("en"), Locale.getDefault());
+    }
+}
+----
+
+== `@DefaultTimeZone`
+
+The default `TimeZone` is specified according to the https://docs.oracle.com/javase/10/docs/api/java/util/TimeZone.html#getTimeZone(java.lang.String)[TimeZone.getTimeZone(String)] method.
+
+[source,java]
+----
+@Test
+@DefaulTimeZone("CET")
+void test_with_short_zone_id() {
+    assertEquals(TimeZone.getTimeZone("CET"), TimeZone.getDefault());
+}
+
+@Test
+@DefaultTimeZone("America/Los_Angeles")
+void test_with_long_zone_id() {
+    assertEquals(TimeZone.getTimeZone("America/Los_Angeles"), TimeZone.getDefault());
+}
+----
+
+Any method level `@DefaultTimeZone` configurations will override class level configurations:
+
+[source,java]
+----
+@DefaultTimeZone("CET")
+class MyTimeZoneTests {
+
+    @Test
+    void test_with_class_level_configuration() {
+        assertEquals(TimeZone.getTimeZone("CET"), TimeZone.getDefault());
+    }
+
+    @Test
+    @DefaultTimeZone("America/Los_Angeles")
+    void test_with_method_level_configuration() {
+        assertEquals(TimeZone.getTimeZone("America/Los_Angeles"), TimeZone.getDefault());
+    }
+}
+----
--- a/docs/default-locale-timezone.adoc
+++ b/docs/default-locale-timezone.adoc
@@ -7,7 +7,11 @@ After the annotated element has been executed the initial default value is resto
 
 == `@DefaultLocale`
 
-The default `Locale` is specified according to the constructors of https://docs.oracle.com/javase/10/docs/api/java/util/Locale.html[Locale].
+The default `Locale` is specified according to https://docs.oracle.com/javase/8/docs/api/java/util/Locale.html#constructor.summary[its constructors] - it needs either:
+
+* `language` or
+* `language` and `country` or
+* `language`, `country`, and `variant`
 
 [source,java]
 ----
@@ -30,8 +34,8 @@ void test_with_lanuage_and_country_and_vairant() {
 }
 ----
 
-Note that variant can only be specified when the country is also specified.
-If a variant is specified without a country an `ExtentionConfigurationException` will be thrown.
+Note that `variant` can only be specified if `country` is also specified.
+If `variant` is specified without `country`, an `ExtentionConfigurationException` will be thrown.
 
 Any method level `@DefaultLocale` configurations will override class level configurations.
 
@@ -55,7 +59,7 @@ class MyLocaleTests {
 
 == `@DefaultTimeZone`
 
-The default `TimeZone` is specified according to the https://docs.oracle.com/javase/10/docs/api/java/util/TimeZone.html#getTimeZone(java.lang.String)[TimeZone.getTimeZone(String)] method.
+The default `TimeZone` is specified according to the https://docs.oracle.com/javase/8/docs/api/java/util/TimeZone.html#getTimeZone(java.lang.String)[TimeZone.getTimeZone(String)] method.
 
 [source,java]
 ----
--- a/src/main/java/org/junitpioneer/jupiter/DefaultLocale.java
+++ b/src/main/java/org/junitpioneer/jupiter/DefaultLocale.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright 2015-2018 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junitpioneer.jupiter;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.junit.jupiter.api.extension.ExtendWith;
+
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+@ExtendWith(DefaultLocaleExtension.class)
+public @interface DefaultLocale {
+
+	String value();
+}
--- a/src/main/java/org/junitpioneer/jupiter/DefaultLocale.java
+++ b/src/main/java/org/junitpioneer/jupiter/DefaultLocale.java
@@ -17,7 +17,7 @@ import java.lang.annotation.Target;
 
 import org.junit.jupiter.api.extension.ExtendWith;
 
-@Target(ElementType.METHOD)
+@Target({ ElementType.METHOD, ElementType.TYPE })
 @Retention(RetentionPolicy.RUNTIME)
 @ExtendWith(DefaultLocaleExtension.class)
 public @interface DefaultLocale {
--- a/src/main/java/org/junitpioneer/jupiter/DefaultLocale.java
+++ b/src/main/java/org/junitpioneer/jupiter/DefaultLocale.java
@@ -22,5 +22,9 @@ import org.junit.jupiter.api.extension.ExtendWith;
 @ExtendWith(DefaultLocaleExtension.class)
 public @interface DefaultLocale {
 
-	String value();
+	String language();
+
+	String country() default "";
+
+	String variant() default "";
 }
--- a/src/main/java/org/junitpioneer/jupiter/DefaultLocale.java
+++ b/src/main/java/org/junitpioneer/jupiter/DefaultLocale.java
@@ -36,6 +36,7 @@ import org.junit.jupiter.api.extension.ExtendWith;
  *
  * @since 0.2
  * @see java.util.Locale#getDefault()
+ * @see DefaultTimeZone
  */
 @Target({ ElementType.METHOD, ElementType.TYPE })
 @Retention(RetentionPolicy.RUNTIME)
--- a/src/main/java/org/junitpioneer/jupiter/DefaultLocale.java
+++ b/src/main/java/org/junitpioneer/jupiter/DefaultLocale.java
@@ -22,17 +22,17 @@ import org.junit.jupiter.api.extension.ExtendWith;
  * returned by {@link java.util.Locale#getDefault()} for a test execution.
  *
  * <p>The {@link java.util.Locale} to set as the default locale can be
- * configured using only a language, a language and a country or a language, a
+ * configured using only a language, a language and a country, or a language, a
  * country and a variant. If {@link #variant()} is set, but {@link #country()}
  * is not, a
  * {@link org.junit.jupiter.api.extension.ExtensionConfigurationException} will
- * be thrown. After the annotated element has been executed, the default Locale
- * will be restored to its original value.</p>
+ * be thrown. After the annotated element has been executed, the default
+ * {@code Locale} will be restored to its original value.</p>
  *
  * <p>{@code @DefaultLocale} can be used on the method and on the class level.
- * If a class is annotated, the configured Locale will be the default Locale
- * for all tests inside that class. Any method level configurations will
- * override the class level default Locale.</p>
+ * If a class is annotated, the configured {@code Locale} will be the default
+ * {@code Locale} for all tests inside that class. Any method level
+ * configurations will override the class level default {@code Locale}.</p>
  *
  * @since 0.2
  * @see java.util.Locale#getDefault()
@@ -43,9 +43,23 @@ import org.junit.jupiter.api.extension.ExtendWith;
 @ExtendWith(DefaultLocaleExtension.class)
 public @interface DefaultLocale {
 
+	/**
+	 * An ISO 639 alpha-2 or alpha-3 language code, or a language subtag up to
+	 * 8 characters in length. See the {@link java.util.Locale} class
+	 * description about valid language values.
+	 */
 	String language();
 
+	/**
+	 * An ISO 3166 alpha-2 country code or a UN M.49 numeric-3 area code. See
+	 * the {@link java.util.Locale} class description about valid country
+	 * values.
+	 */
 	String country() default "";
 
+	/**
+	 * Any arbitrary value used to indicate a variation of a {@code Locale}.
+	 * See the {@link java.util.Locale} class description for the details.
+	 */
 	String variant() default "";
 }
--- a/src/main/java/org/junitpioneer/jupiter/DefaultLocale.java
+++ b/src/main/java/org/junitpioneer/jupiter/DefaultLocale.java
@@ -17,6 +17,26 @@ import java.lang.annotation.Target;
 
 import org.junit.jupiter.api.extension.ExtendWith;
 
+/**
+ * {@code @DefaultLocale} is a JUnit Jupiter extension to change the value
+ * returned by {@link java.util.Locale#getDefault()} for a test execution.
+ *
+ * <p>The {@link java.util.Locale} to set as the default locale can be
+ * configured using only a language, a language and a country or a language, a
+ * country and a variant. If {@link #variant()} is set, but {@link #country()}
+ * is not, a
+ * {@link org.junit.jupiter.api.extension.ExtensionConfigurationException} will
+ * be thrown. After the annotated element has been executed, the default Locale
+ * will be restored to its original value.</p>
+ *
+ * <p>{@code @DefaultLocale} can be used on the method and on the class level.
+ * If a class is annotated, the configured Locale will be the default Locale
+ * for all tests inside that class. Any method level configurations will
+ * override the class level default Locale.</p>
+ *
+ * @since 0.2
+ * @see java.util.Locale#getDefault()
+ */
 @Target({ ElementType.METHOD, ElementType.TYPE })
 @Retention(RetentionPolicy.RUNTIME)
 @ExtendWith(DefaultLocaleExtension.class)
--- a/src/main/java/org/junitpioneer/jupiter/DefaultLocaleExtension.java
+++ b/src/main/java/org/junitpioneer/jupiter/DefaultLocaleExtension.java
@@ -62,7 +62,7 @@ class DefaultLocaleExtension implements BeforeAllCallback, BeforeEachCallback, A
 		return AnnotationSupport
 				.findAnnotation(context.getElement(), DefaultLocale.class)
 				.map(DefaultLocaleExtension::createLocale)
-				.orElseThrow(() -> new ExtensionConfigurationException("TODO"));
+				.orElseThrow(() -> new ExtensionConfigurationException("The extension is active, but the corresponding annotation could not be found. (This may be a bug.)"));
 		//@formatter:on
 	}
 
--- a/src/main/java/org/junitpioneer/jupiter/DefaultLocaleExtension.java
+++ b/src/main/java/org/junitpioneer/jupiter/DefaultLocaleExtension.java
@@ -95,6 +95,6 @@ class DefaultLocaleExtension implements BeforeAllCallback, BeforeEachCallback, A
 	}
 
 	private void resetDefaultLocale(ExtensionContext context) {
-		Locale.setDefault((Locale) context.getStore(NAMESPACE).get(KEY));
+		Locale.setDefault(context.getStore(NAMESPACE).get(KEY, Locale.class));
 	}
 }
--- a/src/main/java/org/junitpioneer/jupiter/DefaultLocaleExtension.java
+++ b/src/main/java/org/junitpioneer/jupiter/DefaultLocaleExtension.java
@@ -13,27 +13,37 @@ package org.junitpioneer.jupiter;
 import java.util.Locale;
 import java.util.Optional;
 
+import org.junit.jupiter.api.extension.AfterAllCallback;
 import org.junit.jupiter.api.extension.AfterEachCallback;
+import org.junit.jupiter.api.extension.BeforeAllCallback;
 import org.junit.jupiter.api.extension.BeforeEachCallback;
 import org.junit.jupiter.api.extension.ExtensionConfigurationException;
 import org.junit.jupiter.api.extension.ExtensionContext;
 
-public class DefaultLocaleExtension implements BeforeEachCallback, AfterEachCallback {
+class DefaultLocaleExtension implements BeforeAllCallback, BeforeEachCallback, AfterAllCallback, AfterEachCallback {
 
 	private Locale defaultLocale;
 
 	@Override
+	public void beforeAll(final ExtensionContext context) throws Exception {
+		setDefaultLocale(context);
+	}
+
+	@Override
 	public void beforeEach(ExtensionContext context) throws Exception {
-		defaultLocale = Locale.getDefault();
+		setDefaultLocale(context);
+	}
+
+	private void setDefaultLocale(final ExtensionContext context) {
+		saveDefaultLocale();
 
 		Locale configuredLocale = readLocaleFromAnnotation(context);
 
 		Locale.setDefault(configuredLocale);
 	}
 
-	@Override
-	public void afterEach(ExtensionContext context) throws Exception {
-		Locale.setDefault(defaultLocale);
+	private void saveDefaultLocale() {
+		defaultLocale = Locale.getDefault();
 	}
 
 	private Locale readLocaleFromAnnotation(ExtensionContext context) {
@@ -57,5 +67,19 @@ public class DefaultLocaleExtension implements BeforeEachCallback, AfterEachCall
 		else {
 			return new Locale(annotation.language());
 		}
+	}
+
+	@Override
+	public void afterEach(ExtensionContext context) throws Exception {
+		resetDefaultLocale();
+	}
+
+	@Override
+	public void afterAll(final ExtensionContext context) throws Exception {
+		resetDefaultLocale();
+	}
+
+	private void resetDefaultLocale() {
+		Locale.setDefault(defaultLocale);
 	}
 }
--- a/src/main/java/org/junitpioneer/jupiter/DefaultLocaleExtension.java
+++ b/src/main/java/org/junitpioneer/jupiter/DefaultLocaleExtension.java
@@ -25,12 +25,12 @@ class DefaultLocaleExtension implements BeforeAllCallback, BeforeEachCallback, A
 	private Locale defaultLocale;
 
 	@Override
-	public void beforeAll(final ExtensionContext context) throws Exception {
+	public void beforeAll(final ExtensionContext context) {
 		setDefaultLocale(context);
 	}
 
 	@Override
-	public void beforeEach(ExtensionContext context) throws Exception {
+	public void beforeEach(ExtensionContext context) {
 		setDefaultLocale(context);
 	}
 
@@ -70,12 +70,12 @@ class DefaultLocaleExtension implements BeforeAllCallback, BeforeEachCallback, A
 	}
 
 	@Override
-	public void afterEach(ExtensionContext context) throws Exception {
+	public void afterEach(ExtensionContext context) {
 		resetDefaultLocale();
 	}
 
 	@Override
-	public void afterAll(final ExtensionContext context) throws Exception {
+	public void afterAll(final ExtensionContext context) {
 		resetDefaultLocale();
 	}
 
--- a/src/main/java/org/junitpioneer/jupiter/DefaultLocaleExtension.java
+++ b/src/main/java/org/junitpioneer/jupiter/DefaultLocaleExtension.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2015-2018 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junitpioneer.jupiter;
+
+import java.util.Locale;
+import java.util.Optional;
+
+import org.junit.jupiter.api.extension.AfterEachCallback;
+import org.junit.jupiter.api.extension.BeforeEachCallback;
+import org.junit.jupiter.api.extension.ExtensionContext;
+
+public class DefaultLocaleExtension implements BeforeEachCallback, AfterEachCallback {
+
+	private Locale defaultLocale;
+
+	@Override
+	public void beforeEach(ExtensionContext context) throws Exception {
+		defaultLocale = Locale.getDefault();
+
+		Locale configuredLocale = readLocaleFromAnnotation(context);
+
+		Locale.setDefault(configuredLocale);
+	}
+
+	@Override
+	public void afterEach(ExtensionContext context) throws Exception {
+		Locale.setDefault(defaultLocale);
+	}
+
+	private Locale readLocaleFromAnnotation(ExtensionContext context) {
+		Optional<DefaultLocale> annotation = context.getElement().flatMap(
+			elem -> Optional.of(elem.getAnnotation(DefaultLocale.class)));
+
+		return annotation.map(DefaultLocale::value).map(Locale::new).orElse(Locale.getDefault());
+	}
+}
--- a/src/main/java/org/junitpioneer/jupiter/DefaultLocaleExtension.java
+++ b/src/main/java/org/junitpioneer/jupiter/DefaultLocaleExtension.java
@@ -35,7 +35,17 @@ class DefaultLocaleExtension implements BeforeAllCallback, BeforeEachCallback, A
 
 	@Override
 	public void beforeEach(ExtensionContext context) {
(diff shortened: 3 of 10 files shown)


src/main/java/org/junitpioneer/jupiter/DefaultLocale.java
```diff
@@ -0,0 +1,26 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.junit.jupiter.api.extension.ExtendWith;
+
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+@ExtendWith(DefaultLocaleExtension.class)
+public @interface DefaultLocale {
```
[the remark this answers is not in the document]
sormuras: For me, `@DefaultLocale` is okay. It matches its getter: `Locale.getDefault()`

src/main/java/org/junitpioneer/jupiter/DefaultLocale.java
```diff
@@ -0,0 +1,26 @@
...
+package org.junitpioneer.jupiter;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.junit.jupiter.api.extension.ExtendWith;
+
+@Target(ElementType.METHOD)
```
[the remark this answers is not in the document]
sormuras: Yes, would be neat. Setting a default locale for all tests in a class makes sense. An annotation attached to a test method may override this class-level default locale, though.

src/main/java/org/junitpioneer/jupiter/DefaultLocale.java
```diff
@@ -0,0 +1,26 @@
...
+@Retention(RetentionPolicy.RUNTIME)
+@ExtendWith(DefaultLocaleExtension.class)
+public @interface DefaultLocale {
+
+	String value();
```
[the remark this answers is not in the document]
sormuras: Yes, all 3 fields should be supported and default to an empty string.

User may already provide bad input for the single argument case. So, same error handling applies to 3x bad input or illegal combinations.

src/test/java/org/junitpioneer/jupiter/DefaultLocaleTests.java
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	@DefaultLocale(language = "en", country = "EN", variant = "gb")
+	@Test
+	@DisplayName("sets the default locale using a language, a country and a variant")
+	void setsLanguageAndCountryAndVariant() {
+		assertEquals(new Locale("en", "EN", "gb"), Locale.getDefault());
+	}
+
+	@Disabled("TODO: How to test this?")
```
britter: I need some help here. How can I test that I expect an Exception to be thrown from the extension?

src/test/java/org/junitpioneer/jupiter/DefaultLocaleTests.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.Locale;
+
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+@DisplayName("DefaultLocale extension")
+class DefaultLocaleTests {
```
britter: I don't know how to test the class level annotation since I see no way of knowing which locale has been the default locale before the class executed. Any idea?

src/main/java/org/junitpioneer/jupiter/DefaultLocale.java
```diff
@@ -0,0 +1,50 @@
...
+ * country and a variant. If {@link #variant()} is set, but {@link #country()}
+ * is not, a
+ * {@link org.junit.jupiter.api.extension.ExtensionConfigurationException} will
+ * be thrown. After the annotated element has been executed, the default Locale
+ * will be restored to its original value.</p>
+ *
+ * <p>{@code @DefaultLocale} can be used on the method and on the class level.
+ * If a class is annotated, the configured Locale will be the default Locale
+ * for all tests inside that class. Any method level configurations will
+ * override the class level default Locale.</p>
```
britter: Does the documentation make sense?

src/main/java/org/junitpioneer/jupiter/DefaultLocale.java
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @since 0.2
+ * @see java.util.Locale#getDefault()
+ */
+@Target({ ElementType.METHOD, ElementType.TYPE })
+@Retention(RetentionPolicy.RUNTIME)
+@ExtendWith(DefaultLocaleExtension.class)
+public @interface DefaultLocale {
+
+	String language();
```
britter: I'd like to have `language()` as alias for `value()`. This is possible with `@AliasFor` from spring framework. Is there something comparable in JUnit?

src/main/java/org/junitpioneer/jupiter/DefaultLocale.java
sormuras: Sounds good to me.

src/main/java/org/junitpioneer/jupiter/DefaultLocale.java
sormuras: No. And their we won't  add `@AliasFor` support in the near future. There exists a ticket regarding that feature ...

sormuras: I'd prefer `value()`  over `language()` here. But that's debatable.

britter: I don't have a strong opinion on this. While `value()` is more concise, using `language()` hat the advantage that it follows the naming of the constructor parameters of java.util.Locale. It's your call ;-)

src/main/java/org/junitpioneer/jupiter/DefaultTimeZoneExtension.java
```diff
@@ -0,0 +1,68 @@
...
+		TimeZone.setDefault(configuredTimeZone);
+	}
+
+	private void saveDefaultTimeZone() {
+		defaultTimeZone = TimeZone.getDefault();
+	}
+
+	private TimeZone readTimeZoneFromAnnotation(ExtensionContext context) {
+		Optional<DefaultTimeZone> annotation = context.getElement().flatMap(
+			elem -> Optional.of(elem.getAnnotation(DefaultTimeZone.class)));
```
britter: This should probably be `Optional.ofNullable` otherwise this could lead to NPEs being thrown.

src/main/java/org/junitpioneer/jupiter/DefaultLocaleExtension.java
```diff
@@ -0,0 +1,85 @@
...
+		Locale.setDefault(configuredLocale);
+	}
+
+	private void saveDefaultLocale() {
+		defaultLocale = Locale.getDefault();
+	}
+
+	private Locale readLocaleFromAnnotation(ExtensionContext context) {
+		Optional<DefaultLocale> annotation = context.getElement().flatMap(
+			elem -> Optional.of(elem.getAnnotation(DefaultLocale.class)));
```
britter: This should probably be `Optional.ofNullable` otherwise this could lead to NPEs being thrown.

sormuras:
In order to support finding _inherited_ and _meta_ annotations the extension should use the utilities exposed here https://junit.org/junit5/docs/current/api/org/junit/platform/commons/support/AnnotationSupport.html

britter:
Done!

src/main/java/org/junitpioneer/jupiter/DefaultLocale.java
sbrannen: Since there are three `String` attributes in the annotation and the likelihood that one would need to supply more than just the language is rather high, I'd prefer `language` in this particular use case.

src/main/java/org/junitpioneer/jupiter/DefaultLocaleExtension.java
```diff
@@ -0,0 +1,86 @@
...
+import org.junit.jupiter.api.extension.AfterEachCallback;
+import org.junit.jupiter.api.extension.BeforeAllCallback;
+import org.junit.jupiter.api.extension.BeforeEachCallback;
+import org.junit.jupiter.api.extension.ExtensionConfigurationException;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.platform.commons.support.AnnotationSupport;
+
+class DefaultLocaleExtension implements BeforeAllCallback, BeforeEachCallback, AfterAllCallback, AfterEachCallback {
+
+	private Locale defaultLocale;
```
sbrannen: I have not tested this in an IDE, but storing this state in a field within the extension looks brittle.

I think this needs to be stored in the `ExtensionContext.Store` for the current execution level (i.e., class or method).

src/main/java/org/junitpioneer/jupiter/DefaultLocaleExtension.java
```diff
@@ -0,0 +1,86 @@
...
+
+	private Locale readLocaleFromAnnotation(ExtensionContext context) {
+		Optional<DefaultLocale> annotation = AnnotationSupport.findAnnotation(context.getElement(),
+			DefaultLocale.class);
+
+		return annotation.map(DefaultLocaleExtension::createLocale).orElse(Locale.getDefault());
+	}
+
+	private static Locale createLocale(DefaultLocale annotation) {
+		if (!annotation.country().isEmpty() && !annotation.variant().isEmpty()) {
```
sbrannen: What if the `language` is empty or blank?

src/main/java/org/junitpioneer/jupiter/DefaultTimeZoneExtension.java
```diff
@@ -0,0 +1,69 @@
...
+import org.junit.jupiter.api.extension.AfterAllCallback;
+import org.junit.jupiter.api.extension.AfterEachCallback;
+import org.junit.jupiter.api.extension.BeforeAllCallback;
+import org.junit.jupiter.api.extension.BeforeEachCallback;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.platform.commons.support.AnnotationSupport;
+
+class DefaultTimeZoneExtension implements BeforeAllCallback, BeforeEachCallback, AfterAllCallback, AfterEachCallback {
+
+	private TimeZone defaultTimeZone;
```
sbrannen: Same points as made for the `DefaultLocaleExtension`

src/main/java/org/junitpioneer/jupiter/DefaultTimeZoneExtension.java
```diff
@@ -0,0 +1,69 @@
...
+
+	private void saveDefaultTimeZone() {
+		defaultTimeZone = TimeZone.getDefault();
+	}
+
+	private TimeZone readTimeZoneFromAnnotation(ExtensionContext context) {
+		Optional<DefaultTimeZone> annotation = AnnotationSupport.findAnnotation(context.getElement(),
+			DefaultTimeZone.class);
+
+		return annotation.map(DefaultTimeZone::value).map(TimeZone::getTimeZone).orElse(TimeZone.getDefault());
```
sbrannen: What is `value` is empty or blank?

src/test/java/org/junitpioneer/jupiter/DefaultLocaleTests.java
sbrannen: You can execute the test class via the `Launcher` API, checking the existing default locale both before and after the execution.

src/test/java/org/junitpioneer/jupiter/DefaultLocaleTests.java
sbrannen: Write a true _unit test_ for the extension.

src/main/java/org/junitpioneer/jupiter/DefaultLocaleExtension.java
britter: I talked to this with @sormuras and we agreed on the simpler approach. But I can change it to `ExtensionContext.Store`. WDYT @sormuras?

src/main/java/org/junitpioneer/jupiter/DefaultLocaleExtension.java
britter: That value will be passed to the Locale constructor. I've just tested it:

```
new Locale("")
new Locale("  ")
```

Both are fine and don't throw exceptions, so I wouldn't be more restrictive here.

src/main/java/org/junitpioneer/jupiter/DefaultTimeZoneExtension.java
britter: see above

src/main/java/org/junitpioneer/jupiter/DefaultTimeZoneExtension.java
britter: see above

src/test/java/org/junitpioneer/jupiter/DefaultLocaleTests.java
britter: @sormuras was talking about a dedicated API which might make it to JUnit 5.4. My plan was to add these tests once that has been released, but I can also try to come up with something based on your proposal. Will work on this on the weekend.

src/test/java/org/junitpioneer/jupiter/DefaultLocaleTests.java
britter: Working on this on the weekend!

docs/default-locale-timezone.adoc
```diff
@@ -0,0 +1,93 @@
+:page-title: @DefaultLocale and @DefaultTimeZone
+:page-description: JUnit Jupiter extensions to change the values returned from Locale.getDefault() and TimeZone.getDefault()
+
+The `@DefaultLocale` and `@DefaultTimeZone` annotations can be used to change the values returned from `Locale.getDefault()` and `TimeZone.getDefault()`.
+Both annotations work on the test class level and on the test method level.
+After the annotated element has been executed the initial default value is restored.
+
+== `@DefaultLocale`
+
+The default `Locale` is specified according to the constructors of https://docs.oracle.com/javase/10/docs/api/java/util/Locale.html[Locale].
```
nicolaiparlog: Since we're targeting Java 8, can you link to https://docs.oracle.com/javase/8/docs/api/java/util/Locale.html instead?

Also, I think this is more explanatory:

    The default `Locale` is specified according to https://docs.oracle.com/javase/8/docs/api/java/util/Locale.html#constructor.summary[its constructors] - it needs either:
    * `language` or
    * `language` and `country` or
    * `language`, `country`, and `variant`

docs/default-locale-timezone.adoc
```diff
@@ -0,0 +1,93 @@
...
+}
+
+@Test
+@DefaultLocale(language = "en", country = "EN", variant = "gb")
+void test_with_lanuage_and_country_and_vairant() {
+    assertEquals(new Locale("en", "EN", "gb"), Locale.getDefault());
+}
+----
+
+Note that variant can only be specified when the country is also specified.
```
nicolaiparlog: Highlight code:

    Note that `variant` can only be specified if `country` is also specified.

docs/default-locale-timezone.adoc
```diff
@@ -0,0 +1,93 @@
...
+    @DefaultLocale(language = "en")
+    void test_with_method_level_configuration() {
+        assertEquals(new Locale("en"), Locale.getDefault());
+    }
+}
+----
+
+== `@DefaultTimeZone`
+
+The default `TimeZone` is specified according to the https://docs.oracle.com/javase/10/docs/api/java/util/TimeZone.html#getTimeZone(java.lang.String)[TimeZone.getTimeZone(String)] method.
```
nicolaiparlog: Link to Java 8 docs.

src/main/java/org/junitpioneer/jupiter/DefaultLocale.java
```diff
@@ -0,0 +1,51 @@
...
+import java.lang.annotation.Target;
+
+import org.junit.jupiter.api.extension.ExtendWith;
+
+/**
+ * {@code @DefaultLocale} is a JUnit Jupiter extension to change the value
+ * returned by {@link java.util.Locale#getDefault()} for a test execution.
+ *
+ * <p>The {@link java.util.Locale} to set as the default locale can be
+ * configured using only a language, a language and a country or a language, a
```
nicolaiparlog: Add Oxford comma for better readability:

    configured using only a language, a language and a country, or a language, a

docs/default-locale-timezone.adoc
```diff
@@ -0,0 +1,93 @@
...
+
+@Test
+@DefaultLocale(language = "en", country = "EN", variant = "gb")
+void test_with_lanuage_and_country_and_vairant() {
+    assertEquals(new Locale("en", "EN", "gb"), Locale.getDefault());
+}
+----
+
+Note that variant can only be specified when the country is also specified.
+If a variant is specified without a country an `ExtentionConfigurationException` will be thrown.
```
nicolaiparlog: If `variant` is specified without `country`, an `ExtentionConfigurationException` will be thrown.

src/main/java/org/junitpioneer/jupiter/DefaultLocale.java
```diff
@@ -0,0 +1,51 @@
...
+ * country and a variant. If {@link #variant()} is set, but {@link #country()}
+ * is not, a
+ * {@link org.junit.jupiter.api.extension.ExtensionConfigurationException} will
+ * be thrown. After the annotated element has been executed, the default Locale
```
nicolaiparlog: Either `{@code Locale}` or `locale` (note the capitalization).

src/main/java/org/junitpioneer/jupiter/DefaultLocale.java
```diff
@@ -0,0 +1,51 @@
...
+ * will be restored to its original value.</p>
+ *
+ * <p>{@code @DefaultLocale} can be used on the method and on the class level.
+ * If a class is annotated, the configured Locale will be the default Locale
```
nicolaiparlog: Capitalization.

src/main/java/org/junitpioneer/jupiter/DefaultLocale.java
```diff
@@ -0,0 +1,51 @@
...
+ * for all tests inside that class. Any method level configurations will
+ * override the class level default Locale.</p>
```
nicolaiparlog: Capitalization.

src/main/java/org/junitpioneer/jupiter/DefaultLocale.java
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * @since 0.2
```
nicolaiparlog: This will still go into `0.1.x` (right @sormuras?), so `@since 0.1` should be fine.

src/main/java/org/junitpioneer/jupiter/DefaultLocale.java
nicolaiparlog: I prefer `language`, too.

Isn't it common to write the locale in [snake case](https://en.wikipedia.org/wiki/Snake_case)? In that case, `value` could accept a string like `"en_EN_gb"` and take it apart. We don't need that from the get go, though, and if this is not totally harebrained, we should discuss it in a separate issue.

src/main/java/org/junitpioneer/jupiter/DefaultLocale.java
```diff
@@ -0,0 +1,51 @@
...
+@Target({ ElementType.METHOD, ElementType.TYPE })
+@Retention(RetentionPolicy.RUNTIME)
+@ExtendWith(DefaultLocaleExtension.class)
+public @interface DefaultLocale {
+
+	String language();
+
+	String country() default "";
+
+	String variant() default "";
```
nicolaiparlog: The `Locale` constructor has this to say about the parameters:

    language - An ISO 639 alpha-2 or alpha-3 language code, or a language subtag up to 8 characters in length. See the Locale class description about valid language values.
    country - An ISO 3166 alpha-2 country code or a UN M.49 numeric-3 area code. See the Locale class description about valid country values.
    variant - Any arbitrary value used to indicate a variation of a Locale. See the Locale class description for the details.

I think it would be nice to add these as comments to the attributes.

src/main/java/org/junitpioneer/jupiter/DefaultLocaleExtension.java
nicolaiparlog: I was taught (and am teaching people) to always use the extension store as there is no guarantee for an extension instance's lifecycle. In case I'm missing something, I'd also prefer the store.

src/main/java/org/junitpioneer/jupiter/DefaultLocaleExtension.java
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	@Override
+	public void beforeEach(ExtensionContext context) {
+		setDefaultLocale(context);
+	}
+
+	private void setDefaultLocale(final ExtensionContext context) {
+		saveDefaultLocale();
+
```
nicolaiparlog: Why

the

empty

lines?

:wink:

src/main/java/org/junitpioneer/jupiter/DefaultLocaleExtension.java
```diff
@@ -0,0 +1,86 @@
...
+import org.junit.jupiter.api.extension.ExtensionConfigurationException;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.platform.commons.support.AnnotationSupport;
+
+class DefaultLocaleExtension implements BeforeAllCallback, BeforeEachCallback, AfterAllCallback, AfterEachCallback {
+
+	private Locale defaultLocale;
+
+	@Override
+	public void beforeAll(final ExtensionContext context) {
```
nicolaiparlog: I prefer not to use `final` on method parameters. Not because I think reassigning parameters is a good idea (quite the opposite!), but I don't like how `final` clutters the code.

src/main/java/org/junitpioneer/jupiter/DefaultLocaleExtension.java
```diff
@@ -0,0 +1,86 @@
...
+		setDefaultLocale(context);
+	}
+
+	@Override
+	public void beforeEach(ExtensionContext context) {
+		setDefaultLocale(context);
```
nicolaiparlog: If the extension is applied to the class, this is called before each test method and does "something" (the operation results in no state change, but that's still not ideal). I'd prefer to guard this by checking whether the annotation is present on the test method with `AnnotationSupport::isAnnotated`.

src/main/java/org/junitpioneer/jupiter/DefaultLocaleExtension.java
```diff
@@ -0,0 +1,86 @@
...
+
+	private void saveDefaultLocale() {
+		defaultLocale = Locale.getDefault();
+	}
+
+	private Locale readLocaleFromAnnotation(ExtensionContext context) {
+		Optional<DefaultLocale> annotation = AnnotationSupport.findAnnotation(context.getElement(),
+			DefaultLocale.class);
+
+		return annotation.map(DefaultLocaleExtension::createLocale).orElse(Locale.getDefault());
```
nicolaiparlog: With the `isAnnotated` guard from above it would be an error if the element were not annotated, so I would throw an `ExtensionConfigurationException`. I also think this can be one pipeline:

	//@formatter:off
	return AnnotationSupport
			.findAnnotation(context.getElement(), DefaultLocale.class)
			.map(DefaultLocaleExtension::createLocale)
			.orElseThrow(() -> new ExtensionConfigurationException("TODO"));
	//@formatter:on

src/main/java/org/junitpioneer/jupiter/DefaultLocaleExtension.java
nicolaiparlog: Agree with @britter.

src/main/java/org/junitpioneer/jupiter/DefaultLocaleExtension.java
```diff
@@ -0,0 +1,86 @@
...
+				"@DefaultLocale.country must not be empty when @DefaultLocale.variant is set!");
+		}
+		else {
+			return new Locale(annotation.language());
+		}
+	}
+
+	@Override
+	public void afterEach(ExtensionContext context) {
+		resetDefaultLocale();
```
nicolaiparlog: If `beforeEach` is guarded, this should be, too.

src/main/java/org/junitpioneer/jupiter/DefaultTimeZone.java
```diff
@@ -0,0 +1,45 @@
...
+ * {@link java.util.TimeZone#getTimeZone(String)}. After the annotated element
+ * has been executed, the default TimeZone will be restored to its original
+ * value.</p>
+ *
+ * <p>{@code @DefaultTimeZone} can be used on the method and on the class
+ * level. If a class is annotated, the configured TimeZone will be the default
+ * TimeZone for all tests inside that class. Any method level configurations
+ * will override the class level default TimeZone.</p>
+ *
+ * @since 0.2
```
nicolaiparlog: See above.

src/main/java/org/junitpioneer/jupiter/DefaultTimeZone.java
```diff
@@ -0,0 +1,45 @@
...
+ * configured bei specifying the TimeZone ID as defined by
+ * {@link java.util.TimeZone#getTimeZone(String)}. After the annotated element
+ * has been executed, the default TimeZone will be restored to its original
+ * value.</p>
+ *
+ * <p>{@code @DefaultTimeZone} can be used on the method and on the class
+ * level. If a class is annotated, the configured TimeZone will be the default
+ * TimeZone for all tests inside that class. Any method level configurations
+ * will override the class level default TimeZone.</p>
+ *
```
nicolaiparlog: Capitalization - it's either `{@code TimeZone}` or `time zone`.

src/main/java/org/junitpioneer/jupiter/DefaultTimeZoneExtension.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.TimeZone;
+
+import org.junit.jupiter.api.extension.AfterAllCallback;
+import org.junit.jupiter.api.extension.AfterEachCallback;
+import org.junit.jupiter.api.extension.BeforeAllCallback;
+import org.junit.jupiter.api.extension.BeforeEachCallback;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.platform.commons.support.AnnotationSupport;
+
+class DefaultTimeZoneExtension implements BeforeAllCallback, BeforeEachCallback, AfterAllCallback, AfterEachCallback {
```
nicolaiparlog: Same comments as above:

* guard `before`/`afterEach`
* remove empty lines in method bodies
* throw exception in `readTimeZoneFromAnnotation`

src/test/java/org/junitpioneer/jupiter/DefaultLocaleTests.java
nicolaiparlog: These tests can be written like the ones in `TestIntegrationTests`.

src/test/java/org/junitpioneer/jupiter/DefaultLocaleTests.java
```diff
@@ -0,0 +1,81 @@
...
+
+	@BeforeAll
+	static void globalSetUp() {
+		DEFAULT_LOCALE_BEFORE_TEST = Locale.getDefault();
+		if (!DEFAULT_LOCALE_BEFORE_TEST.equals(Locale.CANADA)) {
+			Locale.setDefault(Locale.CANADA);
+		}
+		else {
+			// you seem to be from Canada...
+			Locale.setDefault(Locale.CHINESE);
```
nicolaiparlog: This took me a while. :joy:

src/main/java/org/junitpioneer/jupiter/DefaultLocale.java
sormuras: ShipKit only increments `BUILD` in `MAJOR.MINOR.BUILD`. So, yes, it should be `0.1.x` and agree to mention it as `@since 0.1` here.

src/main/java/org/junitpioneer/jupiter/DefaultLocaleExtension.java
sormuras: _"Use the Store, Luke!"_

src/test/java/org/junitpioneer/jupiter/DefaultLocaleTests.java
nicolaiparlog: Even better, have a look at [`RepeatFailedTestTests`](https://github.com/junit-pioneer/junit-pioneer/pull/110/files#diff-94a60a248ed1fcd2fa7028a25ea6664e).

src/main/java/org/junitpioneer/jupiter/DefaultLocale.java
sdavids: Another thing you might consider:

https://docs.oracle.com/javase/8/docs/api/java/util/Locale.html#forLanguageTag-java.lang.String-

`@DefaultLocale(languageTag = "en-US")`

would be more succinct than

`@DefaultLocale(language = "en", country="US")`

src/main/java/org/junitpioneer/jupiter/DefaultLocale.java
britter: Okay, so no SemVer? This is a new feature, so I expected a minor version bump.

nicolaiparlog: You're right. I'm inclined to make the next one `0.2` - what do you think @sormuras? And what file does that go into? (Can't check right now.)

@britter We're doing this "every commit to master triggers a release"-thing (see #87), so we have to make the version bump part of this commit.

sormuras: Edit the `version` in `version.properties` to the next version you want.

https://github.com/junit-pioneer/junit-pioneer/blob/961177d5c4cd99f220870c96658f17a769137807/version.properties#L3

Never(!) set it to a previous version, though.

sormuras: Having said that, increasing the minor version on each commit to `master` should be documented (and enforced?) somewhere.

britter:
Resolved all discussions expect @sdavids comment about language tag. @nicolaiparlog @sbrannen @sormuras should we add support for specifying the language tag as well?

src/main/java/org/junitpioneer/jupiter/DefaultLocale.java
nicolaiparlog: @sormuras Not every commit, though, right? See #113.

nicolaiparlog:
Let's discuss and possibly implement `language-tag` in a different issue/PR. @britter Can you create an issue for that?

src/main/java/org/junitpioneer/jupiter/DefaultLocaleExtension.java
```diff
@@ -0,0 +1,100 @@
...
+	private void storeDefaultLocale(ExtensionContext context) {
+		context.getStore(NAMESPACE).put(KEY, Locale.getDefault());
+	}
+
+	private Locale readLocaleFromAnnotation(ExtensionContext context) {
+		//@formatter:off
+		return AnnotationSupport
+				.findAnnotation(context.getElement(), DefaultLocale.class)
+				.map(DefaultLocaleExtension::createLocale)
+				.orElseThrow(() -> new ExtensionConfigurationException("TODO"));
```
nicolaiparlog: Mmh, maybe `"The extension is active, but the corresponding annotation could not be found. (This may be a bug.)"` ?

src/main/java/org/junitpioneer/jupiter/DefaultTimeZoneExtension.java
```diff
@@ -0,0 +1,85 @@
...
+		context.getStore(NAMESPACE).put(KEY, TimeZone.getDefault());
+	}
+
+	private TimeZone readTimeZoneFromAnnotation(ExtensionContext context) {
+		//@formatter:off
+		return AnnotationSupport
+				.findAnnotation(context.getElement(), DefaultTimeZone.class)
+				.map(DefaultTimeZone::value)
+				.map(TimeZone::getTimeZone)
+				.orElseThrow(() -> new ExtensionConfigurationException("TODO"));
```
nicolaiparlog: See above.

src/main/java/org/junitpioneer/jupiter/DefaultLocaleExtension.java
```diff
@@ -0,0 +1,100 @@
...
+		if (annotationPresentOnTestMethod(context)) {
+			setDefaultLocale(context);
+		}
+	}
+
+	private boolean annotationPresentOnTestMethod(ExtensionContext context) {
+		//@formatter:off
+		return context.getTestMethod()
+				.map(testMethod -> AnnotationSupport.isAnnotated(testMethod, DefaultLocale.class))
+				.orElse(false);
```
nicolaiparlog: I would have written it like this:

```java
return context.getTestMethod()
		.filter(testMethod -> AnnotationSupport.isAnnotated(testMethod, DefaultLocale.class))
		.isPresent;
```

Funny, how similar and yet different the pipelines. :smile: (No need to change this.)

src/main/java/org/junitpioneer/jupiter/DefaultLocaleExtension.java
```diff
@@ -0,0 +1,100 @@
...
+		}
+	}
+
+	@Override
+	public void afterAll(ExtensionContext context) {
+		resetDefaultLocale(context);
+	}
+
+	private void resetDefaultLocale(ExtensionContext context) {
+		Locale.setDefault((Locale) context.getStore(NAMESPACE).get(KEY));
```
nicolaiparlog: `Store::get` has an overload that takes the desired type and does the cast for you.

src/main/java/org/junitpioneer/jupiter/DefaultTimeZoneExtension.java
```diff
@@ -0,0 +1,85 @@
...
+		}
+	}
+
+	@Override
+	public void afterAll(ExtensionContext context) {
+		resetDefaultTimeZone(context);
+	}
+
+	private void resetDefaultTimeZone(ExtensionContext context) {
+		TimeZone.setDefault((TimeZone) context.getStore(NAMESPACE).get(KEY));
```
nicolaiparlog: See above.

src/test/java/org/junitpioneer/jupiter/DefaultLocaleTests.java
```diff
@@ -0,0 +1,152 @@
...
+@DisplayName("DefaultLocale extension")
+class DefaultLocaleTests extends AbstractPioneerTestEngineTests {
+
+	private static Locale TEST_DEFAULT_LOCALE;
+	private static Locale DEFAULT_LOCALE_BEFORE_TEST;
+
+	@BeforeAll
+	static void globalSetUp() {
+		DEFAULT_LOCALE_BEFORE_TEST = Locale.getDefault();
+		TEST_DEFAULT_LOCALE = new Locale("custom");
```
nicolaiparlog: This is smoother than before, but I miss Canada and China. :laughing:

src/test/java/org/junitpioneer/jupiter/DefaultLocaleTests.java
```diff
@@ -0,0 +1,152 @@
...
+	@Nested
+	@DisplayName("when configured incorrect")
+	class ConfigurationFailureTests {
+
+		@Test
+		@DisplayName("should fail when variant is set but country is not on method level")
+		void shouldFailWhenVariantIsSetButCountryIsNotOnMethodLevel() {
+			ExecutionEventRecorder eventRecorder = executeTestsForClass(MethodLevelInitializationFailureTestCase.class);
+
+			assertEquals(1, eventRecorder.getTestFailedCount());
```
nicolaiparlog: I wonder whether there would be much value in asserting the concrete error message or at least the exception type. What do you think?

src/test/java/org/junitpioneer/jupiter/DefaultTimeZoneTests.java
```diff
@@ -0,0 +1,113 @@
...
+	static void globalSetUp() {
+		DEFAULT_TIMEZONE_BEFORE_TEST = TimeZone.getDefault();
+		TimeZone utc = TimeZone.getTimeZone("UTC");
+		if (!DEFAULT_TIMEZONE_BEFORE_TEST.equals(utc)) {
+			TimeZone.setDefault(utc);
+		}
+		else {
+			TimeZone.setDefault(TimeZone.getTimeZone("GMT"));
+		}
+		TEST_DEFAULT_TIMEZONE = TimeZone.getDefault();
```
nicolaiparlog: I find this a little clearer (clarifies roles of UTC and GMT; no double-negative for else branch):

```java
// the extension sets UTC as test time zone unless it is already
// the system's time zone; in that case it uses GMT
DEFAULT_TIMEZONE_BEFORE_TEST = TimeZone.getDefault();
TimeZone utc = TimeZone.getTimeZone("UTC");
TimeZone gmt = TimeZone.getTimeZone("GMT");
if (DEFAULT_TIMEZONE_BEFORE_TEST.equals(utc)) {
	TimeZone.setDefault(gmt);
}
else {
	TimeZone.setDefault(utc);
}
TEST_DEFAULT_TIMEZONE = TimeZone.getDefault();
```

nicolaiparlog:
To merge this, we need a message for the squashed commit. @britter, do you want to do that? (If so, [here's how](https://github.com/junit-pioneer/junit-pioneer/blob/master/CONTRIBUTING.md#merging); if not, I can do that.)

src/test/java/org/junitpioneer/jupiter/DefaultLocaleTests.java
britter: If think assertion the exception type is a good idea. I've added that. Checking exception messages is not a good idea IMHO, since that can change over time to provide more context and there is really no value in checking that text.

OTOH this is now a catch all for `ExtensionConfigurationExceptions`... so maybe it would be a good idea to check the test. I'm indifferent here. I think I'm fine with checking the type for now.

sormuras:
`#`103 
`#`104

2x`#` are missing.

nicolaiparlog:
Yeah, done! :tada: Let's let this sit for a day or two, to give @sbrannen a chance to accept the changes, and then merge it at the beginning of the week. Thanks for the contribution and the diligent work. :100:

britter:
Awesome! I'm really happy with the solution. Thanks for all the feedback.

sbrannen:
I'm trusting the other reviewers and have therefore _approved the changes_. 😉

src/main/java/org/junitpioneer/jupiter/DefaultLocaleExtension.java
```diff
@@ -0,0 +1,100 @@
...
+
+	@Override
+	public void beforeAll(ExtensionContext context) {
+		setDefaultLocale(context);
+	}
+
+	@Override
+	public void beforeEach(ExtensionContext context) {
+		if (annotationPresentOnTestMethod(context)) {
+			setDefaultLocale(context);
```
sbrannen: Note that the invocation of both of these methods actually results in the annotation being looked up twice, which seems a bit odd to me. Though I suppose it's not much of a performance hit. 😇

nicolaiparlog: We can have `annotationPresentOnTestMethod` return an `Optional<$whatever>` and then act on that. :smiley:

britter: Created #115 to implement this optimization

Issue #100, opened 2018-07-31
TempDirectory is not cleaning himself after tests are finished


JorisPotier:
Hi,

I try to use the TempDirectory extension in order to work to a clean workspace for each test, and finally delete this workspace. I'm using a custom FS (Jimfs) exactly like in this sample extracted from the docs : 

```
class MyTests {
    private static FileSystem fileSystem;

    @BeforeAll
    static void createFileSystem() {
        fileSystem = Jimfs.newFileSystem();
    }

    @AfterAll
    static void closeFileSystem() throws Exception {
        fileSystem.close();
    }

    @RegisterExtension
    Extension tempDirectory = TempDirectory.createInCustomDirectory(() ->
        Files.createDirectories(fileSystem.getPath("tmp")));

    @Test
    void test(@TempDir Path tempDir) {
        Path file = tempDir.resolve("test.txt");
        writeFile(file);
        assertExpectedFileContent(file);
    }
}
```

My problem is that the output directory is not deleted (and also the files contained in it, because my tests successfully write in it) by himself after the tests are done. Does this operation have to be handled manually in a `@AfterEach`/`@AfterAll` method ?

Thanks for your answers.

nicolaiparlog:
Looking at [the code](https://github.com/junit-pioneer/junit-pioneer/blob/961177d5c4cd99f220870c96658f17a769137807/src/main/java/org/junitpioneer/jupiter/TempDirectory.java#L262-L288) it seems like files and the directory should be deleted once the test using it finishes.

@marcphilipp Any idea what is going on here?

marcphilipp:
Given the following slightly modified version of your test class, both `test.txt` and the parent folder created by the `TempDirectory` extension are deleted.

```java

class MyTests {
    private static FileSystem fileSystem;

    @BeforeAll
    static void createFileSystem() {
        fileSystem = Jimfs.newFileSystem();
    }

    @AfterAll
    static void closeFileSystem() throws Exception {
        assertEquals(0, Files.list(fileSystem.getPath("tmp")).count());
        fileSystem.close();
    }

    @RegisterExtension
    Extension tempDirectory = TempDirectory.createInCustomDirectory(() ->
            Files.createDirectories(fileSystem.getPath("tmp")));

    @Test
    void test(@TempDir Path tempDir) throws IOException {
        Path file = tempDir.resolve("test.txt");
        Files.write(file, List.of("foo"));
[...]

PR #111, opened 2018-09-22
Ensure temp directory gets deleted

sdavids:
If Files#delete throws an IOException the remaining paths will not get deleted.

By additionally using File#file.deleteOnExit we ensure the paths will be deleted on JVM termination.

Note: Abnormal JVM termination might still leave some files on the file system—unfortunately we can do nothing about that.

Issue: #100

---

### DIFF
--- a/src/main/java/org/junitpioneer/jupiter/TempDirectory.java
+++ b/src/main/java/org/junitpioneer/jupiter/TempDirectory.java
@@ -13,6 +13,7 @@ package org.junitpioneer.jupiter;
 import static java.nio.file.FileVisitResult.CONTINUE;
 import static java.util.Objects.requireNonNull;
 
+import java.io.File;
 import java.io.IOException;
 import java.lang.annotation.Documented;
 import java.lang.annotation.ElementType;
@@ -253,6 +254,7 @@ public class TempDirectory implements ParameterResolver {
 
 		CloseablePath(Path dir) {
 			this.dir = dir;
+			deleteOnExit(dir);
 		}
 
 		Path get() {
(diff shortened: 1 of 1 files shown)


src/main/java/org/junitpioneer/jupiter/TempDirectory.java
```diff
@@ -286,5 +302,21 @@ private FileVisitResult deleteAndContinue(Path path) throws IOException {
...
+		static void deleteOnExit(Path path) {
+			if (path == null) {
+				return;
+			}
+			try {
+				File file = path.toFile();
+				if (file == null) {
+					return;
+				}
+				file.deleteOnExit();
```
jbduncan: Hmm. I don't think we can assume that `Path`s will always be translatable to `File`s, because the user may pass in a filesystem other than the default disk-based one e.g. Jimfs. 🤔

jbduncan: Oh sorry, I see that you already take care of this in the `catch` block below. Please ignore my earlier review/comment. :)

nicolaiparlog:
@marcphilipp Do you have time to take a look?

nicolaiparlog:
Closed in favor of #112.

nicolaiparlog:
@sdavids Thank you very much for giving this a shot (even though we didn't end up merging it), I appreciate it! :smiley: If you're looking for ways to contribute, check the [_up for grabs_ label](https://github.com/junit-pioneer/junit-pioneer/issues?q=is%3Aissue+is%3Aopen+label%3Aup-for-grabs) - it contains a few interesting idea that you could implement. If any other issue is more interesting to you, leave a comment on it and we can see whether you can tackle that. If you're interested in something, but aren't sure how to tackle it, ask away. Someone will help you out (but it may take a while :wink:).

PR #117, opened 2018-10-05
Upgrade Gradle wrapper to the latest version

britter:
---


nicolaiparlog:
The wrapper version is not updated in `.infra/gradle/gradle-wrapper.properties`, but in `build.gradle`. To apply the changes, delete `.infra/gradle` and run `gradle wrapper`. This will automatically set `gradle-wrapper.properties` and update the JAR.

marcphilipp:
I'd recommend that we remove the `version` from `build.gradle` as its unnecessary and duplicates what's in `gradle-wrapper.properties`.

britter:
The jar does not need to be updated. It is just some helper code that is used to download the gradle distribution.

I agree with @marcphilipp the wrapper configuration in `build.gradle` is not needed. Usually one uses `gradle init` to initialize a gradle build with wrapper and then once a new gradle version is available `./gradlew wrapper --gradle-version 4.10.2` is used to upgrade the properties file.

So, should I remove the wrapper configuration as part of this PR?

nicolaiparlog:
I'm no Gradle user, so I'll trust you two. :smiley: Then, yes, please remove the configuration.

marcphilipp:
We could just remove the version and keep the config for the location. However, I agree that we should rather switch to the default one.

Regarding not updating the wrapper jar: While that works in most cases, it is recommended to always update it along with the properties.

nicolaiparlog:
I don't like cluttering the top level folder with all of these configurations, artifacts, what-have-you (Travis, Checkstyle, Formatter, Shipkit, ...) and tried to ban them into `.infra` wherever possible. (AFAIK this is not yet possible for Shipkit, see mockito/shipkit#762.) So please keep the location for the JAR as is.

This leaves us with two options:

1. having a non-default place to define the wrapper version (in `build.gradle`)
2. having the wrapper configuration spread across two files

Personally, I find the latter more annoying than the former, but that may only come from me not knowing Gradle well and thus not having any expectation where I'd go looking for that config. If the Gradle buffs prefer 2., I'm ok with that.

marcphilipp:
The configuration for the `wrapper` task is in `build.gradle`, the configuration for the Gradle wrapper (gradlew) in `gradle-wrapper.properties`. Executing the former creates the latter. So, while closely related, they are actually two different things. Thus, I think 2. is ok.

britter:
I've removed the gradle version from the wrapper task configuration. Furthermore I've updated the gradle-wrapper.jar. From my PoV this can be merged now.

Squash message:

> Update gradle to the latest version (4.10.2)
>
> This includes updating the gradle-wrapper.jar to the latest version. Furthermore
> the gradleVersion has been removed from the wrapper task configuration, since it
> is already defined in gradle-wrapper.properties.
>
> PR: #117

nicolaiparlog:
Both  builds couldn't download Spotless, but works on my machine - weird. Restarted them before investigating.

nicolaiparlog:
There you go...

Issue #68, opened 2018-05-02
@CartesianProductTest extension


sormuras:
## Overview

Provide a similar, but complete and fleshed-out, extension that provides a `@CartesianProductTest` annotation implemention.

### Usage outline

```java
@CartesianProductTest({"0", "1"})
void threeBits(String a, String b, String c) {
	int value = Integer.parseUnsignedInt(a + b + c, 2);
	assertTrue((0b000 <= value) && (value <= 0b111));
}

@CartesianProductTest
@DisplayName("S ⨯ T ⨯ U")
void nFold(String string, Class<?> type, TimeUnit unit, TestInfo info) {
	assertTrue(string.endsWith("a"));
	assertTrue(type.isInterface());
	assertTrue(unit.name().endsWith("S"));
	assertTrue(info.getTags().isEmpty());
}

static CartesianProductTest.Sets nFold() {
	return new CartesianProductTest.Sets()
			.add("Alpha", "Omega")
			.add(Runnable.class, Comparable.class, TestInfo.class)
			.add(TimeUnit.DAYS, TimeUnit.HOURS);
}
```

### Implementation idea

https://github.com/junit-team/junit5-samples/tree/master/junit5-jupiter-extensions/src/main/java/com/example/cartesian

### Todos

- [ ] Copy sources from `junit5-jupiter-extensions` and adopt to pioneers' (package) structure.
- [ ] Write unit and integration tests for the `@CartesianProductTest` and its provider.
- [ ] Support more basic types in the annotation, in addition to `String`?
- [ ] Document usage and limitations

sormuras:
Ideally the `@CartesianProductTest` would be able to reuse any existing argument source, including the `@RangeSource` described in #123. For example:

```java
@CartesianProductTest
@EnumSource(TimeUnit.class)
@ValueSource(ints = {1, 2, 3})
@RangeSource(2, 10, 2)
void test(TimeUnit unit, int a, int b) {...}
```

marcphilipp:
The method would be called with {0, 1} × {0, 1} × {0, 1} = [(0,0,0), (0,0,1), ..., (1,1,1)].

matthieu-vergne:
I see. It seems to me quite specific and unnatural, though. I would prefer to write it with something more explicit, like:
```java
enum BinaryString {"0", "1"}

@CartesianProductTest
@EnumSource(BinaryString.class)
@EnumSource(BinaryString.class)
@EnumSource(BinaryString.class)
[...]

PR #123, opened 2018-10-05
Add @RangeSource argument source

### LINKED ISSUE #44: Add a RangeSource annotation for junit-jupiter-params
See https://github.com/junit-team/junit5/issues/851#issuecomment-300494134

mureinik:
This PR adds a series of argument source that provide ranges of primitieve numerical values to be used with JUnit Jupiter's `@ParameterizedTest`.
The annotations are `@ByteRangeSource`, `@ShortRangeSource`, `@IntRangeSource`, `@LongRangeSource`, `@FloatRangeSource` and `@DoubleRangeSource`.

All these annotations follow the same form:

    @IntRangeSource(to, from, step, closed)

If `step` is omitted, a default of `1` is taken.
If `closed` is omitted, a default of `false` is taken.

Closes #44
Related to junit-team/junit5#581

---


ris58h:
Great job! A couple of thing I would like to suggest:
1. Add `boolean closed() default false` parameter to the annotation to control inclusive/exclusive behavior of `to`  parameter. See Java's `IntStream.range` and `IntStream.rangeClosed`.
2. Support not only `int` but `long` values as well.

smoyer64:
I think the right way to support ``long`` values is to use long parameters.  If the method invocation asks for an ``int`` then range check it and either throw an exception or cast it.

src/main/java/org/junitpioneer/jupiter/RangeSource.java
```diff
@@ -0,0 +1,48 @@
...
+import java.lang.annotation.Target;
+
+import org.junit.jupiter.params.provider.ArgumentsSource;
+
+/**
+ * {@code @RangeSource} is an {@link ArgumentsSource} which provides access to a range of {@code int} values.
+ *
+ * The supplied values will be provided as arguments to the annotated {@code @ParameterizedTest} method.
+ *
+ * @see ArgumentsSource
```
nicolaiparlog: Once we know which version this lands in, we need a `@since` tag.

src/main/java/org/junitpioneer/jupiter/RangeSourceProvider.java
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * Provides a range of {@code int}s, as defined by {@link RangeSource}.
+ *
+ * @see RangeSource
+ */
+public class RangeSourceProvider implements ArgumentsProvider, AnnotationConsumer<RangeSource> {
+	/**
+	 * The beginning of the range, inclusive.
+	 */
+	private int[] values;
```
nicolaiparlog: [Extensions have to be stateless](https://blog.codefx.org/design/architecture/junit-5-extension-model/#Stateless), so we can't use a field here. I recommend to compute the values in `provideArguments`.

src/main/java/org/junitpioneer/jupiter/RangeSource.java
```diff
@@ -0,0 +1,48 @@
...
+ * @see org.junit.jupiter.params.ParameterizedTest
+ */
+@Target({ ElementType.ANNOTATION_TYPE, ElementType.METHOD })
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@ArgumentsSource(RangeSourceProvider.class)
+public @interface RangeSource {
+	/**
```
nicolaiparlog: Can you add an empty line at the beginning of each class? (Weird that the formatter doesn't enforce this).

src/test/java/org/junitpioneer/jupiter/RangeSourceTests.java
```diff
@@ -0,0 +1,104 @@
...
+ * A test for the {@link RangeSource} annotation and the underlying {@link RangeSourceProvider}.
+ */
+public class RangeSourceTests {
+	private static final List<Integer> allValues = new ArrayList<>();
+
+	/**
+	 * Fill up {@code allValues} with some values.
+	 * These values will be removed as {@link #ascending(int)} and {@link #descending(int)} receive their arguments,
+	 * and at the end of the test suite, {@link #assertAllValuesSupplied()} will check that all the values were
+	 * properly provided.
```
nicolaiparlog: Hah, this is very creative. :D

I'd still prefer it a little more formal, though. Can you have a look at, for example, [`DefaultLocaleTests`](https://github.com/junit-pioneer/junit-pioneer/blob/master/src/test/java/org/junitpioneer/jupiter/DefaultLocaleTests.java#L31) to see how to test extensions? It follows these steps:

* create an inner test class with the setup you want to test; these look very similar to how the user's tests will eventually look
* in _your_ tests, ask `AbstractPioneerTestEngineTests::executeTestsForClass` to execute these inner test classes
* assert proper execution

This includes misconfigurations.

nicolaiparlog:
Re `boolean closed() default false`, I think that's a great idea. @mureinik, you said you added this, but I didn't see it in the PR. Forgot to push? :smiley: (Happens to me all the time...)

Re `int` vs `long` (vs `double` vs `float`) - I see these options:

1. `@RangeSource` with `fromInt`, `fromLong` (`fromFloat`, `fromDouble`)
2. `@RangeSource` with `from` that accepts `long` and, if required, converts to `int` after range check (what about `float`/`double`, then?)
3. `@IntRangeSource`, `@LongRangeSource` (`@FloatRangeSource`, `@DoubleRangeSource`), each with `from`

My thoughts:

1. A little more verbose for simple cases, but I like its one-size-fits-all character. Also, it could be extended to support several parameters as long as they are of different types.
2. I liked it at first, but it doesn't seem to extend naturally to `float`/`double`, which is unfortunate.
3. Very clear cut, I like it.

What are your thoughts?

PS: the whole "covering several parameters" thingy reminds me of #68. Maybe @sormuras has something to add?

mureinik:
@nicolaiparlog 
WRT `closed()` - indeed, I forgot to push. How embarrassing :-( 
WRT your other feedback - first of all, thanks! I'll review and address it in the next couple of days

src/main/java/org/junitpioneer/jupiter/RangeSourceProvider.java
sbrannen: Actually, in contrast to an `Extension`, an `ArgumentsProvider` that implements `AnnotationConsumer` is (always?) stateful. 😉

src/main/java/org/junitpioneer/jupiter/RangeSource.java
```diff
@@ -0,0 +1,53 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.junit.jupiter.params.provider.ArgumentsSource;
+
+/**
+ * {@code @RangeSource} is an {@link ArgumentsSource} which provides access to a range of {@code int} values.
+ *
+ * The supplied values will be provided as arguments to the annotated {@code @ParameterizedTest} method.
```
sbrannen:
```suggestion
 * <p>The supplied values will be provided as arguments to the annotated {@code @ParameterizedTest} method.
```

src/main/java/org/junitpioneer/jupiter/RangeSourceProvider.java
nicolaiparlog: Oh, have to catch up on that. :blush: @mureinik: Nevermind, then.

sormuras:
> PS: the whole "covering several parameters" thingy reminds me of #68. Maybe @sormuras has something to add?

No, this idea and PR are looking good to me. And as "regular" argument source it could be used in a `@CartesianProductTest`, soon...

nicolaiparlog:
@sormuras That makes sense.

@mureinik Are you still working on this? Do you want to make any changes regarding the different data types?

mureinik:
@nicolaiparlog sorry again for the apparent ghosting, the day job caught up to me. I'll carve out some time for this in the next couple of days.

nicolaiparlog:
Don't hurry because of me, I'm notoriously laggy in replying. :wink: Do it in your own time - I just wanted to know whether you still want to do this at all.

src/main/java/org/junitpioneer/jupiter/RangeSource.java
mureinik: I'll add a `@since 4.0` annotation as a placeholder in the meanwhile

src/main/java/org/junitpioneer/jupiter/RangeSource.java
mureinik: Sure thing

mureinik:
Pushed some minor fixes just so I won't lose them. Still need to fix the tests as per the given feedback and think about the type issue

mureinik:
Updated the patch to address feedback given here, except for the multiple types issues. I wanted to double check I was heading the right direction (especially WRT writing the tests) before I go ahead and extend the functionality.

> Re `int` vs `long` (vs `double` vs `float`) - I see these options:
> 
> 1. `@RangeSource` with `fromInt`, `fromLong` (`fromFloat`, `fromDouble`)
> 2. `@RangeSource` with `from` that accepts `long` and, if required, converts to `int` after range check (what about `float`/`double`, then?)
> 3. `@IntRangeSource`, `@LongRangeSource` (`@FloatRangeSource`, `@DoubleRangeSource`), each with `from`
> 
> My thoughts:
> 
> 1. A little more verbose for simple cases, but I like its one-size-fits-all character. Also, it could be extended to support several parameters as long as they are of different types.
> 2. I liked it at first, but it doesn't seem to extend naturally to `float`/`double`, which is unfortunate.
> 3. Very clear cut, I like it.

2 - I'm not a fan of this option. Unless I'm missing something, JUnit won't automatically handle these casts for you, so now the provider needs to inspect the method that's annotated with it, and it add a whole new dimension of complexcity

1 vs 3 - 3 is probably neater to imoplement, but it seems less intuitive as a user (at least to me). 1 seems like a more natural extension of JUnit's [`@ValueSource`](https://junit.org/junit5/docs/5.3.1/api/org/junit/jupiter/params/provider/ValueSource.html) from a user perspective. It may require some ugly validation code to make sure your `from`s and `to`s are aligned, but IMHO it's a small price to pay.

Other thoughts/opinions are more than welcome, of course.

src/main/java/org/junitpioneer/jupiter/RangeSource.java
nicolaiparlog: I unresolved this, so we keep track of it.

nicolaiparlog:
> 1 seems like a more natural extension of JUnit's @ValueSource from a user perspective.

That's a good argument, let's go with that then.

mureinik:
I played around with this a bit, and unfortuntely it won't fly. Numerical annotation members are primitives, so you don't have a natural way to specify which arguments you do or don't want to use (unlike the `@ValueSource` usecade, where the members are arrays - so by default they are all `null` except the one you choose to assign values to).

Option 3 is actually quite doable, and I'll upload a patch series that attempts to do that.

nicolaiparlog:
Hey @mureinik, I'm alive! :blush: I just rebased your branch onto the recent master and made some formatting changes. I will now add a review. It would be awesome if, after all this time, you want to continue working on this. Either way, let me know.

src/main/java/org/junitpioneer/jupiter/RangeSourceProvider.java
```diff
@@ -0,0 +1,131 @@
...
+						"Expected exactly one annotaiton to provide an ArgumentSource, found %d.",
+						argumentsSources.size()));
+		// @formatter:on
+
+		Annotation argumentsSource = argumentsSources.get(0);
+		Class<? extends Annotation> argumentsSourceClass = argumentsSource.annotationType();
+
+		Method stepMethod = argumentsSourceClass.getMethod("step");
+		Number stepValue = (Number) stepMethod.invoke(argumentsSource);
+		double stepDouble = stepValue.doubleValue();
```
nicolaiparlog: This conversation is lossy. `long` contains numbers that `double` can't represent ([apparently](https://stackoverflow.com/a/3793950/2525313) the smallest one is 9,007,199,254,740,993 (2^53 + 1)). Also, `float from + float step` can lead to different results than `(float) ((double) float from + (double) float step)`.

One solution would be to have a `Range` interface with methods that return `Number` and concrete implementations for each primitive type,

src/main/java/org/junitpioneer/jupiter/RangeSourceProvider.java
```diff
@@ -0,0 +1,131 @@
...
+
+		Method closeMethod = argumentsSourceClass.getMethod("closed");
+		Boolean closedValue = (Boolean) closeMethod.invoke(argumentsSource);
+
+		// Once (if) Java 8 support is dropped, this boiler-plate can be cleaned up and we can just use
+		// IntStream.iterate(rangeSource.from(), i -> i < rangeSource.to(), i -> i += rangeSource.step());
+		double val = fromDouble;
+		while (val < toDouble && stepDouble > 0.0 || val > toDouble && stepDouble < 0.0
+				|| closedValue && val == toDouble) {
+			values.add(val);
```
nicolaiparlog: Instead of putting all (of possibly very many) elements into a list, I would create the stream directly. If the `Range` interface I mentioned above implements `Iterator<Number>` (that should be straight-forward) you can turn it into a stream as follows:

```java
StreamSupport.stream(
		Spliterators.spliterator(sourceIterator, size, Spliterator.ORDERED),
		false);
```

where `size` is roughly `(to - from) / step` (with some rounding).

src/test/java/org/junitpioneer/jupiter/RangeSourceProviderTests.java
```diff
@@ -0,0 +1,285 @@
...
+			};
+
+			ExtensionContext ec = spy(ExtensionContext.class);
+			when(ec.getElement()).thenReturn(Optional.of(dummy.getClass().getMethod("toString")));
+
+			RangeSourceProvider provider = new RangeSourceProvider();
+			PreconditionViolationException e = assertThrows(PreconditionViolationException.class,
+				() -> provider.provideArguments(ec));
+			assertEquals("Expected exactly one annotaiton to provide an ArgumentSource, found 0.", e.getMessage());
+		}
```
nicolaiparlog: I think there are better ways to test misconfigurations. Shouldn't it be possible to create a regular test method that is poorly annotated and use the common `executeTestsForClass` approach to verify errors?

src/main/java/org/junitpioneer/jupiter/ByteRangeSource.java
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * @since 4.0
+ * @see ArgumentsSource
+ * @see org.junit.jupiter.params.ParameterizedTest
+ */
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@ArgumentsSource(RangeSourceProvider.class)
+public @interface ByteRangeSource {
```
nicolaiparlog: Hm, if we want to use this source for #68, we need this source to be repeatable. No reason to implement that now, though (I guess?).

mureinik:
Hi @nicolaiparlog Sorry for the late replay - I'm alive too, but was traveling and didn't closely monitor this PR.
I'll try to take a look and implement your review over the weekend.

mureinik:
@nicolaiparlog Took a little longer to get to this than expected, but I pushed a new revision. Changes since the last revision:

- Rebased on top of the current master and made sure that each patch individually also passes spotless (i.e., essentially took you formatting improvements and applied them on the patches that broke formatting - thanks for the help BTW!)
- Refactored out the logic of creating the stream of numbers to a `Range` abstract class (in order to share functionality) and added a concrete implementation for each numeric type. I wonder if this didn't get to the point we have too many classes, and they would be better organized in sub-package, or even as inner classes. This should solve both the lossy conversion and the performance issue of filling a list with all the elements you commented on.

Open issues:
- Regarding documentation - I'm all for it, but some guidance would be appreciated. What exactly would you want to see there?
- Regarding having the annotation repeatable - I feel that #68 isn't quite fleshed out yet, so I'd rather not prematurely involve this PR in those considerations. If whoever works on it needs to make the range sources repeatable, they could as part of that PR (unless you particularly prefer it in this PR - just say no, In any case, it seems like a simple change)
- Regarding testing invalid annotations - I'll give it a look as per your suggestion. I preferred pushing a partial solution to get some feedback instead of waiting till I have something complete.

mureinik:
@nicolaiparlog Test refactoring done (really looks cleaner now, thanks for the pointer!), ready for your review.

nicolaiparlog:
Thanks for working on this, will take a look soon (ish). One random note:

> made sure that each patch individually also passes spotless

That's very diligent, but in the end, we will squash your entire contribution into a single commit, so no need to invest time into that.

mureinik:
Thanks @nicolaiparlog .

I noticed this PR now has merge conflicts due to some recent merge. I've pushed an updated patch series with the following changes:
- Rebased on top of the current master (the 0.4.2 release tag)
- Bumped the `junit-jupiter-params` requirement to 5.3.2 to align with the `junit-jupiter-api` requirement
- Updated the `@since` tags to 0.5, since 0.4 is clearly wrong at this point

src/main/java/org/junitpioneer/jupiter/RangeSourceProvider.java
```diff
@@ -0,0 +1,97 @@
...
+								.stream(annotations.annotationType().getAnnotationsByType(ArgumentsSource.class))
+								.anyMatch(annotation -> getClass().equals(annotation.value())))
+						.collect(Collectors.toList()))
+				.get();
+
+		Preconditions.condition(
+				argumentsSources.size() == 1,
+				() -> String.format(
+						"Expected exactly one annotation to provide an ArgumentSource, found %d.",
+						argumentsSources.size()));
```
sormuras: Couldn't you make use of [findAnnotation](https://junit.org/junit5/docs/current/api/org/junit/platform/commons/support/AnnotationSupport.html#findAnnotation(java.util.Optional,java.lang.Class))

src/main/java/org/junitpioneer/jupiter/ByteRangeSource.java
```diff
@@ -0,0 +1,55 @@
...
+
+	/**
+	 * The size of the step between the {@code from} and the {@code to}.
+	 */
+	byte step() default 1;
+
+	/**
+	 * Whether the range is closed (inclusive of the {@link #to()}) or not.
+	 */
+	boolean closed() default false;
```
nicolaiparlog: A totally non-representative poll of three guys in Twitch chat (plus me!) showed that we all prefer `true` as a default here. It's odd in Java terms, but if I write `@RangeSource(from = 1, to = 3)`, I expect `3` to be in there. Your take on this?

src/main/java/org/junitpioneer/jupiter/ByteRange.java
```diff
@@ -0,0 +1,27 @@
...
+package org.junitpioneer.jupiter;
+
+public class ByteRange extends Range<Byte> {
+	public ByteRange(byte from, byte to, byte step, boolean closed) {
+		super(from, to, step, closed);
+	}
+
+	@Override
+	public Byte nextValue() {
+		return (byte) (getCurrent() + getStep());
```
nicolaiparlog: When changing this to `getCurrent().byteValue() + getStep().byteValue()` (to get rid of the cast), the compiler helpfully pointed out that this may overflow. We need to discuss how to handle that. Wrap around? Or throw an exception?

At fist I thought, this can't happen because when `step`ping from `from` to `to` we'd stop early, but with `from=120`, `to=125`, and `step=10` it _could_ happen, right?

src/main/java/org/junitpioneer/jupiter/Range.java
```diff
@@ -0,0 +1,104 @@
...
+		if (!hasNext()) {
+			throw new NoSuchElementException();
+		}
+		current = getNextValue();
+		return current;
+	}
+
+	Stream<N> asStream() {
+		return StreamSupport.stream(Spliterators.spliteratorUnknownSize(this, Spliterator.ORDERED), false);
+	}
```
nicolaiparlog: I'd prefer it this way, smaller API surface:

```java
	Stream<N> asStream() {
		return StreamSupport.stream(Spliterators.spliteratorUnknownSize(new ValueIterator(), Spliterator.ORDERED), false);
	}

	private class ValueIterator implements Iterator<N> {

		@Override
		public boolean hasNext() {
			N nextValue = getNextValue();
			int cmp = nextValue.compareTo(to);
			return (cmp < 0 != sign < 0) || (closed && cmp == 0);
		}

		@Override
		public N next() {
			if (!hasNext()) {
				throw new NoSuchElementException();
			}
			current = getNextValue();
			return current;
		}

	}
```

nicolaiparlog:
First of all, great work Allon, thanks for doing this. :+1: About your last comments:

> I wonder if this didn't get to the point we have too many classes, and they would be better organized in sub-package, or even as inner classes.

Yes, I arrived at the same conclusion. Let's put this into a package. Is `range` too generic? Should it be `parameterrange` or `argumentsrange`? Opinions welcome!

> Regarding documentation - I'm all for it, but some guidance would be appreciated. What exactly would you want to see there?

Basically an Asciidoc file in `docs` folder. But let's do that once everything else is settled.

> Regarding having the annotation repeatable - I feel that #68 isn't quite fleshed out yet, so I'd rather not prematurely involve this PR in those considerations.

Good call.

src/main/java/org/junitpioneer/jupiter/RangeSourceProvider.java
mureinik: @sormuras `findAnnotation` returns the first annotation that matches the given the type. Here, we want to get all of them in order to check that conflicting range sources aren't used on the same method.

src/main/java/org/junitpioneer/jupiter/RangeSourceProvider.java
```diff
@@ -0,0 +1,97 @@
...
+
+		Method closeMethod = argumentsSourceClass.getMethod("closed");
+		Boolean closedValue = (Boolean) closeMethod.invoke(argumentsSource);
+
+		// @formatter:off
+		Preconditions.condition(
+				fromMethod.getReturnType() == toMethod.getReturnType(),
+				() -> String.format(
+						"Illegal range. Different classes of from (%s) and to (%s).",
+						fromMethod.getReturnType(), toMethod.getReturnType()));
```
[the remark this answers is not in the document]
mureinik: It's a remnant of one of the previous iterations. The idea wat to prevent any future contributor implementing a range with conflicting `from` and `to`, but that's probably a mute point now. I'll remove it.

src/main/java/org/junitpioneer/jupiter/RangeSourceProvider.java
```diff
@@ -0,0 +1,97 @@
...
+		// @formatter:on
+
+		String argumentSourceName = argumentsSourceClass.getName();
+		String rangeName = argumentSourceName.substring(0, argumentSourceName.length() - "Source".length());
+		Class rangeClass = Class.forName(rangeName);
+		Range<?> range = (Range) rangeClass.getConstructors()[0].newInstance(fromValue, toValue, stepValue,
+			closedValue);
```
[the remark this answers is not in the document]
mureinik: Interesting idea.
In an ideal world, I'd like to have a `getRangeClass()` method on the actual `RangeSource`, but I'm not aware of any way to make it unmodifiable in Java. Even with a default value, that'd leave the user with an odd API where they can't really supply anything other than that default.

In the current code I tried to work around the dependency between a `Range` and a `RangeSource` with naming conventions, but I think your suggestion to make it explicit is more idiomatic to Java. I'll cook up an implementation and update this PR.

src/main/java/org/junitpioneer/jupiter/ByteRangeSource.java
mureinik: Looking at other similar APIs, like `IntStream.range` or even `List.subList` or `String.substring`, the "to" argument is always exclusive. I think having "to" here inclusive by default would be confusing than having it exclusive by default.

src/main/java/org/junitpioneer/jupiter/ByteRange.java
mureinik: TBH, I'm not a big fan of either of these suggestions.
To an average user `ByteRangeSource(from = 120, to = 125, step = 10)` looks like a valid range source (albeit not a particularly elegant one...). We could recognize overflows by comparing the relation between the current and next value to the sign of the step (e.g., if we calculate the next value and overflow, it will be smaller than the current value, which shouldn't happen if the step is positive), and just terminate the stream. Thus, this range source will return a single value, `120`.

I'll add a patch to treat it, and then we can discuss against some code.

src/main/java/org/junitpioneer/jupiter/Range.java
mureinik: Just to be clear - if you're suggesting remove `asStream` from `Range`, sure, that's not a problem, will do.

If you're suggesting separating `Range` from a new `ValueIterator` class, I'm not sure I'm following.

mureinik:
@nicolaiparlog @sormuras pushed a new revision with the following updates:

1. Rebased on top of master
1. Updated the copyright notices to 2020, as per 7910b5d33b6f01453ab7839c60f55cf3a466598a
1. Removed a useless check that the `to` and `from` arguments have the same type
1. Added a `@RangeClass` annotation to allow the various `RangeSource`s to map to the appropriate `Range`s without relying on naming conventions
1. Added detection for overflows, and terminated the argument stream once its detected
1. Moved `asStream` out of `Range` and in to `RangeSourceProvider`
1. Moved all the classes into the `range` subpackage. The name is a bit short, but given the context of `org.junitpioneer.jupiter` I think it's explanatory enough. This could of course be renamed if we feel there's a better name for it.

I've left all the comments unresolved so you could review and decide for yourself if the changes in this lastest set of patches indeed resolve the issues or not.

TODO: I still haven't added a documentation file. As stated by @nicolaiparlog: "But let's do that once everything else is settled."

src/main/java/org/junitpioneer/jupiter/RangeSourceProvider.java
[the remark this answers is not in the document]
nicolaiparlog: Looks good.

src/main/java/org/junitpioneer/jupiter/ByteRangeSource.java
nicolaiparlog: Yeah, it's "Java conventions" vs "common sense". Let's leave it as it is and see whether users complain. We're pre-1.0, so we can break all the things! :wink:

src/main/java/org/junitpioneer/jupiter/ByteRange.java
nicolaiparlog: Just checked, the current implementation would result in a silent wrap, i.e. `120 + 10 ~> -126`. Indeed a corner case we need to discuss.

src/main/java/org/junitpioneer/jupiter/Range.java
nicolaiparlog: Yes, I was suggesting to make `ValueIterator` an inner class, so that `Range` does not implement `Iterator`. One week later, I don't really see a problem with that, though. An iterable range makes sense. Let's leave it as it is. :)

src/main/java/org/junitpioneer/jupiter/ByteRange.java
nicolaiparlog: Oh, only now saw the implementation in `hasNext` - good job!

src/main/java/org/junitpioneer/jupiter/range/Range.java
```diff
@@ -0,0 +1,100 @@
...
+
+	@Override
+	public boolean hasNext() {
+		if (current == null) {
+			return true;
+		}
+		N nextValue = getNextValue();
+		int cmp = nextValue.compareTo(to);
+		int overflowCheck = nextValue.compareTo(current);
+		return ((cmp < 0 != sign < 0) && (overflowCheck < 0 == sign < 0)) || (closed && cmp == 0);
```
Bukama: I suggest to extract these three comparisons to a boolean variable with a meaningfull each and put the variables in the return statement. At least for me it's very hard to read and @nicolaiparlog  was also struggeling on stream lately ;)

nicolaiparlog:
Great, I think we're getting ready to ship this. As to documentation, have a look into the top-level `docs` folder:

* we need a `*.adoc` file explaining the feature in a few paragraphs with some code snippets (we need to be crystal clear that this builds on parameterized tests)
* we need an entry in `docs-nav.yaml`

Good work on the Javadoc, by the way! :+1:

nicolaiparlog:
Oh, one more thing on the package: Let's call it `params` in accordance with Jupiter's package name for parameterized tests. That also clarifies for us which future extensions belong into that package and which don't. Will you do the honors? :blush:

mureinik:
> Oh, one more thing on the package: Let's call it `params` in accordance with Jupiter's package name for parameterized tests. That also clarifies for us which future extensions belong into that package and which don't. Will you do the honors?

Sure, will do.

src/main/java/org/junitpioneer/jupiter/range/Range.java
mureinik: I'm open to suggestions :-)

mureinik:
@nicolaiparlog ready for your review - renamed the `range` package to `params` as commented above, and added a first attempt to document this feature.

nicolaiparlog:
Thanks so much @mureinik for riding this out. Another PR that took from 2018 to 2020 to merge. :joy:

mureinik:
Hey, better late than never :-) Thanks for merging!

Issue #124, opened 2018-10-10
Site build fails


sormuras:
Triggering the site build fails with `code=403, message=Forbidden`.

https://travis-ci.org/junit-pioneer/junit-pioneer/jobs/439802792#L838

nicolaiparlog:
Looks like [the Pioneer CI setup](https://travis-ci.org/junit-pioneer/junit-pioneer/settings) simply didn't have an environment variable with the Travis API token. I/we must have accidentally deleted it at some point. 🤪

PR #132, opened 2018-11-01
Reactivate site build with more verbose error messages

nicolaiparlog:
The site build could not be triggered, apparently because [the CI setup](https://travis-ci.org/junit-pioneer/junit-pioneer/settings) didn't set the environment variable `TRAVIS_API_TOKEN`. We must have accidentally deelted it. I added it back and this commit reactivates the deployment step.

The class `TriggerTravisTask` now checks whether the token is empty and prints an informative message before exiting if it is. This is more informative than a 403 from the Travis endpoint.

Closes: #124

---

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,12 +48,12 @@ deploy:
     on:
       branch: master
       jdk: oraclejdk8
-# - provider: script
-#   skip_cleanup:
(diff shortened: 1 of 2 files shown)


nicolaiparlog:
Whoever reviews this and finds it acceptable can immediately merge it. The commit message is good as it is.

PR #133, opened 2018-11-01
Introduce system property extension

### LINKED ISSUE #129: System property extension
While looking for a way to handle system properties in JUnit 5 tests [on Stack Overflow](https://stackoverflow.com/q/46846503/3429133), I came up with the following extension (see [this gist](https://gist.github.com/beatngu13/724a9a78bba61da96566ac5478c2d37b)) based on the provided answer:

```java
class SystemPropertyExtensionTest {

	@BeforeAll
	static void setUpOnce() {
		System.setProperty( "some property", "old value" );
	}

	@SystemProperty( key = "some property" )
	@Test
	void extension_should_set_property_to_null() {
		Assertions.assertNull( System.getProperty( "some property" ) );
	}

	@SystemProperty( key = "some property", value = "new value" )
	@Test
	void extension_should_set_property_to_value() {
		Assertions.assertEquals( System.getProperty( "some property" ), "new value" );
	}

	@SystemProperty( key = "some property" )
	@SystemProperty( key = "another property", value = "new value" )
	@Test
	void extension_should_be_repeatable() {
		Assertions.assertNull( System.getProperty( "some property" ) );
		Assertions.assertEquals( System.getProperty( "another property" ), "new value" );
	}

	@AfterAll
	static void tearDownOnce() {
		Assertions.assertEquals( System.getProperty( "some property" ), "old value" );
	}

}
```

After posting it [on Twitter](https://twitter.com/beatngu1101/status/1054723804822429696), @sbrannen drew my attention to JUnit Pioneer. I would be happy to submit a PR, which is why I have opened this question to debate the solution.

IMHO there are currently two minor issues:

* ~~The annotation probably should be repeatable (as already pointed out [on Stack Overflow](https://stackoverflow.com/questions/46846503/properly-set-system-properties-in-junit-5#comment80723376_46851140)), in case somebody has to use more than one property.~~ (Adapted gist accordingly.)
* Since `null` is not allowed for default values, I have picked `""` to see if the property only needs to be cleared or set to a specific value. Are there situations where someone has to set a property to an empty string, where this wouldn't work anymore?

beatngu13:
Introduces a Jupiter extension to clear (`@ClearSystemProperty`) or set (`@SetSystemProperty`) system properties for tests. Both annotations work on the test method and class level, are repeatable as well as combinable. Resolves #129.

```java
@ClearSystemProperty(key = "some property")
@Test
void clearSystemProperty() {
	assertThat(System.getProperty("some property")).isNull();
}

@SetSystemProperty(key = "some property", value = "new value")
@Test
void setSystemProperty() {
	assertThat(System.getProperty("some property")).isEqualTo("new value");
}
```

Resolves #129

---

### DIFF
--- a/docs/system-properties.adoc
+++ b/docs/system-properties.adoc
@@ -2,7 +2,7 @@
 :page-description: JUnit Jupiter extensions to clear and set the values of system properties
 
 The `@ClearSystemProperty` and `@SetSystemProperty` annotations can be used to clear, respectively, set the values of system properties for a test execution.
-Both annotations work on the test method level and are repeatable as well as combinable.
+Both annotations work on the test method and class level, are repeatable as well as combinable.
 After the annotated method has been executed, the initial default value is restored.
 
 For example, clearing a system property for a test execution can be done as follows:
@@ -38,5 +38,18 @@ void test() {
     assertNull(System.getProperty("1st property"));
     assertNull(System.getProperty("2nd property"));
     assertEquals("new value", System.getProperty("3rd property"));
+}
+----
+
+Note that class level configurations are overwritten by method level configurations:
+
+----
+@ClearSystemProperty(key = "some property")
+class MySystemPropertyTest {
+	@Test
+	@SetSystemProperty(key = "some property", value = "new value")
+	void test() {
+		assertEquals("new value", System.getProperty("some property"));
+	}
 }
 ----
--- a/docs/system-properties.adoc
+++ b/docs/system-properties.adoc
@@ -3,7 +3,8 @@
 
 The `@ClearSystemProperty` and `@SetSystemProperty` annotations can be used to clear, respectively, set the values of system properties for a test execution.
 Both annotations work on the test method and class level, are repeatable as well as combinable.
-After the annotated method has been executed, the initial default value is restored.
+After the annotated method has been executed, the properties mentioned in the annotation will be restored to their original value or will be cleared if they didn't have one before.
+Other system properties that are changed during the test, are *not* restored.
 
 For example, clearing a system property for a test execution can be done as follows:
 
@@ -29,6 +30,7 @@ void test() {
 
 As mentioned before, both annotations are repeatable and they can also be combined:
 
+[source,java]
 ----
 @Test
 @ClearSystemProperty(key = "1st property")
@@ -43,6 +45,7 @@ void test() {
 
 Note that class level configurations are overwritten by method level configurations:
 
+[source,java]
 ----
 @ClearSystemProperty(key = "some property")
 class MySystemPropertyTest {
--- a/docs/system-properties.adoc
+++ b/docs/system-properties.adoc
@@ -49,10 +49,10 @@ Note that class level configurations are overwritten by method level configurati
 ----
 @ClearSystemProperty(key = "some property")
 class MySystemPropertyTest {
-	@Test
-	@SetSystemProperty(key = "some property", value = "new value")
-	void test() {
-		assertEquals("new value", System.getProperty("some property"));
-	}
+    @Test
+    @SetSystemProperty(key = "some property", value = "new value")
+    void test() {
+        assertEquals("new value", System.getProperty("some property"));
+    }
 }
 ----
--- a/docs/system-properties.adoc
+++ b/docs/system-properties.adoc
@@ -1,0 +1,42 @@
+:page-title: @ClearSystemProperty and @SetSystemProperty
+:page-description: JUnit Jupiter extensions to clear and set the values of system properties
+
+The `@ClearSystemProperty` and `@SetSystemProperty` annotations can be used to clear, respectively, set the values of system properties for a test execution.
+Both annotations work on the test method level and are repeatable as well as combinable.
+After the annotated method has been executed, the initial default value is restored.
+
+For example, clearing a system property for a test execution can be done as follows:
+
+[source,java]
+----
+@Test
+@ClearSystemProperty(key = "some property")
+void test() {
+    assertNull(System.getProperty("some property"));
+}
+----
+
+And setting a system property for a test execution:
+
+[source,java]
+----
+@Test
+@SetSystemProperty(key = "some property", value = "new value")
+void test() {
+    assertEquals("new value", System.getProperty("some property"));
+}
+----
+
+As mentioned before, both annotations are repeatable and they can also be combined:
+
+----
+@Test
+@ClearSystemProperty(key = "1st property")
+@ClearSystemProperty(key = "2nd property")
+@SetSystemProperty(key = "3rd property", value = "new value")
+void test() {
+    assertNull(System.getProperty("1st property"));
+    assertNull(System.getProperty("2nd property"));
+    assertEquals("new value", System.getProperty("3rd property"));
+}
+----
--- a/src/main/java/org/junitpioneer/jupiter/ClearSystemProperties.java
+++ b/src/main/java/org/junitpioneer/jupiter/ClearSystemProperties.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2018 the original author or authors.
+ * Copyright 2015-2019 the original author or authors.
  *
  * All rights reserved. This program and the accompanying materials are
  * made available under the terms of the Eclipse Public License v2.0 which
--- a/src/main/java/org/junitpioneer/jupiter/ClearSystemProperties.java
+++ b/src/main/java/org/junitpioneer/jupiter/ClearSystemProperties.java
@@ -20,7 +20,7 @@ import org.junit.jupiter.api.extension.ExtendWith;
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.METHOD)
 @ExtendWith(SystemPropertyExtension.class)
-public @interface ClearSystemProperties {
+@interface ClearSystemProperties {
 
 	ClearSystemProperty[] value();
 
--- a/src/main/java/org/junitpioneer/jupiter/ClearSystemProperties.java
+++ b/src/main/java/org/junitpioneer/jupiter/ClearSystemProperties.java
@@ -1,3 +1,12 @@
+/*
+ * Copyright 2015-2018 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v20.html
+ */
 
 package org.junitpioneer.jupiter;
 
--- a/src/main/java/org/junitpioneer/jupiter/ClearSystemProperties.java
+++ b/src/main/java/org/junitpioneer/jupiter/ClearSystemProperties.java
@@ -1,3 +1,4 @@
+
 package org.junitpioneer.jupiter;
 
 import java.lang.annotation.ElementType;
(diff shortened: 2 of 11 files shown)


src/main/java/org/junitpioneer/jupiter/DefaultTimeZone.java
```diff
@@ -22,7 +22,7 @@
  * returned by {@link java.util.TimeZone#getDefault()} for a test execution.
  *
  * <p>The {@link java.util.TimeZone} to set as the default {@code TimeZone} is
- * configured bei specifying the {@code TimeZone} ID as defined by
```
smoyer64: Is this change related to this PR? (don't get me wrong, I appreciate the effort to clean things up as their found).

smoyer64: If you wouldn't mind, and I promise I'll merge it immediately (are you signed up for HacktoberFest?).  As I noted above, I **do** appreciate people who help improve documentation!

beatngu13: There you go: https://github.com/junit-pioneer/junit-pioneer/pull/147 ✌️ Will shortly rebase and update this branch.

src/main/java/org/junitpioneer/jupiter/ClearSystemProperty.java
```diff
@@ -0,0 +1,42 @@
...
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.METHOD)
+@Repeatable(ClearSystemProperties.class)
+@ExtendWith(SystemPropertyExtension.class)
+public @interface ClearSystemProperty {
+
+	/**
+	 * The key of the system property to be cleared.
+	 */
+	String key();
```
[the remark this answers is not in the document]
beatngu13: I think I would go for the symmetric variant (i.e. with `key()`), but I have no strong opinion on that. If you prefer `value()`, I'm happy to change it.

beatngu13:
Updated the docs, please let me know if there is anything else missing or you would like to change something.

I've also changed the copyright note, which was outdated. Do you know why Spotless complains about this change?

https://travis-ci.org/junit-pioneer/junit-pioneer/jobs/606463317#L291

src/main/java/org/junitpioneer/jupiter/ClearSystemProperties.java
```diff
@@ -0,0 +1,27 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.junit.jupiter.api.extension.ExtendWith;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.METHOD)
```
nicolaiparlog: Does it make sense to allow class-level annotations as well?

src/main/java/org/junitpioneer/jupiter/ClearSystemProperty.java
```diff
@@ -0,0 +1,42 @@
...
+ * of a system property for a test execution.
+ *
+ * <p>The key of the system property to be cleared must be specified via
+ * {@link #key()}. After the annotated method has been executed, all system
+ * properties will be restored to their original value.</p>
+ *
+ * @since 0.4
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.METHOD)
```
nicolaiparlog: (See above)

src/main/java/org/junitpioneer/jupiter/ClearSystemProperty.java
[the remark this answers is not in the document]
nicolaiparlog: I agree, symmetry over brevity.

src/main/java/org/junitpioneer/jupiter/SystemPropertyExtension.java
```diff
@@ -0,0 +1,86 @@
...
+	@Override
+	public void afterEach(final ExtensionContext context) throws Exception {
+		if (annotationsPresentOnTestMethod(context)) {
+			resetOriginalSystemProperties(context);
+		}
+	}
+
+	private void resetOriginalSystemProperties(final ExtensionContext context) {
+		final Properties backup = context.getStore(NAMESPACE).get(KEY, Properties.class);
+		System.setProperties(backup);
```
nicolaiparlog: This restores all properties to the values they had before the test was run. If tests change system properties on purpose (sounds fishy, I know, but weirder things have happened), this extension overrides/negates that change. I think it would be better to just store/restore the properties used by the annotation.

To go further, maybe we should only restore system properties that have the value given to this extension? If a test sets any system property (even the one set for the test by this extension), we should not reset that. At least that's what my gut says atm. What do you think?

src/test/java/org/junitpioneer/jupiter/SystemPropertyExtensionTests.java
```diff
@@ -0,0 +1,95 @@
...
+
+	@Nested
+	@DisplayName("used with ClearSystemProperty")
+	class ClearSystemPropertyTests {
+
+		@ClearSystemProperty(key = "some property")
+		@Test
+		@DisplayName("should clear system property")
+		void shouldClearSystemProperty() {
+			assertThat(System.getProperty("some property")).isNull();
```
nicolaiparlog: A little over the top, maybe, but...

```java
assertThat(System.getProperty("another property")).isEqualTo("old value")
```

?

src/test/java/org/junitpioneer/jupiter/SystemPropertyExtensionTests.java
```diff
@@ -0,0 +1,95 @@
...
+	class MixedSystemPropertyTests {
+
+		@ClearSystemProperty(key = "some property")
+		@SetSystemProperty(key = "another property", value = "new value")
+		@Test
+		@DisplayName("should be mixable")
+		void clearAndSetSystemPropertyShouldBeMixable() {
+			assertThat(System.getProperty("some property")).isNull();
+			assertThat(System.getProperty("another property")).isEqualTo("new value");
+		}
```
nicolaiparlog: Good thing, this test is here. Gave the idea to check what happens in this case:

```java
@ClearSystemProperty(key = "some property")
@SetSystemProperty(key = "some property", value = "new value")
```

I'd consider this an implementation error and think the extension should throw an exception to make the developer aware of it by throwing an exception.

docs/system-properties.adoc
```diff
@@ -0,0 +1,42 @@
+:page-title: @ClearSystemProperty and @SetSystemProperty
+:page-description: JUnit Jupiter extensions to clear and set the values of system properties
+
+The `@ClearSystemProperty` and `@SetSystemProperty` annotations can be used to clear, respectively, set the values of system properties for a test execution.
+Both annotations work on the test method level and are repeatable as well as combinable.
+After the annotated method has been executed, the initial default value is restored.
```
nicolaiparlog: See comments below - we may have to update these lines.

beatngu13:
Thanks for your review, very helpful comments! Resolved all but:

* https://github.com/junit-pioneer/junit-pioneer/pull/133#discussion_r356209953
* https://github.com/junit-pioneer/junit-pioneer/pull/133#discussion_r356213963

Have to think about them … 🤔

src/main/java/org/junitpioneer/jupiter/SystemPropertyExtension.java
beatngu13: I now only store and restore those system properties that have been specified via the extension, not all per se. But I think it is fine to restore a system property in this situation:

>If a test sets any system property (even the one set for the test by this extension), we should not reset that.

I would argue that this is defined behavior as specified in the Javadoc. I also believe other extensions such as `DefaultLocaleExtension` and `DefaultTimeZoneExtension` work the same.

What do you say?

nicolaiparlog: Makes sense, let's go with that. For now - maybe I'll come back to it in a separate issue.

nicolaiparlog:
Hey @beatngu13, looking very good. I got sidetracked into refactoring storing/restoring the backup properties ~> 45b54ad. I wasn't quite happy how the code (almost) repeated itself when storing and when restoring, but I'm not sure, whether my refactoring improved matters much (or at all).

Have a look and let me know what you think - we can revert it if you want. Otherwise we're done here. :raised_hands: Time to come up with [a commit message](https://github.com/junit-pioneer/junit-pioneer/blob/master/CONTRIBUTING.md#merging).

Bukama:
> Hey @beatngu13, looking very good. I got sidetracked into refactoring storing/restoring the backup properties ~> [45b54ad](https://github.com/junit-pioneer/junit-pioneer/commit/45b54adcddf75519813a1c4ef2801d881d0b6224). I wasn't quite happy how the code (almost) repeated itself when storing and when restoring, but I'm not sure, whether my refactoring improved matters much (or at all).
> 
> Have a look and let me know what you think - we can revert it if you want. Otherwise we're done here. 🙌 Time to come up with [a commit message](https://github.com/junit-pioneer/junit-pioneer/blob/master/CONTRIBUTING.md#merging).

As written in chat yesterday I like the refactoring as it's reduce duplicate code and improves readabilty of some things (like the check for properties that were `@Set` and `@Clear`ed). And, you introduced a check for duplicate annotations - even if we have to wait what will come out of [Junit5 Issue 2131](https://github.com/junit-team/junit5/issues/2131).

beatngu13:
Hey @nicolaiparlog @Bukama,

I wanted to join the Twitch stream, but our daughter was born earlier than we expected … 😉

However, I had a quick look at the changes and they are looking really good! I also didn't like the redundancy I had in my code, so thanks for taking the time and fixing it. Also the fact that we now check for duplicate annotations is really helpful.

Will go over the changes in detail and create a proper commit message ASAP.

src/main/java/org/junitpioneer/jupiter/SystemPropertyExtension.java
```diff
@@ -0,0 +1,167 @@
...
+				.orElse(false);
+		//@formatter:on
+	}
+
+	private void handleSystemProperties(ExtensionContext context) {
+		Set<String> propertiesToClear;
+		Map<String, String> propertiesToSet;
+		try {
+			propertiesToClear = findRepeatableAnnotations(context, ClearSystemProperty.class).stream()
+					// collect to map because the collector throws an IllegalStateException on
```
verhas: Generally: if there is a need for an inline comment then it is a code smell. In this case, the code uses a `Map` instead of a `Set` because the collector to the `Map` throws an exception, while the collector to the `Set` does not (which is btw. absolutely logical).

The nice solution is to use a special set that throws an exception when an element already in the set is added:

```java
private static class UniqueThrowingSet extends HashSet {
        @Override
        public boolean add(Object o) {
            if (contains(o)) {
                throw new IllegalStateException("Duplicate element cannot be added");
            }
            return super.add(o);
        }
    }
```
I am not advocating this solution against the current solution, because this seems to be a bit overkill for the actual problem.

src/main/java/org/junitpioneer/jupiter/SystemPropertyExtension.java
```diff
@@ -73,68 +97,71 @@ private void handleSystemProperties(ExtensionContext context) {
...
+		context.getStore(NAMESPACE).get(BACKUP, SystemPropertyBackup.class).restoreProperties();
+	}
+
+	/**
+	 * Stores which system properties need to be cleared or set to their old values after the test.
+	 */
+	private static class SystemPropertyBackup {
+
+		private final Map<String, String> propertiesToSet;
+		private final Collection<String> propertiesToUnset;
```
beatngu13: Wouldn't `Set` be more appropriate?

beatngu13:
Proposed commit message:

```
Implement @ClearSystemProperty and @SetSystemProperty

Sometimes it is helpful to clear and/or set system properties for a
test execution. This PR introduces two annotations,
`@ClearSystemProperty` and `@SetSystemProperty`, that allow to do
this in a save manner with a proper restore mechanism. Both
annotations work on the test method and class level, are repeatable
as well as combinable.

Fixes: #129

PR: #133
```

nicolaiparlog:
Thanks a lot, @beatngu13 for the work and perseverance. :+1:

nicolaiparlog:
Wow, PR went from 2018 to 2020.

PR #151, opened 2019-10-22
Add GitHub release provider

### LINKED ISSUE #148: (Automatically) update GitHub releases
Latest release is 0.3.2 (although yet not available on [Maven Central](https://repo1.maven.org/maven2/org/junit-pioneer/junit-pioneer/)), but according to the [releases page](https://github.com/junit-pioneer/junit-pioneer/releases), it is 0.1.2.

I assume this is because nobody published the existing tags as actual releases on GitHub—as this gets easily forgotten.

However, one can automate this with [Travis' release provider for GitHub](https://docs.travis-ci.com/user/deployment/releases/):

```yaml
- provider: releases
  api_key: "${GH_TOKEN}"
  on:
    branch: master
    jdk: openjdk8
```

If you want to, I can submit a PR for this, but one of you guys has to provide a Github OAuth token as an environment variable (e.g. `GH_TOKEN` as above) in Travis.

beatngu13:
To automatically publish release on GitHub too. Resolves #148.

---

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,12 +34,6 @@ deploy:
     on:
       branch: master
       jdk: openjdk8
-  - provider: script
-    skip_cleanup:
-    script: ./gradlew triggerSiteBuild -P travisApiToken=$TRAVIS_API_TOKEN --stacktrace
-    on:
-      branch: master
-      jdk: openjdk8
   - provider: releases
     api_key: "${GITHUB_WRITE_TOKEN}"
     skip_cleanup: true
@@ -47,6 +41,12 @@ deploy:
     file: "./build/libs/junit-pioneer-*.jar"
     on:
       tags: true
+  - provider: script
+    skip_cleanup:
+    script: ./gradlew triggerSiteBuild -P travisApiToken=$TRAVIS_API_TOKEN --stacktrace
+    on:
+      branch: master
+      jdk: openjdk8
 
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,8 +43,7 @@ deploy:
   - provider: releases
     api_key: "${GITHUB_WRITE_TOKEN}"
     on:
-      branch: master
-      jdk: openjdk8
+      tags: true
 
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,6 +40,11 @@ deploy:
     on:
       branch: master
       jdk: openjdk8
+  - provider: releases
+    api_key: "${GITHUB_WRITE_TOKEN}"
+    on:
+      branch: master
+      jdk: openjdk8
 
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,6 +42,9 @@ deploy:
       jdk: openjdk8
   - provider: releases
     api_key: "${GITHUB_WRITE_TOKEN}"
+    skip_cleanup: true
+    file_glob: true
+    file: "./build/libs/junit-pioneer-*.jar"
     on:
       tags: true
 


.travis.yml
```diff
@@ -40,6 +40,11 @@ deploy:
     on:
       branch: master
       jdk: openjdk8
+  - provider: releases
+    api_key: "${GITHUB_WRITE_TOKEN}"
+    on:
+      branch: master
+      jdk: openjdk8
```
nicolaiparlog: Don't we also need `on.tags: true`?

.travis.yml
```diff
@@ -40,6 +40,11 @@ deploy:
     on:
       branch: master
       jdk: openjdk8
+  - provider: releases
+    api_key: "${GITHUB_WRITE_TOKEN}"
```
nicolaiparlog: Don't we also need `skip_cleanup: true`?

.travis.yml
beatngu13: I've had a look at the other release providers, and they only use these two preconditions. If I understood correctly, Pioneer is being pushed to Maven Central whenever something gets merged into `master`. Is this correct? The GitHub release provider here would follow this pattern, i.e., publish on GitHub when merged into `master`.

.travis.yml
beatngu13: Only if you want to publish additional artifacts, but then this would also need `file: ...`. The source code archives (see e.g. [release of v0.1.2.](https://github.com/junit-pioneer/junit-pioneer/releases/tag/v0.1.2)) are added per default.

beatngu13: Here are two examples from our repos:

1. Publish drafts (i.e. someone still has to manually confirm the release) without additional artifacts:
    * [`.travis.yml`](https://github.com/retest/recheck/blob/master/.travis.yml#L34)
    * [resulting releases](https://github.com/retest/recheck/releases)
2. Publish drafts with an additional fat JAR (needs `skip_cleanup` and `file`):
    * [`.travis.yml`](https://github.com/retest/recheck-web/blob/master/.travis.yml#L41)
    * [resulting releases](https://github.com/retest/recheck-web/releases)

.travis.yml
nicolaiparlog: Shipkit only releases commits to master under certain conditions:

- the commit message doesn't contain `[ci skip-release]`
- all checks (e.g. tests) are successful
- at least one main artifact (that includes `...-source.jar` and `...-javadoc.jar`) has changed

It also tags exactly the commits that fulfill these conditions, so `on.tags: true` should do the trick for us.

.travis.yml
nicolaiparlog: So you're saying the release plugin knows what files to add to the release? That's mildly surprising to me and [the documentation](https://docs.travis-ci.com/user/deployment/releases/#uploading-multiple-files) doesn't seem to mention it, but since it seems to work for Recheck, we can give it a try. :grin:

.travis.yml
beatngu13: Allright, I see. Good you double-checked this!

AFAIK one cannot combine `on.branch` and `on.tags` (see https://github.com/travis-ci/travis-ci/issues/7488). Are tags only created for releases? Because then it would be sufficient to check for tags only.

nicolaiparlog: Yes, at the moment tags are only created for releases, but relying on that seems a little fragile. I don't get why Travis can't check both and asked over at the issue you linked. Let's see what comes of it.

nicolaiparlog: No, wait, let's not see. Instead, let's edit and merge this and maybe come back to it in the future if need be.

.travis.yml
beatngu13: Magic! 🧙‍♂️

But no, it doesn't know what files to upload. Travis/GitHub only automatically archives (as zip and tar.gz) the source code of the release commit and adds this to the assets.

If there is nothing additionally to upload, there is no need to use `file` and `skip_cleanup`. Or is there anything you would like to add, in contrast to e.g. release v0.1.2?

nicolaiparlog: Oh wait, so this is just a ZIP/TAR of the entire repo at that point in time? Well, that's useless isn't it? You can get the same by navigating to [a tag](https://github.com/junit-pioneer/junit-pioneer/tree/v0.3.2) and hit download there. No, if we do this, we should do it right:

* include library JAR
* include sources JAR
* include Javadoc JAR

We could even go further and try to [pass a name and a body](https://docs.travis-ci.com/user/deployment/releases/#advanced-options), but that will be trickier.

I understand that this is more work than you were planning for. If you don't want to do this, that's perfectly ok, just let me know.

beatngu13: No problem, I can take care of it. I just wonder: Would anyone actually use Pioneer _without_ a build tool / something that takes care of dependency management, i.e., download the JARs from here, add the missing dependencies manually and then use it?

nicolaiparlog: Idk, I wouldn't. :grin: But if not, then why go through the trouble of dealing with GitHub's releases at all?

beatngu13: That's a valid question and I've asked this myself too some time ago. Let me show you this piece of art:

grafik

😉

I'd say it is a bit more user-friendly to present releases like this, but overall just a matter of taste.

However, as you already pointed out, we can also leave it as it is?

nicolaiparlog: > However, as you already pointed out, we can also leave it as it is?

Nah, no I want to do it. But right, with bells and whistles! :blush:

beatngu13: I'm a Gradle noob, but I assume `"./build/libs/junit-pioneer-*.jar"` would include the library, sources and Javadoc JARs? At least that's my impression after doing `gradle assemble` locally.

Maybe we have to trigger assembly in `before_deploy` to make sure the JARs are actually there. Could you please verify this? For example, create a tag with this release provider being active, then we should see what gets uploaded to GitHub. (You can also add `draft: true` for testing purpose, then the release isn't automatically published.)

nicolaiparlog: Ack on first paragraph.

Re second: If I get the docs correctly, I'd add `skip_cleanup: true` and hope that it means the previous build step doesn't get cleaned up, so we can reuse its artifacts.

nicolaiparlog: Oh, it's already there, nice. To be safe, lets move the `release` blog to line 37 (i.e. before triggering the website build), so it can reuse the same artifacts.

nicolaiparlog:
I was just writing the commit message when checking [the releases](https://github.com/junit-pioneer/junit-pioneer/releases) - wtf is going on there? Where does the release come from? :open_mouth:

beatngu13:
Looks like you have merged #152 yesterday? The script-based release providers are triggered on:

```yaml
on:
  branch: master
  jdk: openjdk8
```

Since the [conditions you have mentioned](https://github.com/junit-pioneer/junit-pioneer/pull/151#discussion_r337377066) also seem to be fulfilled, Shipkit created a release (935f0c3a532e62a0e404f5cb8121874d46501682, which is the parent of d98a3e439b5d1646c3c269bd4d72501b4ba0b9bb from the aforementioned PR). See also the build log:

https://travis-ci.org/junit-pioneer/junit-pioneer/jobs/601464876#L334

I just don't know why it is marked as a "proper" release here on GitHub, since the GitHub release provider hasn't been merged and the previous release don't follow this style too?

nicolaiparlog:
Yeah, I know why there was a release, but...

> I just don't know why it is marked as a "proper" release here on GitHub, since the GitHub release provider hasn't been merged and the previous release don't follow this style too?

Exactly! How weird! Anyway... as I see it, this PR is ready to be merged. And even if we, for some reason, now get proper GitHib-releases, this change adds helpful JARs to that, so let's merge it.

Issue #135, opened 2018-11-27
Add declarative annotations for "issue" type support and infrastructure


mkobit:
For example, Spock has [`@Issue`](http://spockframework.org/spock/javadoc/1.0/spock/lang/Issue.html) which is used to

> Indicates that a feature method or specification relates to one or more issues in an external issue tracking system.

Which might look like

```groovy
@Issue("https://github.com/junit-pioneer/junit-pioneer/issues/131")
```

It can often be useful to annotate test with some external issue for reporting, conditionally enabling/disabling, or to build additional tooling on.

Shorthand may also be desired if the project itself has some known destination:

```groovy
@Issue("junit-pioneer/junit-pioneer/issues/131")
```

It would be nice to have some kind of support for annotating code with issue references such that they can used for some of the previous use cases.

A few cases to consider for the API definition and consideration:

* Arbitrary issue URL - `@Issue("https://github.com/junit-pioneer/junit-pioneer/issues/131")`
* Arbitrary issue URL with prefix coming from some other configuration -  `@Issue("junit-pioneer/junit-pioneer/issues/131")`
* Public GitHub shorthand - `@GitHub("junit-pioneer/junit-pioneer", 131)`
* Private GitHub enterprise, JIRA, GitLab, etc.
* Meta-annotated issue - `@JUnitPioneerIssue(131)`

nicolaiparlog:
If this is just about showing the URL for an issue, a comment would do just fine. I would only use an annotation if the issue could somehow be relevant for test execution, but I struggle to see a use case for that. Do you have any idea, @mkobit?

marcphilipp:
Related issue: https://github.com/junit-team/junit5/issues/1900

nicolaiparlog:
Thanks for pointing that out. So from what I gathered, there are two possible features:

* add given issue to report
* gather all test results for an issue and report them

The former is trivial and we could do that. The latter sounds much more interesting! I'm imagining a dynamically generated test tree where each issue is a node and each test belonging to that issue is a subnode. By storing the results of the original tests when they were run, the generated tests could trivially pass or throw the same exception, effectively giving a second view on the same test results. Thing is, we'd need a way to guarantee that this test tree is generated at the end of a run when all results are in - is there a way to do that?

Feature set aside, I see three different ways going forward:

1. do whatever we can now, pioneering (get it?) a feature that may become part of JUnit 5 at some point
2. wait for JUnit 5 to decide on a report format and then do whatever we can
[...]

PR #153, opened 2019-11-26
[WIP-Need suggestion]Pioneer135/issue annotation

Bukama:
I started to create the `@Issue` annotation but I'm stuck. I tried to write an extension like described in [Nicolais Blog](https://blog.codefx.org/design/architecture/junit-5-extension-model/) but failing to write tests for it.

I want to write tests to check if a testmethod is annotated with the annotation and if yes check the value is correct.  I tried several things the last two hours (like implementing the `ParameterResolver` but still don't get access to the `ExtensionContext` to check the value of the `Issue` key.

When I was trying to implement the `@Requirement`-annotation I used an attribute in the `Testdescriptor`class (see [the commit here](https://github.com/junit-team/junit5/pull/1940/commits/ea51b0556d56eec13fbe63da21795373efe9ee01) but this was not the way the Junit-Team wants such extensions.

Would appreciate some suggestions how to move on to check the annotations and then the generated test result xml :/ Take your time with answering (if you want), as I'm in hospital from tomorrow for the next about ten days and won't have a notebook there.

---


nicolaiparlog:
This PR relates to #135. It's ok not to test what you have written so far. Just continue implementing your desired feature and we'll figure out how to write tests when we know what exactly that looks like. (I call it DDT - Development Driven Tests.)

Bukama:
I'll close this PR as I'm not able to implement it, I can't even write any tests to ensure if the Extension is working. @sormuras said "Believe, it will work", but I don't believe this. I don't want to make my first open source "contribution" is one which only consists of a "hey some magic happens here - believe me"-thing. And that he can't tell a way to tests extensions doesn't convinces me either - sorry Chris ;)

But to be honest just only one point, why I close it. The second one is that I don't see any way to get the annotated value into the result.

When I look at the [changelist ](https://github.com/junit-team/junit5/compare/master...Bukama:issues/1900-new-annotation-requirement) of my rejected PR in the Jupiter repository there are changes that are not possible (or I don't see in pioneer, like:

- extending and using the ReportEntry as it's final
- extending and using the TestDescriptor
- changing and using the report XSD
- changing and using the ReportXMLWriter

Maybe one other finds some magic to do this, but I'm sucking to much to do it.

Issue #137, opened 2018-12-04
Drop support for confusing Test annotation


shollander:
I believe that `@org.junitpioneer.vintage.Test` should be removed. It only introduces classpath pollution and confusion, and does not provide any real benefit. JUnit Jupiter provides much better methods for dealing with expected exceptions and timeout of tests. For those in migration, JUnit Vintage can be used. We should be encouraging users to use JUnit Jupiter improvements instead of helping them continue on with legacy practives.

At the very least it should be renamed so that it cannot be confused with `org.junit.jupiter.api.Test`

nicolaiparlog:
I understand your argument regarding import pollution and the annotation's limited use case. To clarify on the latter: My idea was that with this annotation, one can move from 4 to Jupiter with a full-text search/replace on `import org.junit.Test` before gradually updating those methods to Jupiter's superior facilities. I wish we could poll users of Pioneer and developers interested in a migration from 4 to 5/Jupiter to see whether they find this useful.

Here are the options as I see them:

* keep everything as it is ~> has the downsides you mention
* rename, e.g. to `VintageTest` ~> negates the benefits by making search/replace considerably more complex
* deprecate ~> may be a good middle ground as it (a) helps choosing the right import and (b) marks tests as deprecated, which spurs proper migration
* remove ~> then it's gone :confused: 

@junit-pioneer/junit-pioneer-team What do you think?

marcphilipp:
I'd be in favor of deprecating it.

PR #154, opened 2019-11-30
Deprecate vintage @Test

nicolaiparlog:
See #137 

---


nicolaiparlog:
Proposed commit message:

```
Deprecate vintage @Test

With Pioneer as a dependency, IDEs offer to import its vintage
`@Test` annotation, which is rarely what the user wants. To
reduce this risk of accidental use (and to emphasize that it's
intended as an intermediate step in a full migration) it's now
marked as deprecated (it is not intended to be removed, though).

Closes: #137
PR: #154
```

PR #156, opened 2019-12-11
Simplify and align search for annotations

nicolaiparlog:
All done. Let's bikeshed the name `Util`.

---

### DIFF
--- a/src/main/java/org/junitpioneer/jupiter/Util.java
+++ b/src/main/java/org/junitpioneer/jupiter/Util.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2015-2020 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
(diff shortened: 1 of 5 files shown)


src/main/java/org/junitpioneer/jupiter/Util.java
```diff
@@ -0,0 +1,41 @@
...
+	 * <em>present</em> or <em>meta-present</em> on the test method belonging
+	 * to the specified {@code context}.
+	 */
+	public static boolean annotationPresentOnTestMethod(ExtensionContext context,
+			Class<? extends Annotation> annotationType) {
+		//@formatter:off
+        return context.getTestMethod()
+                .map(testMethod -> AnnotationSupport.isAnnotated(testMethod, annotationType))
+                .orElse(false);
+        //@formatter:on
```
jbduncan: Oops, looks like the whitespace here is misaligned. ;)

nicolaiparlog: Good catch! Some spaces snuck in. :scream:

nicolaiparlog:
I think we'll stick with `Util` for now. We can change the name when we know more about what kind of stuff accrues there.

PR #162, opened 2020-01-19
Fix distinct set collector

beatngu13:
A typo was introduced during refactoring (73d1d26) that caused the combiner to not work properly.

This PR adds a bunch of tests (ff16e46), fixes the aforementioned issue (6a90863) and does some refactoring (c30ae03 and 400983e).

Proposed commit message:

```
Fix distinct set collector

[As above, but without commit references.]

PR: 162
```

---

### DIFF
--- a/src/test/java/org/junitpioneer/jupiter/UtilsTest.java
+++ b/src/test/java/org/junitpioneer/jupiter/UtilsTest.java
@@ -85,7 +85,7 @@ class UtilsTest {
 
 			BinaryOperator<Set<Object>> combiner = collector.combiner();
 
-			assertThat(combiner.apply(left, right)).containsExactly(leftElement, rightElement);
+			assertThat(combiner.apply(left, right)).containsExactlyInAnyOrder(leftElement, rightElement);
 		}
 
 		@Test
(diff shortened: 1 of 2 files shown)


src/test/java/org/junitpioneer/jupiter/UtilsTest.java
```diff
@@ -0,0 +1,109 @@
...
+		void setUp() throws Exception {
+			collector = Utils.distinctToSet();
+		}
+
+		@Test
+		@DisplayName("should use hash set")
+		void supplierShouldUseHashSet() throws Exception {
+			Supplier<Set<Object>> supplier = collector.supplier();
+
+			assertThat(supplier.get()).isInstanceOf(HashSet.class);
```
nicolaiparlog: Do you think we should really verify that it's a `HashSet` - any other implementation should be fine as well, right?

nicolaiparlog:
Also: I'm very embarrassed. #shameonme

src/test/java/org/junitpioneer/jupiter/UtilsTest.java
beatngu13: Yes, another implementation would do the job too. I just thought it would make sense to specify the current implementation as part of the tests. Feel free to delete the test case.

nicolaiparlog: I'll do that then. :)

nicolaiparlog:
Good work, thanks! :+1:

PR #170, opened 2020-02-12
Disable parameterised tests selectively

### LINKED ISSUE #163: Disable Parameterised tests selectively
### **Current situation**
Currently any @Disabled extensions that are there in the junit5 framework disable any test at the class level or at the invocation level before any of the parameters are resolved and the tests are launched

This leads to disable all the tests that were potentially going to run from any of the Parameterized test sources and it reports the test as only one skipped test.

### **Suggestion**
To possible have an extension similar to @Disabled (and it's counterparts) which is able to disable a single instance of the Parameterized test rather than all of them and only marks that individual test as skipped while others can run

Currently because of the following it is not straight forward to build something like this 
https://github.com/junit-team/junit5/issues/1139
https://github.com/junit-team/junit5/issues/1668
https://github.com/junit-team/junit5/issues/1884

And the only way is to use reflection to get access to the arguments for a Parameterized test's resolved argument

It would be nice to have native support for this as well, however as I understand, for programatic argument sources, specially ArgumentProvider and MethodSource, it isn't always sensible to let the arguments get initialised if we're skipping the tests

Currently I've built the following solution for this (it needs to be "productionalized" the answer is just a hint to actual impl) 
https://stackoverflow.com/questions/59716009/is-it-possible-to-disable-a-parameterized-test-basis-arguments-in-junit5

However, it's only valid till Junit5 version 5.3.x, after 5.5.x the Extension classes are package-private and the same solution needs a little bit of tweaking.

Michael1993:
Mostly copy-paste work. Hope it's OK. Closes #163

---

### DIFF
--- a/src/main/java/org/junitpioneer/jupiter/Disable.java
+++ b/src/main/java/org/junitpioneer/jupiter/Disable.java
@@ -1,3 +1,13 @@
+/*
+ * Copyright 2015-2020 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v20.html
+ */
+
 package org.junitpioneer.jupiter;
 
 import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


nicolaiparlog:
The initial code was really just to quickly show a non-declarative solutions. I pondered it for a while - here are my thoughts:

* we surely need ways to express combinations beyond concrete tuples (e.g. "disable if `number` is in [1, 4] and `text` is `"b"`)
* I'm not sure whether we can do everything in just one class (I doubt it, actually)
* maybe `Disable` should be generic, but then we need one type per number of test parameters (1-5 should be enough "for everybody")
* are we rebuilding an assertion library? It feels like assertions are mostly the same thing, but I think they're not good at asserting combinations, which is essential to what we want to do

What about this?

```java
@ParameterizedTest
@CsvSource({ "1, a", "1, b", "1, c", "2, a", "2, b", "2, c", "3, a", "3, b", "3, c" })
public void test(int number, String text) {
	Disable.whenPasses(() -> assertThat(number).is(5));

	Disable.whenAllPass(
		() -> assertThat(number).is(5),
		() -> assertThat(text).is("c"));

	Disable.whenAnyPasses(
		() -> assertThat(number).is(5),
		() -> assertThat(text).is("c"));
}
```

Compared to `Disable.when(number, text).is(1, "a")`:

* more code for the user to write
* less code for us to write
* more powerful ~> fewer feature requests
* presumably users already know their assertion library

Maybe I'm overlooking some obvious failure, but I really like it. I'm really curious to learn what you think!

Michael1993:
I mean I still don't fully understand the use of this feature. You could just write:

```java

@ParameterizedTest
@CsvSource({"1, a", "2, b", "3, c", "4, d", "5, e", "6, f"}
public void test(int number, String text) {
  if (!(number > 1 && number <= 4 && text.equals("a"))) {
    //the actual test
  }
}
```

Isn't that effectively the same thing?

nicolaiparlog:
No, it's not the same thing. :) With an ordinary `if` the test is considered to have passed (i.e. "green"), whereas with `Disabled`, it gets marked as, well, disabled, which is closer to the truth.

Michael1993:
This is not only outdated but the approach is outright wrong, so I'm closing this.

PR #174, opened 2020-02-18
Add Annotation for System Environment Variables

### LINKED ISSUE #167: Add Annotation for System Environment Variables
Similiar to @SetSystemProperty and @ClearSystemProperty it would be useful if there were a option to set and clear System Enviroment Variables (java.lang.System.getenv). 

"We can update [System] Properties at runtime while Environment Variables are an immutable copy of the Operating System's variables." [1].

I found a JUnit 4 Rule that does it [2] and a not released Lambda solution [3] from the same author. Both solutions have this warning "This rule uses reflection for modifying internals of the environment variables map. It fails if your SecurityManager forbids such modifications." 

I am not sure about the license stuff for reusing parts of [2] or [3] in this project  (see also [4]). Would that be a problem?

[1] https://www.baeldung.com/java-system-get-property-vs-system-getenv
[2] https://github.com/stefanbirkner/system-rules/blob/master/src/main/java/org/junit/contrib/java/lang/system/EnvironmentVariables.java
[3] https://github.com/stefanbirkner/system-lambda/blob/master/src/main/java/com/github/stefanbirkner/systemlambda/SystemLambda.java (WithEnvironmentVariables)
[4] https://github.com/stefanbirkner/system-rules/issues/55#issuecomment-473049418

Hancho2009:
resolves #167 Add Annotation for System Environment Variables

---

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,7 @@
 language: java
 
-os:
-  - linux
-  - osx
-
 jdk:
   - openjdk8
-
-matrix:
-  exclude:
-    - os: osx
-      jdk: openjdk8
 
 before_cache:
   - rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ os:
   - osx
 
 jdk:
+  - oraclejdk8
   - openjdk8
 
 before_cache:
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,11 +8,6 @@ jdk:
   - oraclejdk11
   - openjdk8
 
-jobs:
-  exclude:
-    - os: osx
-      jdk: openjdk8
-
 before_cache:
   - rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
   - rm -fr $HOME/.gradle/caches/*/plugin-resolution/
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ os:
   - osx
 
 jdk:
-  - oraclejdk11
   - openjdk8
 
 matrix:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,17 @@
 language: java
 
 os:
-  - windows
   - linux
   - osx
 
 jdk:
-  - oraclejdk8
+  - oraclejdk11
   - openjdk8
+
+jobs:
+  exclude:
+    - os: osx
+      jdk: openjdk8
 
 before_cache:
   - rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,11 @@ jdk:
   - oraclejdk11
   - openjdk8
 
+matrix:
+  exclude:
+    - os: osx
+      jdk: openjdk8
+
 before_cache:
   - rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
   - rm -fr $HOME/.gradle/caches/*/plugin-resolution/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,10 @@
 language: java
 
+os:
+  - windows
+  - linux
+  - osx
+
 jdk:
   - openjdk8
 
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -203,3 +203,11 @@ tasks.register<org.junitpioneer.gradle.TriggerTravisTask>("triggerSiteBuild") {
     apiToken = travisApiToken
     message = "Triggered by successful JUnit Pioneer build for %COMMIT"
 }
+
+test {
+    testLogging {
+        // set options for log level LIFECYCLE
+        events "failed"
+        exceptionFormat "full"
+    }
+}
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -104,6 +104,12 @@ tasks {
             includeTestsMatching("*Tests")
         }
         systemProperty("java.util.logging.manager", "org.apache.logging.log4j.jul.LogManager")
+        // `SystemPropertyExtension` uses reflection to change environment variables;
+        // this prevents the corresponding warning (and keeps working on Java 8)
+        // IF YOU ADD MORE OPTIONS; CONSIDER REPLACING `-XX:+IgnoreUnrecognizedVMOptions WITH A CONDITIONAL
+        jvmArgs(
+                "-XX:+IgnoreUnrecognizedVMOptions",
+                "--add-opens=java.base/java.util=ALL-UNNAMED")
     }
 
     javadoc {
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -96,6 +96,9 @@ sonarqube {
 tasks {
 
     test {
+        testLogging {
+            setExceptionFormat("full")
+        }
         useJUnitPlatform()
         filter {
             includeTestsMatching("*Tests")
@@ -202,12 +205,4 @@ tasks.register<org.junitpioneer.gradle.TriggerTravisTask>("triggerSiteBuild") {
     branch = "grandmaster"
     apiToken = travisApiToken
     message = "Triggered by successful JUnit Pioneer build for %COMMIT"
-}
-
-test {
-    testLogging {
-        // set options for log level LIFECYCLE
-        events "failed"
-        exceptionFormat "full"
-    }
 }
--- a/docs/enviroment-variables.adoc
+++ b/docs/enviroment-variables.adoc
@@ -6,6 +6,8 @@ Both annotations work on the test method and class level, are repeatable as well
 After the annotated method has been executed, the variables mentioned in the annotation will be restored to their original value or will be cleared if they didn't have one before.
 Other environment variables that are changed during the test, are *not* restored.
 
+Warning: Your SecurityManager has to allow modifications.
+
 For example, clearing a environment variable for a test execution can be done as follows:
 
 [source,java]
--- a/docs/enviroment-variables.adoc
+++ b/docs/enviroment-variables.adoc
@@ -1,0 +1,58 @@
+:page-title: @ClearEnvironmentVariable and @SetEnvironmentVariable
+:page-description: JUnit Jupiter extensions to clear and set the values of environment variable
+
+The `@ClearEnvironmentVariable` and `@SetEnvironmentVariable` annotations can be used to clear, respectively, set the values of environment variables for a test execution.
+Both annotations work on the test method and class level, are repeatable as well as combinable.
+After the annotated method has been executed, the variables mentioned in the annotation will be restored to their original value or will be cleared if they didn't have one before.
+Other environment variables that are changed during the test, are *not* restored.
+
+For example, clearing a environment variable for a test execution can be done as follows:
+
+[source,java]
+----
+@Test
+@ClearEnvironmentVariable(key = "some variable")
+void test() {
+	assertThat(System.getenv("some variable")).isNull();
+}
+----
+
+And setting a environment variable for a test execution:
+
+[source,java]
+----
+@Test
+@SetEnvironmentVariable(key = "some variable", value = "new value")
+void test() {
+	assertThat(System.getenv("some variable")).isEqualsTo("new value");
+}
+----
+
+As mentioned before, both annotations are repeatable and they can also be combined:
+
+[source,java]
+----
+@Test
+@ClearEnvironmentVariable(key = "1st variable")
+@ClearEnvironmentVariable(key = "2nd variable")
+@SetEnvironmentVariable(key = "3rd variable", value = "new value")
+void test() {
+	assertThat(System.getenv("1st variable")).isNull();
+	assertThat(System.getenv("2nd variable")).isNull();
+	assertThat(System.getenv("3rd variable")).isEqualsTo("new value");
+}
+----
+
+Note that class level configurations are overwritten by method level configurations:
+
+[source,java]
+----
+@ClearEnvironmentVariable(key = "some variable")
+class MySystemPropertyTest {
+	@Test
+	@SetEnvironmentVariable(key = "some variable", value = "new value")
+	void test() {
+		assertThat(System.getenv("some variable")).isEqualsTo("new value");
+	}
+}
+----
--- a/docs/enviroment-variables.adoc
+++ b/docs/enviroment-variables.adoc
@@ -6,7 +6,10 @@ Both annotations work on the test method and class level, are repeatable as well
 After the annotated method has been executed, the variables mentioned in the annotation will be restored to their original value or will be cleared if they didn't have one before.
 Other environment variables that are changed during the test, are *not* restored.
 
-Warning: Your SecurityManager has to allow modifications.
+WARNING: Java considers environment variables to be immutable, so this extension uses reflection to change them.
+This requires that the `SecurityManager` allows modifications and can potentially break on different operating systems and Java versions.
+Be aware that this is a fragile solution and consider finding a better one for your specific situation.
+If you're running on Java 9 or later, you may have to add `--add-opens=java.base/java.util=ALL-UNNAMED` to your test execution to prevent warnings or even errors.
 
 For example, clearing a environment variable for a test execution can be done as follows:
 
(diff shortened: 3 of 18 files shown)


Hancho2009:
@nicolaiparlog please give me some feedback.

.travis.yml
```diff
@@ -1,9 +1,19 @@
 language: java
 
+os:
+  - linux
+  - osx
```
nicolaiparlog: Cool that we test on an additional OS now! :+1:

.travis.yml
```diff
@@ -1,9 +1,19 @@
...
+
 jdk:
+  - oraclejdk11
   - openjdk8
   - openjdk11
 
+matrix:
+  exclude:
+    - os: osx
+      jdk: openjdk8
```
nicolaiparlog: Why does this combination have to be excluded?

.travis.yml
```diff
@@ -1,9 +1,19 @@
 language: java
 
+os:
+  - linux
+  - osx
+
 jdk:
+  - oraclejdk11
```
nicolaiparlog: Why do we test on another JDK 11?

build.gradle.kts
```diff
@@ -89,6 +89,9 @@ yamlValidator {
 tasks {
 
     test {
+        testLogging {
+            setExceptionFormat("full")
+        }
```
nicolaiparlog: What does this do? :blush:

docs/enviroment-variables.adoc
```diff
@@ -0,0 +1,60 @@
+:page-title: @ClearEnvironmentVariable and @SetEnvironmentVariable
+:page-description: JUnit Jupiter extensions to clear and set the values of environment variable
+
+The `@ClearEnvironmentVariable` and `@SetEnvironmentVariable` annotations can be used to clear, respectively, set the values of environment variables for a test execution.
+Both annotations work on the test method and class level, are repeatable as well as combinable.
+After the annotated method has been executed, the variables mentioned in the annotation will be restored to their original value or will be cleared if they didn't have one before.
+Other environment variables that are changed during the test, are *not* restored.
+
+Warning: Your SecurityManager has to allow modifications.
```
nicolaiparlog: I think we need a clearer warning that this uses reflection and may technically break on any minor Java release or on different OSes. Users should be aware that while this may be the best way to test env variables, it's not actually a good one. ;)

src/main/java/org/junitpioneer/jupiter/EnvironmentVariableUtils.java
```diff
@@ -0,0 +1,116 @@
...
+	 */
+	private static void setInProcessEnvironmentClass(Consumer<Map<String, String>> consumer)
+			throws ClassNotFoundException, NoSuchFieldException {
+		Class<?> processEnvironmentClass = Class.forName("java.lang.ProcessEnvironment");
+		consumer.accept(getFieldValue(processEnvironmentClass, null, "theEnvironment"));
+		consumer.accept(getFieldValue(processEnvironmentClass, null, "theCaseInsensitiveEnvironment"));
+	}
+
+	/*
+	 * Works on Linux
```
nicolaiparlog: And OS X?

src/main/java/org/junitpioneer/jupiter/EnvironmentVariableUtils.java
```diff
@@ -0,0 +1,116 @@
...
+import java.util.Map;
+import java.util.function.Consumer;
+
+import org.junit.jupiter.api.extension.ExtensionConfigurationException;
+
+/**
+ * This class modifies the internals of the environment variables map with reflection.
+ * Warning: If your {@link SecurityManager} does not allow modifications, it fails.
+ */
+public class EnvironmentVariableUtils {
```
nicolaiparlog: I don't think this should be public API. Was that an accident or on purpose?

nicolaiparlog:
Huh, how did I break the build? Looks like Java 11 bytecode is executed on Java 8 - weird. Feel free to take a look at it, otherwise I can do it.

.travis.yml
Hancho2009: I think they working on windows support too but it is not ready as you can see in this [build](https://travis-ci.org/junit-pioneer/junit-pioneer/builds/652600148) [Windows OpenJDK 11](https://travis-ci.org/junit-pioneer/junit-pioneer/jobs/652600151)

.travis.yml
Hancho2009: This combination is not supported. See [here](https://travis-ci.org/junit-pioneer/junit-pioneer/jobs/652617824)
`Expected feature release number in range of 9 to 15, but got: 8`
They only support openJDK 9 to 15 for osx.

.travis.yml
Hancho2009: I think I read that the Implementations can be different so I added this. oraclejdk8 did not work for linux  (I think if a different distro is used than it is possible) and osx.
My guess is that the implementations are probably the same so I am ok with it if you want to remove it.

build.gradle.kts
Hancho2009: The code did not work on linux (and osx I think) and a test faild but only the first line of the exception was shown and not the full stacktrace. You can see it [here](https://travis-ci.org/junit-pioneer/junit-pioneer/jobs/652150699)

> > Task :test
> org.junitpioneer.jupiter.EnvironmentVariableExtensionTests > initializationError FAILED
>     java.lang.NullPointerException at EnvironmentVariableExtensionTests.java:34
> 123 tests completed, 1 failed, 1 skipped

src/main/java/org/junitpioneer/jupiter/EnvironmentVariableUtils.java
Hancho2009: Same as Linux as far as I can tell.

src/main/java/org/junitpioneer/jupiter/EnvironmentVariableUtils.java
Hancho2009: accident 😳

src/main/java/org/junitpioneer/jupiter/EnvironmentVariableUtils.java
```diff
@@ -16,64 +16,66 @@
...
 	}
 
 	private static void modifyEnvironmentVariables(Consumer<Map<String, String>> consumer) {
 		try {
 			tryProcessEnvironmentClassFallbackSystemEnvClass(consumer);
 		}
-		catch (NoSuchFieldException e) {
-			throw new RuntimeException("Could not modify environment variables");
+		catch (ReflectiveOperationException ex) {
+			throw new ExtensionConfigurationException("Could not modify environment variables", ex);
```
Hancho2009: Keep in mind that `ex` is always form the second method never from the first. The best would proably if we can save also the exception from the first method and if both fail somehow add both exceptions here but I do not know a way do that. Maybe we should change the message that reflects this.
'Could not modify environment variables. The attached cause exception is only from the second of the two methods.' 
what do you think?

src/main/java/org/junitpioneer/jupiter/EnvironmentVariableUtils.java
```diff
@@ -16,64 +16,66 @@
...
 	}
 
 	private static void tryProcessEnvironmentClassFallbackSystemEnvClass(Consumer<Map<String, String>> consumer)
-			throws NoSuchFieldException {
+			throws ReflectiveOperationException {
 		try {
 			setInProcessEnvironmentClass(consumer);
 		}
-		catch (NoSuchFieldException | ClassNotFoundException e) {
+		catch (ReflectiveOperationException ex) {
```
Hancho2009: that is new to me - cool change

Hancho2009:
> Huh, how did I break the build? Looks like Java 11 bytecode is executed on Java 8 - weird. Feel free to take a look at it, otherwise I can do it.

I do not know much about gradle and travis-ci
https://docs.travis-ci.com/user/common-build-problems/#my-tests-broke-but-were-working-yesterday
Can you re run this one? https://travis-ci.org/junit-pioneer/junit-pioneer/builds/652682481
I noticed that in the working one there were two osx jobs what is correct and the newer ones that fail only have one.

docs/enviroment-variables.adoc
Hancho2009: You mentioned  in #174 
> - put a warning into the .adoc
> - put a warning into the Javadoc
> - print a warning once per test suite/class/method (?) execution if the extension is active

I did not know excatly how much you wanted here (and where for the last point).
Would at be ok if you do that?

src/main/java/org/junitpioneer/jupiter/EnvironmentVariableUtils.java
Hancho2009: Instead of the cause we could also use to suppressed too better represent that they are on the same level:

```
    public static void main(String[] args) {
        try {
            try {
                throw new ReflectiveOperationException("First try");
            } catch (ReflectiveOperationException e) {
                try {
                    throw new ReflectiveOperationException("Second try");
                } catch (ReflectiveOperationException ex) {
                    RuntimeException runtimeException = new RuntimeException("all");
                    runtimeException.addSuppressed(e);
                    runtimeException.addSuppressed(ex);
                    throw runtimeException;
                }
            }
        } catch (RuntimeException runtimeE)
        {
            runtimeE.printStackTrace();
        }
    }
```
Output:
```
java.lang.RuntimeException: all
	at org.junitpioneer.Test.main(Test.java:13)
	Suppressed: java.lang.ReflectiveOperationException: First try
		at org.junitpioneer.Test.main(Test.java:8)
	Suppressed: java.lang.ReflectiveOperationException: Second try
		at org.junitpioneer.Test.main(Test.java:11)
```

.travis.yml
nicolaiparlog: OpenJDK 11 and Oracle JDK 11 are very similar and I'm willing to bet that they behave the same here, so let's remove this JDK to reduce build times.

build.gradle.kts
nicolaiparlog: Let's keep it, then. :)

docs/enviroment-variables.adoc
nicolaiparlog: Sure.

src/main/java/org/junitpioneer/jupiter/EnvironmentVariableUtils.java
nicolaiparlog: Thought so. Maybe "Works on Unix" is more fitting, then.

src/main/java/org/junitpioneer/jupiter/EnvironmentVariableUtils.java
nicolaiparlog: (A) Great catch about the first exception, which went missing. Adding it as suppressed to the second exception looks right to me because the code that lead to the second exception did indeed suppress the first.

(B) I would still add the second exception as cause to the final (third) one because that causal chain represents well what actually happened.

As far as I can tell the code still does (A) and I would keep it that way.

nicolaiparlog:
I will look into the build failure some other day. :) Thanks for moving this forward, I really appreciate it. :+1:

.travis.yml
```diff
@@ -1,9 +1,19 @@
...
   - openjdk8
   - openjdk11
 
+matrix:
```
Michael1993: If I understand the [documentation](https://docs.travis-ci.com/user/customizing-the-build/#build-matrix) correctly, this should be `jobs:` instead of `matrix:`.

Michael1993: Currently that's my best guess why a build (that should be excluded anyway?) fails.

Michael1993: Nevermind, I'm just being dumb.

.travis.yml
Michael1993: Okay, so, I *think* this [might be causing issues on osx](https://docs.travis-ci.com/user/languages/java/#using-java-10-and-later)?

Hancho2009: I am not sure what you excactly mean.

> OracleJDK 11 and later are supported on Linux, and OpenJDK 10 and later are supported on Linux and macOS using install-jdk.sh.

Do you mean because Oracle JDK 11 is not mentioned for macOS that it is not supported?

It also worked once https://travis-ci.org/junit-pioneer/junit-pioneer/builds/652682481
It could be that travis Ci changed something and because of that it no longer works. So to test this it would be good to re run the previously working build again (I dont know how).
I suspect that because the working build had 2 macOS jobs but in the failing ones there is only 1 macOS job and I think the .travis.yml did not change.

Hancho2009:
Did you remove travis-ci?

.travis.yml
Hancho2009: Seems to work. it should be removed anyway https://github.com/junit-pioneer/junit-pioneer/pull/174#discussion_r393946606

What I do not understand is that is once worked https://travis-ci.org/junit-pioneer/junit-pioneer/builds/652682481
https://travis-ci.org/github/junit-pioneer/junit-pioneer/jobs/652682485/config

Michael1993: `travis_fold:start:install_jdk` was completely missing from the breaking build. It looked like Travis just skipped a step? Don't know why.

src/main/java/org/junitpioneer/jupiter/EnvironmentVariableExtension.java
```diff
@@ -0,0 +1,167 @@
...
+class EnvironmentVariableExtension
+		implements BeforeAllCallback, BeforeEachCallback, AfterAllCallback, AfterEachCallback {
+
+	private static final Namespace NAMESPACE = Namespace.create(EnvironmentVariableExtension.class);
+	private static final String BACKUP = "Backup";
+
+	private static final AtomicBoolean REPORTED_WARNING = new AtomicBoolean(false);
+	// package visible to make accessible for tests
+	static final String WARNING_KEY = EnvironmentVariableExtension.class.getSimpleName();
+	static final String WARNING_VALUE = "This extension uses reflection to mutate JDK-internal state, which is fragile. Check the Javadoc or documentation for more details.";
```
nicolaiparlog: What do you think about the increased visibility?

src/main/java/org/junitpioneer/jupiter/EnvironmentVariableExtension.java
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	private void restoreOriginalEnvironmentVariables(ExtensionContext context) {
+		context.getStore(NAMESPACE).get(BACKUP, EnvironmentVariableBackup.class).restoreVariables();
+	}
+
+	// this method is needed by the tests that verify whether the warning is correctly reported
+	static void resetWarning() {
+		REPORTED_WARNING.set(false);
+	}
```
nicolaiparlog: A bit weird to add this method just for tests. What do you think?

nicolaiparlog:
TODOs for myself:

* add warning to Asciidoc and Javadoc
* check whether all copy/paste from other extension is correct

Bukama:
When merged, please remove the "blocked"-label in #180 so @beatngu13 know, that he can start working on it.

src/main/java/org/junitpioneer/jupiter/EnvironmentVariableExtension.java
Bukama: I don't like methods in prod code just for tests.

src/main/java/org/junitpioneer/jupiter/EnvironmentVariableExtension.java
Bukama: I write all my classes package private to access them in tests

src/main/java/org/junitpioneer/jupiter/EnvironmentVariableExtension.java
nicolaiparlog: Me neither. Fixed it by making the field accessible, so this can be done directly by the tests. (Doesn't really change anything functionally, but not having that method there just feels better. :grin: )

nicolaiparlog:
Hey, @Hancho2009, I finally got around to merging this. Thank you very much for your contribution and patience. :raised_hands:

Hancho2009:
> Hey, @Hancho2009, I finally got around to merging this. Thank you very much for your contribution and patience. 🙌

Thank you too this would not happened without your involvement.

PR #175, opened 2020-02-19
Add extension to disable per test name to be able to disable parameterized tests

nishantvas:
Additional extension to be able to skip tests based on test name
This is particularly helpful when dealing with Parameterized or Dynamically Registered tests

https://github.com/junit-pioneer/junit-pioneer/issues/163

---


nicolaiparlog:
Sorry @nishantvas, I didn't have time to look at this and I'm pretty busy the next weeks. I will get to it eventually, I promise! :)

nishantvas:
I didn't really add this to be merged (yet) it was more to indicate the Idea I was trying to indicate with my Issue

nicolaiparlog:
Hey @nishantvas, this is a pretty cool proof-of-concept! I wasn't aware that the conditional extension is executed for each parameterized test run - very clever to make use of that and the test display name. :+1: Keep working on it, it is definitely going into the right direction.

src/main/java/org/junitpioneer/jupiter/DisableIfNameExtension.java
```diff
@@ -0,0 +1,48 @@
...
+public class DisableIfNameExtension implements ExecutionCondition {
+
+	@Override
+	public ConditionEvaluationResult evaluateExecutionCondition(ExtensionContext context) {
+		Optional<DisableIfName> disable = findAnnotation(context.getElement(), DisableIfName.class);
+
+		if (!disable.isPresent()) {
+			return enabled("No instructions to disable");
+		} else if (!context.getTestMethod().isPresent()) {
+			return enabled("Only disable at method level");
```
nicolaiparlog: We could disable on class level, too, right? Would that make sense?

nicolaiparlog:
Of course we still need examples, documentation, and commit message. In case you prefer only writing the code, that's ok as well, just let me know.

nicolaiparlog:
One more note, should it be `@DisableIfDisplayName`?

Bukama:
> One more note, should it be `@DisableIfDisplayName`?

I vote for this so it's clear by the annotation name that the display name is taken into account and not the "coded name"

Bukama:
> Of course we still need examples, documentation, and commit message. In case you prefer only writing the code, that's ok as well, just let me know.

Want to add Javadoc at least for the annotation interface

src/main/java/org/junitpioneer/jupiter/DisableIfNameExtension.java
nishantvas: The problem with disabling at class level is that this will get disabled before the parameterized tests could get registered.
Hence, all tests will end up getting disabled since the test name will never be calculated for a parameterized test

nishantvas:
One more note, should it  be @DisableIfDisplayName?

Sure, makes sense and is more readable.

I can also work on finalising this(the code is more of less in it's final form) and add more examples, documentations.

How do you mean commit messages? Should I renamed the current commit messages? Is there any standard format?

Bukama:
> One more note, should it be @DisableIfDisplayName?
> 
> Sure, makes sense and is more readable.
> 
> I can also work on finalising this(the code is more of less in it's final form) and add more examples, documentations.

That would be great!


> 
> How do you mean commit messages? Should I renamed the current commit messages? Is there any standard format?

Only the text for the final commit is needed as all the previous commits will get squashed while merging the branch into master - most probably done by @nicolaiparlog when you call it finished.

src/main/java/org/junitpioneer/jupiter/DisableIfNameExtension.java
nicolaiparlog: Yeah, that makes sense. And I just checked that the annotation hs `@Target({ ElementType.METHOD })`, so this should not happen anyways.

nicolaiparlog:
> I can also work on finalising this(the code is more of less in it's final form) and add more examples, documentations.

Great, let me know when you're done and I'll review the PR then.

> How do you mean commit messages? Should I renamed the current commit messages? Is there any standard format?

As Bukama mentioned, all commits will be squashed into one and then applied to master. That commit's message has [a standard format](https://github.com/junit-pioneer/junit-pioneer/blob/master/CONTRIBUTING.md#merging), yes. You can propose a message if you want or I can write it if you prefer - your choice. :)

nishantvas:
I've added documentation and some more examples. Also updated the extension's name

It made more sense to be able to have control to disable multiple tests in a set of Parameterized tests.
However, when I tried to make it a repeatable annotation, the Extension seem to get registered multiple times and was posing problems.

Unfortunately, I'll be away for a while (PC access) @nicolaiparlog , @Bukama If there is something amiss, it's perfectly fine if you guys want to pick that up.. Otherwise, i can take care of it next week or so.

docs/disable-if-display-name.adoc
```diff
@@ -0,0 +1,42 @@
+:page-title: @DisableIfDisplayName
+:page-description: JUnit Jupiter extensions to selectively disable Parameterized tests
```
Bukama, line 2: (multiple) `Parameterized ==> parameterized`

docs/disable-if-display-name.adoc
```diff
@@ -0,0 +1,42 @@
...
+
+The `@DisableIfDisplayName` annotation Junit Jupiter extension that can be selectively used to disable Parameterized tests which are dynamically registered on runtime basis their display name.
```
Bukama, line 4: This reads wired. What about this:

> The `@DisableIfDisplayName` annotation can be used to selectivevly disable parameterized....

docs/disable-if-display-name.adoc
```diff
@@ -0,0 +1,42 @@
...
+The annotations is only supported on test method level for Parameterized tests.
+Unlike the `@Disabled` API provided in Junit which disables the test on first encounter of the annotation
+`@DisableIfDisplayName` is validated before each `@ParameterizedTest` so that each test (name) can be individually evaluated
```
Bukama, line 7: Missing `.` at the end of the sentence.

docs/docs-nav.yml
```diff
@@ -16,3 +16,5 @@
         url: /docs/temp-directory/
       - title: "Vintage @Test"
         url: /docs/vintage-test/
+      - title: "@DisableIfDisplayName"
```
Bukama, line 19: Please move up so it's sorted in alphabetically order

src/main/java/org/junitpioneer/jupiter/DisableIfDisplayName.java
```diff
@@ -0,0 +1,62 @@
...
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.condition.DisabledIf;
+import org.junit.jupiter.api.extension.ExecutionCondition;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.params.ParameterizedTest;
+
+/**
+ * {@code @DisableIfDisplayName} is a JUnit Jupiter extension which can be used to
+ * selectively disable {@link ParameterizedTest} basis their {@link ExtensionContext#getDisplayName()}
```
Bukama, line 27: > basis their

Do you mean `based on their` ?

src/main/java/org/junitpioneer/jupiter/DisableIfDisplayName.java
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * <p>
+ * The extension is an {@link ExecutionCondition} which validates dynamically registered tests
+ * This is highly useful since current {@link Disabled} or {@link DisabledIf} annotations disable
+ * the whole test but not the Parameterized tests selectively
+ *
+ * If it is required that we wish to disable selective tests out of the plethora of dynamically
```
Bukama, line 34: `If it is required to disable selective tests...`

src/main/java/org/junitpioneer/jupiter/DisableIfDisplayName.java
```diff
@@ -0,0 +1,62 @@
...
+ * registered Parameterized tests, then we can utilize the following
+ *
+ * Each repeatable annotation will be processed for each test, and Test will be skipped if
```
Bukama, line 37: `and the test will be`

src/main/java/org/junitpioneer/jupiter/DisableIfDisplayName.java
```diff
@@ -0,0 +1,62 @@
...
+ * any of them evaluate to be true against the display name
```
Bukama, line 38: `evaluates true against the display name.`

src/main/java/org/junitpioneer/jupiter/DisableIfDisplayName.java
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * @since 0.5.6
+ * @see DisableIfNameExtension
+ */
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+@ExtendWith(DisableIfNameExtension.class)
+public @interface DisableIfDisplayName {
+
+	/**
```
Bukama, line 48:
```
Displaynames of the tests cases to be disabled. The whole test case name can be stored as well as sub strings of it. The values will be evaluated [...] by default.
If {code regex} is provided, the string will be evaluated as ... display name.

@return Test case display name
```

src/main/java/org/junitpioneer/jupiter/DisableIfDisplayName.java
```diff
@@ -0,0 +1,62 @@
...
+	 * If, {@code regex} is provided, then the string will be evaluated as
+	 * {@link String#matches(String)} against the display name
+	 * @return Test Case display name
+	 */
+	String[] value();
+
+	/**
+	 * @return if the {@code value} is to be evaluated as regular expression or sub-string
+	 */
+	boolean regex() default false;
```
Bukama, line 60: Suggest to rename it to `isRegEx`. The name `regex` suggests that this field holds the regular expression itselfs, which will be confusing as a `boolean` can't hold such.

src/main/java/org/junitpioneer/jupiter/DisableIfNameExtension.java
```diff
@@ -24,25 +25,31 @@
...
+			return enabled("Only disable at method level so the parameterized tests could be registered");
 		}
-		DisableIfName details = disable.get();
-		boolean toDisable;
-		if (details.regex()) {
-			toDisable = context.getDisplayName().matches(details.value());
-		} else {
-			toDisable = context.getDisplayName().contains(details.value());
+
+		Optional<DisableIfDisplayName> disableIf = findAnnotation(testMethod, DisableIfDisplayName.class);
```
Bukama, line 34: Please update to the new methods, provided by #187 after it is merged

nishantvas: Will wait for #187 to be merged to resolve all the other comments then

nicolaiparlog: We just merged it, @nishantvas.

nicolaiparlog:
Since @Bukama already a close look at this, I'm gonna abstain until his comments were addressed. Let me know when that's done, so I can give a final review.

Bukama:
Hey @nishantvas,
do you know if you may find a moment to address my comments / requested changes?

nishantvas:
Hi @Bukama, apologies I've been occupied in some other things. And didn't have access to my PC.
I'll have this for review by tomorrow. 
I appreciate your patience.

Bukama:
Everything fine, take your time!

src/main/java/org/junitpioneer/jupiter/DisableIfDisplayName.java
```diff
@@ -0,0 +1,62 @@
...
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+@ExtendWith(DisableIfNameExtension.class)
+public @interface DisableIfDisplayName {
+
+	/**
+	 *
+	 * Display names of the test cases to be disabled. The whole test case name can be passed as well as sub string
+	 * The values will be evaluated with {@link String#contains(CharSequence)} by default
+	 * If, {@code regex} is provided, the string will be evaluated with {@link String#matches(String)} against the display name
```
Bukama, line 52:
```suggestion
	 * If, {@code isRegEx} is provided, the string will be evaluated with {@link String#matches(String)} against the display name
```

src/main/java/org/junitpioneer/jupiter/DisableIfDisplayName.java
```diff
@@ -0,0 +1,62 @@
...
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.condition.DisabledIf;
+import org.junit.jupiter.api.extension.ExecutionCondition;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.params.ParameterizedTest;
+
+/**
+ * {@code @DisableIfDisplayName} is a JUnit Jupiter extension which can be used to
+ * selectively disable {@link ParameterizedTest} based on their {@link ExtensionContext#getDisplayName()}
```
Bukama, line 27:
```suggestion
 * selectively disable {@link ParameterizedTest} based on their {@link ExtensionContext#getDisplayName()}.
```

src/main/java/org/junitpioneer/jupiter/DisableIfDisplayName.java
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * <p>
+ * The extension is an {@link ExecutionCondition} which validates dynamically registered tests
```
Bukama, line 30:
```suggestion
 * The extension is an {@link ExecutionCondition}, which validates dynamically registered tests.
```

src/main/java/org/junitpioneer/jupiter/DisableIfDisplayName.java
```diff
@@ -0,0 +1,62 @@
...
+ * This is highly useful since current {@link Disabled} or {@link DisabledIf} annotations disable
+ * the whole test but not the Parameterized tests selectively
```
Bukama, line 32:
```suggestion
 * This is highly useful since current {@link Disabled} or {@link DisabledIf} annotations disable
 * the whole test, but not the Parameterized tests selectively.
```

src/main/java/org/junitpioneer/jupiter/DisableIfDisplayName.java
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * If it is required to disable selective tests out of the plethora of dynamically
+ * registered Parameterized tests, then we can utilize the following
```
Bukama, line 35:
```suggestion
 * registered Parameterized tests, then we can utilize the following.
```

src/main/java/org/junitpioneer/jupiter/DisableIfDisplayName.java
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * Each repeatable annotation will be processed for each test and test will be skipped if
+ * any of them evaluates true against the display name
```
Bukama, line 38:
```suggestion
 * Each repeatable annotation will be processed for each test and the test will be skipped if
 * any of them evaluates true against the display name.
```

src/main/java/org/junitpioneer/jupiter/DisableIfDisplayName.java
```diff
@@ -0,0 +1,62 @@
...
+ */
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+@ExtendWith(DisableIfNameExtension.class)
+public @interface DisableIfDisplayName {
+
+	/**
+	 *
+	 * Display names of the test cases to be disabled. The whole test case name can be passed as well as sub string
+	 * The values will be evaluated with {@link String#contains(CharSequence)} by default
```
Bukama, line 51:
```suggestion
	 * Display names of the test cases to be disabled. The whole test case name can be passed as well as  a sub string. 
	 * The values will be evaluated with {@link String#contains(CharSequence)} by default.
```

nishantvas:
@Bukama, thanks for the proof checking.

nicolaiparlog:
Thanks @nishantvas for the extension, this is really nifty. I added a bunch of changes to the PR, most notably that we can now distinguish between deactivating the test container and the individual invocations. While the code looked like it checked that (with `getTestMethod`) that didn't work for me. If you're ok with my changes, there are two more things that need to be discussed/done:

1. This belongs into the `params` package, but that's a problem because it uses our internal utilities. We (i.e. @junit-pioneer/junit-pioneer-team) have to come up with a solution for this.
2. Instead of `value` and `isRegExp` I'd prefer attributes `contains` and `matches`, e.g. `@DisableIfDisplayName(contains = { "1", "2" })` or `@DisableIfDisplayName(matches = ".*disabled?\\s.*")`. What do you think?

nishantvas:
@nicolaiparlog 

```java
if (!context.getUniqueId().contains("test-template-invocation"))
			return enabled("Never disable parameterized test method itself");
```
This does look relatively neater than using the `getTestMethod()`, didn't know about it. Thanks.


Also, the `matches` / `contains` do make sense. But I didn't want to do that instead of `value` because in this case, we'll have to default both or one of these to { } , which can lead to people adding the annotation on a test with it doing nothing or us having to make one of these fields mandatory. The neater solution to do this would be to have two annotations which felt like an overkill at the time

Also, not using the keyword `value` does make it a bit more readable but the annotation name is descriptive to it's means
`@DisableIfDisplayName("something")` However, of course 
`@DisableIfDisplayName(contains = { "1", "2" })`  looks equally if not perhaps more descriptive.

nicolaiparlog:
Regarding my two points:

1. I moved the classes and found a working, yet ugly solution for the utils problem. Not great, but better than exposing internal APIs to our users. Unless somebody has another idea how to keep utils package-visible without duplicating a lot of code (and tests?), I'll stick with this.
2. I implemented `contains` and `matches` attributes and think that it really helped with usability, although it does make the implementation a bit more cumbersome. We do get a nice feature on top, though: We can not combine substring search and regexp matching. Let me know what you think.

If nobody requires any further changes, I'm ok with this as it is and we can merge it. (Finally, sorry for taking so long with this.) Proposed commit message:

```
Disable parameterized test invocations by display name (#163 / #175)

Jupiter's `@Disabled`-based annotations disable entire test template methods and can thus not be used to disable individual parameterized test invocations. This extension adds that capability by matching based on display name. Display names can be evaluated by searching for substrings or matching against regular expressions (or both).

Because this extension belongs into the `params` package but also needs to access `PioneerAnnotationUtils`, we needed to find a way to make them accessible without making the class public. To prevent duplication, we went with a (clever? abominable?) reflection hack.

Closes: #163
PR: #175
```

Bukama:
As said in stream chat I agree with Pioneer. So I'm fine with the current state of the PR too.

nishantvas:
Yea the code change seems perfectly alright. The run time check for no values with annotation usage is quite alright. As long as it's documented. Otherwise it can be a tad misleading.

But I'm happy with this. Thanks for the changes

PR #173, opened 2020-02-18
Basic implementation of the ReportEntry feature.

Michael1993:
Closes #134 

---

I hereby agree to the terms of the [JUnit Pioneer Contributor License Agreement](https://github.com/junit-pioneer/junit-pioneer/blob/master/CONTRIBUTING.md#junit-pioneer-contributor-license-agreement).

PR #183, opened 2020-03-11
AsciiDoc documentation of the @ReportEntry extension

Michael1993:
Depends on #173
---


docs/report-entries.adoc
```diff
@@ -0,0 +1,79 @@
...
+@Test
+@ReportEntry("foo")
+@ReportEntry("bar")
+void reportingTest() {
+    // YOUR TEST CODE HERE
+}
+----
+
+You can also declare your own keys in the annotation (if you don't declare it, JUnit
+will use the value of the entry as the key).
```
Bukama: Does ist really use the value of the entry as the key or does it use the string `value` as the key?

src/main/java/org/junitpioneer/jupiter/ReportEntry.java
```diff
@@ -0,0 +1,48 @@
...
+@ExtendWith(ReportEntryExtension.class)
+public @interface ReportEntry {
+
+	/**
+	 * Specifies the key of the pair that's to be published as a report entry.
+	 * Defaults to {@code "value"} and can't be blank.
+	 *
+	 * @see org.junit.jupiter.api.extension.ExtensionContext#publishReportEntry(String, String) ExtensionContext::publishReportEntry
+	 */
+	String key() default "value";
```
Bukama: I mean this: The String "value" is used as default.

docs/report-entries.adoc
Michael1993: You are right, the new solution does use the String "value" as a key if there is no key specified. My original solution used `publishReportEntry(String value)` <- that does behave this way.

I'll rewrite it.

src/main/java/org/junitpioneer/jupiter/ReportEntry.java
Michael1993: Yes, even the unit tests show that you are correct.

docs/report-entries.adoc
```diff
@@ -65,8 +65,8 @@ void reportingTest() {
 }
 ----
 
-You can also declare your own keys in the annotation (if you don't declare it, JUnit
-will use the value of the entry as the key).
+You can also declare your own keys in the annotation (if you don't declare it, the
+extension will use 'value' as the key).
```
Bukama: This should be correct. Thanks for editing it.

PR #185, opened 2020-03-17
Github actions

aepfli:
This is just a draft and a simple idea for now

the build steps are working, it uploads to sonarqube, and it triggers correctly the build. i need to verify the deploy (or depoy :)) step, as there is a lot going on, i am not aware of, or have not tried so far. But i thought this might be useful.

Any ideas welcome

image

- https://github.com/aepfli/junit-pioneer/pull/4 as an example
---

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -1,0 +1,65 @@
+# This workflow will build a Java project with Gradle
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-gradle
+
+name: JUnit Pioneer Flow
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: '*'
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        # test against latest update of each major Java version, as well as specific updates of LTS versions:
+        java: [ 8, 11 ]
+        os: [ubuntu-latest, windows-latest, macos-latest]
+    name: ${{ matrix.os }} - Java ${{ matrix.java }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup java
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java }}
+      - name: Gradle build
+        uses: eskatos/gradle-command-action@v1
+        with:
+          arguments: --refresh-dependencies --stacktrace --scan clean build
+      - name: Sonarqube
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+        if:  matrix.java == '8' && matrix.os == 'ubuntu-latest'
+        run: ./gradlew jacocoTestReport sonarqube --stacktrace -i
+
+  # This will be handled on a later stage
+  # deploy:
+  #   runs-on: ubuntu-latest
+  #   name: Deploy
+  #   if: github.ref == 'refs/heads/master'
+  #   steps:
+  #     - uses: actions/checkout@v2
+  #     - name: Setup java
+  #       uses: actions/setup-java@v1
+  #       with:
+  #         java-version: 8
+  #     - name: Perform Release
+  #       uses: eskatos/gradle-command-action@v1
+  #       if:
+  #       with:
+  #         arguments: ciPerformRelease
+  #     - name: Upload Release to GitHub
+  #       uses: softprops/action-gh-release@v1
+  #       if: startsWith(github.ref, 'refs/tags/')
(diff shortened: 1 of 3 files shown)


.github/workflows/actions.yml
```diff
@@ -0,0 +1,63 @@
...
+  pull_request:
+    branches: '*'
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        # test against latest update of each major Java version, as well as specific updates of LTS versions:
+        java: [ 8, 9, 10, 11, 12, 13]
```
aepfli:
```suggestion
        java: [ 8, 9, 10, 11, 12, 13, 14]
```

Bukama:
Maybe @sormuras can throw an eye on it - When I remember correctly he set up the github actions for Junit but at least for his BACH project

.github/workflows/actions.yml
sormuras: I'd only keep: `8`, `11`, and introduce `14` as of today.

.github/workflows/actions.yml
```diff
@@ -0,0 +1,63 @@
...
+        uses: eskatos/gradle-command-action@v1
+        with:
+          arguments: --refresh-dependencies --stacktrace --scan clean build
+      - name: Sonarqube
+        env:
+          SONAR_LOGIN: ${{ secrets.SONAR_TOKEN }}
+        if:  matrix.java == '8' && matrix.os == 'ubuntu-latest'
+        run: ./gradlew jacocoTestReport sonarqube --stacktrace -i
+
+  depoy:
```
sormuras: `deploy`, with `l`

.github/workflows/actions.yml
aepfli:
```suggestion
        java: [ 8, 11, 14]
```

aepfli:
i am really thinking of dropping my Travis CI pr instead of this, this seems to be way easier, as nearly all configuration is takin gplace in github

1. https://github.com/junit-pioneer/junit-pioneer/actions we need to create an action here
2. we would need to provide the sonar token for the project in https://github.com/junit-pioneer/junit-pioneer/settings/secrets

and we would be done, this would be working
1. i would still remove the deployment steps for now in Github Actions as this needs more attention
2. furthermore i would deactivate the default builds in travis.yml so we do not use so many resources.

Thoughts?
@Bukama @nicolaiparlog

Bukama:
I appreciate dropping a "3rd party" tool for better / easier integration with the plattform we already use as long as the provided functionality stays the same. This seems to be the case here. 

So I would vote for building up GitHub actions parallel to Travis- When the new script works we should disable Travis and after a (short) period of time dropping this back up scenario in total.

aepfli:
i created an own example project with the setup :) https://github.com/aepfli/SonarCloud-GitHubActions-Gradle-example/pull/2 there we see all functionality and basic configuration needed

.travis.yml
```diff
@@ -18,8 +17,8 @@ install: ./gradlew -version
 
 #Don't build tags
 branches:
-  except:
-  - /^v\d/
+  only:
```
aepfli: as we will have githubactions for normal builds, we we only need this for the special case of "deployments" which should only happen on master

.github/workflows/actions.yml
```diff
@@ -0,0 +1,65 @@
...
+        os: [ubuntu-latest, windows-latest, macos-latest]
+    name: ${{ matrix.os }} - Java ${{ matrix.java }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup java
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java }}
+      - name: Gradle build
+        uses: eskatos/gradle-command-action@v1
```
sormuras: Why use an external action here? We already have the Gradle Wrapper in-place and ready to rock.

aepfli: i thought actually to be the save way, as it also uses the gradle wrapper, but i do not need to care about write permissions etc. - i saw that in junit5 and i found it odd. i did not test it without it. but this is something which we definitly can investigate

aepfli: ah damn, that is something i have seen somewerhe else sorry for wrong reference to junit5

aepfli: also there is https://github.com/eskatos/gradle-github-actions-plugin which might be useful in the future

aepfli: image
no need to specify the gradle wrapper as the gradlew is still in the root directory

sormuras: Okay. Let's use `eskatos/gradle-command-action` and see what it brings to the party. Might consider using it for JUnit 5 as well.

/cc @marcphilipp @eskatos

PR #200, opened 2020-03-27
Moving gradle wrapper to default location

### LINKED ISSUE #190: Config files are not discovered by tools
I just updated IntelliJ to the current Community 2020.1 BETA using the IntelliJ toolbox. When I tried to build pioneer I saw that IntelliJ was installing `Gradle 6.1` and the build failed because the `com.gradle.build-scan` plugins version (`2.4.2`) is not compatible to `Gradle 6.x` (for this version the plugin must be switched to https://docs.gradle.com/enterprise/gradle-plugin/ ).

While analyzing the problem together with @nicolaiparlog  and @aepfli 
 we found out that IntelliJ doesn't recognize the `gradle-wrapper.properties` inside the `.infra/gradle` folder as the following images shows.

gradlewtf

So it used the default `Gradle` installation I installed manually some months ago and which was a `5.x`. With the automatically installed and configured `6.1` version the build broke. To get pioneer build again I had to switch the gradle installation in IntelliJ settings to the specific `5.x` one.

While chatting we remembered a similar issue (#127) with Visual Studio Code.

So we coincided to open an issue to collect further examples and to discuss the current config file locations inside `.infra`. As of now they are in there to keep the root folder of the project clean and gather all infrastructure thigs inside a specific folder. But as shown this is ot the expected default folder for some IDE so the build may get broken if the by default installed tool versions are not fitting to pioneer.

Options we see so far:
* Move config files to expected default positions (and mess up the root folder)
* Update documentation to point contributers to the situation

Bukama:
Title says all ;)

With this location IntelliJ could determine the correct version automaticall with both settings (using wrapper.properties and executing warpper-task) which was not the case before the change.

fixes #190

---


build.gradle.kts
```diff
@@ -184,7 +184,7 @@ publishing {
 }
 
 tasks.wrapper {
-    jarFile = file(".infra/gradle/gradle-wrapper.jar")
+    jarFile = file("gradle/wrapper/gradle-wrapper.jar")
```
aepfli: i think this is the default so you actually do not need to set it at all - actually we do not need the whole block, when we use default settings

Bukama: > i think this is the default so you actually do not need to set it at all - actually we do not need the whole block, when we use default settings

I don't know gradle very well but maybe this is needed for the option to use the gradle setting via a wrapper task. At least IntelliJ offers such an option beside `use default wrapper.properties location` and `use own location`.

aepfli: https://docs.gradle.org/current/dsl/org.gradle.api.tasks.wrapper.Wrapper.html#org.gradle.api.tasks.wrapper.Wrapper:jarFile

this is the default value, and the gradle wrapper task is something, which gradle provides out of the box since verison 5 i think (do not pin me on that). but you can find that information:

> Every vanilla Gradle build comes with a built-in task called wrapper. You’ll be able to find the task listed under the group "Build Setup tasks" when listing the tasks. Executing the wrapper task generates the necessary Wrapper files in the project directory. -  https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:adding_wrapper

To test this, you can simply remove the block and run  `gradle wrapper` in your cli and see, it will work out of the box

Bukama: Ok removed the block and still seems to work. Thanks for the docs!

aepfli:
normally there should be also a change in the gradlew and gradlew.bat as the path has changed, and there is a different path - i am currently not sure. but we need to check those files before merging

Bukama:
> regenerate gradlew and gradlew.bat with `gradle wrapper` as they are pointing currently to the `.infra` folder
> 
> https://github.com/Bukama/junit-pioneer/blob/issue190/movegradleconfig/gradlew#L83
> 
> https://github.com/Bukama/junit-pioneer/blob/issue190/movegradleconfig/gradlew.bat#L82

Done. Thanks again

nicolaiparlog:
Rebased on master to include the 6.3 update. Seems to work, thanks @Bukama and @aepfli. :+1:

PR #204, opened 2020-03-27
Updating docs with a guide to create and work with a fork

Bukama:
To avoid branch conflicts like we had (I think) two weeks ago and do help new contributors to get to work I've updated the docs by adding a guide on how to create and work with a fork.

As this was spontaneous idea there's no issue for this addional documentation.

---

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -75,6 +75,8 @@ The [pull requests template](.github/PULL_REQUEST_TEMPLATE.md) contains a footer
 
 To enforce the [branching strategy](#branching-strategy) pull requests from `master` will be closed.
 
+See the [appendix](#appendix) to create a fork, make changes and open a pull request.
+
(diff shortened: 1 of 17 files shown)


aepfli:
maybe we should also reference https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/working-with-forks

Bukama:
> maybe we should also reference https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/working-with-forks

Added, thanks

nicolaiparlog:
I'm not convinced that `CONTRIBUTING.md` is the best place for this how-to. I prefer the file to stay lean and only discuss project specifics. If we start explaining the underlying tools, the file will be huge. Let's discuss in #205.

Bukama:
Hm after the rebase all files that were rebased are shown as changed... I'll create a new PR

PR #206, opened 2020-03-29
Fixing sonarcloud

aepfli:
As sonarcloud is currently not working, and
as long as we do not have time to proper
investigate we might want to deactivate it.

---

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -28,12 +28,12 @@ jobs:
         uses: eskatos/gradle-command-action@v1
         with:
           arguments: --refresh-dependencies --stacktrace --scan clean build
-  #    - name: Sonarqube
-  #      env:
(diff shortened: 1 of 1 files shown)


nicolaiparlog:
Any idea how long a proper fix would take? I don't mind having a "broken" project for a few days.

aepfli:
Issue:
It seems like GitHub Actions is currently not as advanced as Travis,
[as it does not allow to provide Secrets for a workflow to a fork.](https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#using-encrypted-secrets-in-a-workflow) 

This means, sonarcloud works fine on branches and pullrequests within the same repository,
but not within forks. As the forks will never know the sonar token.

The only way to achieve this currently is by adding the token hardcoded into the
workflow file for GitHub Actions. We could also add it within the gradle script,
but i think it is better to keep it out of there and leave it just in the CI configuration

PR #207, opened 2020-03-29
Bogus PR for Sonar experiments

nicolaiparlog:
Experimental PR to make Sonar work.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# JUnit Pioneer
+# JUnit Pioneer 10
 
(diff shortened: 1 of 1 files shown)


nicolaiparlog:
We found out that GitHub Actions does not make secret tokens available to builds started in forks. This is a security feature and can be observed in the log output where only GitHub's token has a `***` next to it while others, including SonarQube's, are blank. That's why the Sonar integration works in this PR (branch in origin repo), but not in general ones.

Issue #199, opened 2020-03-27
Github Actions for Deployments


aepfli:
JUnit Pioneer currently still relies on travis for the releasing process.

Goal of this ticket is to also migrate this deployment and releaseing process to GitHub Actions

We want to migrate following code to GitHub Actions, maybe in an own build just for master, or as a sub step of the existing  #actions.
https://github.com/junit-pioneer/junit-pioneer/blob/4f3647f88bb3131b19fa77efaac1eb2edf11962c/.travis.yml#L29-L41

This issue is part of #192
and relates to #193

nicolaiparlog:
This effectively removes Shipkit, right? If so, can we discuss this in #193 first?

aepfli:
question is what we favor first. Evaluating shipkit or moving the Build infrastructure to github actions :) - both can be done at the same time. and i see the testing part as the more crucial one, which is taking more time than actually implementing it :)

aepfli:
i opened the Pull Request, for using this with shipkit, and so far i fell in love with Shipkit and how they use releasing. We just had a lot of additional steps in between.

There is just one down side as described in #171 - if we want to add our jars to the release on github releases, we need to put in manual effort. I skipped that for now, because this was never really working so we do not lose functionality, and it seems like this is nothing which shipkit provides as functionality. Neither Shipkit itself, nor PowerMock or Mockito are uploading jars to the release page of github. But they are linking it anyways in the release text.

Issue #192, opened 2020-03-26
Move from Travis CI to GitHub Actions


nicolaiparlog:
GitHub Actions seem the be hot new thing and everybody's using it, including JUnit 5 itself, so let's do that, too. (Yes, that's all the reason I need to give it a try. :) )

nicolaiparlog:
@aepfli started working on this is #185.

Bukama:
Just wanted to add: We already discussed that on Discord and in chat. See also comments on #181 and #185

aepfli:
I created #199 and #201 as the next steps for this ticket. We should keep this ticket also for the cleanup, so we can than remove travis.yml when all the others are finished

Bukama:
Don't forget #28 (it was already linked to #50)

aepfli:
uh 28 is done with sonarcloud, when it finally works :)

nicolaiparlog:
As pointed out by #208, we also need to update the badges in the README once we don't build on Travis anymore.

nicolaiparlog:
It looks like there are two more things to do before we can remove Travis entirely:

* trigger ShipKit via GitHub Actions - see #199 (we can ignore the discussion around ShipKit in #192 for now)
* trigger site built via GitHub Actions - see junit-pioneer/junit-pioneer.github.io#5

Since the site built is currently broken anyways, we're not loosing functionality if we move to GitHub Actions without reimplementing it. I.e. once #199 is merged, we can do the necessary clean-up mentioned above and then remove Travis entirely.

nicolaiparlog:
Closed by 80b3e45

Issue #193, opened 2020-03-26
Evaluate ShipKit


nicolaiparlog:
We currently use ShipKit to deploy every commit. This works well enough, but there are a few things we may want to conisder:

* We move to GitHub Actions (see #192), which also offers shipping - why not consolidate the number of tools?
* The `version.properties` file in the top-level folder is annoying me (see #96); if we stick with ShipKit, I need to open an issue to see whether that could be changed.
* There's some POM-related duplication between `build.gradle` and `gradle/shipkit.gradle`, which is a bit annoying as well. Can this be fixed?
* ShipKit's documentation only shows Travis integration (which we move away from), but [mentions](https://github.com/mockito/shipkit/blob/master/docs/how-shipkit-works.md) "it's not difficult to use Shipkit with any CI" - how?
* ShipKit's approach of continuous deployment occasionally leads to spurious releases where we forgot to add `[ci skip-release]` to a commit message (not great to clutter them with that either) and the tool thought something changed when it didn't (e.g. 0.5.5 / 22f2571). That's not good.
* we occasionally had race conditions for `version.properties` where several PRs' changes of this file conflicted and needed to be updated
* For more to ponder, see #97, #171, and mockito/shipkit#867.

What does everybody think?

nicolaiparlog:
Whatever we end up doing, we probably need to update [`CONTRIBUTING.md`-'s section on ShipKit](https://github.com/junit-pioneer/junit-pioneer/blob/master/CONTRIBUTING.md#versioning).

aepfli:
i am just taking a closer look into shipkit, and i think this tooling is fine. but i also think,  we do have some code duplciation. I mean we have configured a bintray plugin, which has a maven central sync. And (i might be wrong here) is maven central not oss.sonatype.org ?? because our publication is than redundant, if we really do have a sync from bintray setup to maven central.

Hence that i do not think we need the publication step in our build gradle. 

For the rest i will investigate, but if we stick to shipkit i think we need to stick to version.properties :)

beatngu13:
My two cents:

I believe @nicolaiparlog once brought it up during a stream, if making a release is as easy as creating a tag, why struggle with CI skips etc. to avoid accidental releases? Especially since we as library maintainers somewhat still have the concept of releases that may bundle multiple patches—in contrast to e.g. web service development, where one almost always wants to continuously deploy. (See also #211 by @Bukama.)

[...]

PR #213, opened 2020-03-29
GitHub Action for Deployments

aepfli:
Adding GitHub Actions for Deployment which will run
on Linux and with JDK 8. On the side i removed a lot
of unnecessary gradle code for publishing.

To get this running within the base repository, we
need to add following secrets:
- GH_WRITE_TOKEN: for updating the repo and setting the tag
- BINTRAY_USER_NAME: for upload to bintray
- BINTRAY_API_KEY: for upload to bintray
- NEXUS_TOKEN_USERNAME: for syncing bintray to maven central
- NEXUS_TOKEN_PASSWORD: for syncing bintray to maven central

:warning: We need to be careful with this, as this can break the releasing process :warning: 

---


Bukama:
If I apply my basic Jenkins pipeline knowledge this PR looks good to me.

Bukama:
> If I apply my basic Jenkins pipeline knowledge this PR looks good to me.

I have to revert my word. Please update the documentation.

>> Like Mockito, JUnit Pioneer implements a continuous delivery model using Shipkit and Travis CI. 

This is no longer true. But I could live with it, when we decide / progress about #193

aepfli:
> > If I apply my basic Jenkins pipeline knowledge this PR looks good to me.
> 
> I have to revert my word. Please update the documentation.
> 
> > > Like Mockito, JUnit Pioneer implements a continuous delivery model using Shipkit and Travis CI.
> 
> This is no longer true. But I could live with it, when we decide / progress about #193

actually i would wait till the whole travis migration is done, and put that into #192 as a cleanup step. currently we are in a slight transition phase, and we need to do some cleanup afterwards anyways. when the page sync is done

nicolaiparlog:
I got the impression that this is blocked by #193, so I marked it as such.

aepfli:
> I got the impression that this is blocked by #193, so I marked it as such.

the question is, do we favor getting away from travis more than evaluating shipkit. i owuld first proceed with migration of github actions, and than do the evaluation. anyway we need to adapt the buildscripts on a later stage depending on the outcome of shipkit evaluation. i would try to use one CI tool first currently.

nicolaiparlog:
Yeah, I misunderstood the situation. We can go forward with this and change the ShipKit situation separately.

nicolaiparlog:
Finally took some time to go through this. Added all the five tokens you mentioned above under the names you need them. But they all come from an old dusty safe and I have no idea whether they're exactly what you need (e.g. I have a Sonatype token in my safe - let's assume it's the NEXUS_TOKEN we need here) and even still work. Let's find out!

aepfli:
@nicolaiparlog now comes the part, where i am not 100% sure how to handle that in the best way. How do i test it now, if everything works as intended on the pioneer repo. As i am quiet new to releasing with shipkit bintray and jcenter, i suggest we coordinate on this.

i do not want to start with creating new releases of junit-pioneer. Therefor if it is okay, i suggest i shortly will change the name of the artifact to `junit-pioneer-test` and deactivate the automatic sync.

We then can check the jars and poms, if everything is still in place, and i would suggest to continue after that, by setting the right version. we could theoretically create a second branch like `release/testBintray` which would use this branch as a basis. So the git history within this branch would not be filled with unnecessary changes.

So to summarize:
1. create a branch based on that with a naming, which would invoke a shipkit release
    - jcenter sync has to be deactivated in the initial commit
    - the artifact name should be overwritten, or the version should have something like `SNAPSHOT` in its name
2. trigger the pipeline which will result in a deployment
3. checking the files on bintray if they are in a good shape
    - correct pom entries
    - working jar
4. merge this pullrequest and delete the interims branch.

what do you think?

aepfli:
okay so lets write a good commit message:

```
GitHub Action for Deployments 

Adding GitHub Actions for Deployment which will run
on Linux and with JDK 8. This step allows us to remove
travis completely from the project. Additionally we also
remove appveyor as it is also redundant with GitHub 
Actions now.

We additionally removed all references to travis and
replaced them with new refernces to GitHub Actions.

As the Sitebuild was also broken, we also removed this
implementation and we will in the next step start 
implementing a new SiteBuild with GitHub Actions, and
a trigger will than be added to the actions within this repo

Closes: #199 
PR: #213 
Relates: #192 

```

Bukama:
> Closes: #199 
> PR: #213
> Relates: #192 

Also closes #234 , doesn't it ?

aepfli:
> > Closes: #199
> > PR: #213
> > Relates: #192
> 
> Also closes #234 , doesn't it ?
> 
> Why is #192 not closed? For the site deployment there is an own issue in the repo?

which suitebuild issue, there is the one within the junitpioneer page, but in here, we also will need to configure the trigger, hence that i would not close #192 till this last piece is done

.github/workflows/actions.yml
```diff
@@ -45,42 +47,44 @@ jobs:
...
+      - name: Setup java
+        uses: actions/setup-java@v1
+        with:
+          java-version: 8
+      - name: Perform Release
+        uses: eskatos/gradle-command-action@v1
+        env:
+          TRAVIS_BRANCH: ${{ steps.gitbranch.outputs.replaced }}
+          TRAVIS_COMMIT_MESSAGE: ${{ github.event.head_commit.message }}
+          TRAVIS_PULL_REQUEST: ${{ github.event_name == 'pull_request' }}
```
beatngu13, line 79: What do we need these environment variables for?

build.gradle.kts
```diff
@@ -148,67 +137,4 @@ tasks {
...
-    }
-}
-
-tasks.register<org.junitpioneer.gradle.TriggerTravisTask>("triggerSiteBuild") {
-    travisProject = "junit-pioneer/junit-pioneer.github.io"
-    branch = "grandmaster"
-    apiToken = travisApiToken
-    message = "Triggered by successful JUnit Pioneer build for %COMMIT"
-}
+}
```
beatngu13, line 140:
```suggestion
}

```

.github/workflows/actions.yml
```diff
@@ -45,42 +47,44 @@ jobs:
...
+          GH_WRITE_TOKEN: ${{ secrets.GH_WRITE_TOKEN }}
```
beatngu13, line 80: What is the difference to `GITHUB_TOKEN` above? If there is none, I would make the environment variables globally available. Otherwise, I would use the same prefix for consistency reasons (i.e. "GH" or "GITHUB").

.github/workflows/actions.yml
```diff
@@ -45,42 +47,44 @@ jobs:
...
-  #       with:
-  #         arguments: ciPerformRelease
-  #     - name: Upload Release to GitHub
-  #       uses: softprops/action-gh-release@v1
-  #       if: startsWith(github.ref, 'refs/tags/')
-  #       with:
-  #         files: ./build/libs/junit-pioneer-*.jar
-  #       env:
-  #         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  deploy:
```
beatngu13, line 56: If we want to, we can also use a separate YAML file for this?

.github/workflows/actions.yml
aepfli: that is the funny part, shipkit is only compatible with travis currently, and needs those travis environment variables. So i "fake" to be travis to get shipkit working, till they hopefully accept my proposal, and i can finish https://github.com/mockito/shipkit/pull/868

.github/workflows/actions.yml
aepfli: i was also thinking about this, and maybe trigger this from the build script. but i think we should always first run the build, and see if there are no errors, and than deploy.

This might be a good improvement for the future to find a sophisticated way of running both actions in sequence but conditionally on a success of the first one. I did not investigate that far into actions.

.github/workflows/actions.yml
aepfli: good point @nicolaiparlog can you rename the secret and add edit this line afterwards?

The difference is, that the `GITHUB_TOKEN` should not have any write access just read access, where ass the (soon to be) `GITHUB_WRITE_TOKEN` also has write access

beatngu13: Ah, I see. Then maybe `GITHUB_READ_TOKEN` and `GITHUB_WRITE_TOKEN`? (Although I wonder what a read-only token on a public project can do that I cannot do without a token at all?)

aepfli: i renamed the environment variable, we just need to change the name of the secret

aepfli: i would not rename the other one, as it seems to be a default not 100% assinged to the repo but also to each user https://help.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token

aepfli:
this pr just recreates the status quo regarding deployment, there is no alternative  merge behaviour with tags included. This is something we do with #193 and we will handle that afterwards.

nicolaiparlog:
Err, I can't add a `GITHUB_WRITE_TOKEN`. When I enter the needed information and hit _add Secret_, the _Add new secret_ form disappears, but there's no new token. I can add that same token under another name, but whatever token value I use, I simply can't create a token with that name. Weird...

nicolaiparlog:
As I understand, the new token name is a cosmetic improvement and everything already works, right? In that case, I'll merge.

aepfli:
just a sec, with merge please, i check if it works, if you do not rename

aepfli:
good to go!

PR #214, opened 2020-04-01
Improve GitHub Actions workflow

beatngu13:
Improves the current GitHub Actions workflow by several means:

* Cache Gradle and Sonar files, see Travis example e.g. [here](https://github.com/SonarSource/sq-com_example_java-gradle-travis/blob/master/.travis.yml#L17).
* Fetch unshallow to enable blame for Sonar, see warning e.g. [here](https://github.com/junit-pioneer/junit-pioneer/runs/550297033?check_suite_focus=true#step:5:357).
* Use step names consistently.

~Also note that according to the [SonarCloud GitHub Action](https://github.com/SonarSource/sonarcloud-github-action#do-not-use-this-github-action-if-you-are-in-the-following-situations), we should _not_ use the action itself but the SonarQube plugin for Gradle during the build. I'm a Gradle noob, but can have a look later on if nobody is quicker?~

Realized that this is already done. 👍

Proposed commit message:

```
Improve GitHub Actions workflow

* Cache Gradle and Sonar files.
* Fetch unshallow to enable blame for Sonar.
* Use step names consistently.

PR: 214
```

---

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -20,6 +20,9 @@ jobs:
     name: ${{ matrix.os }} - Java ${{ matrix.java }}
     steps:
       - uses: actions/checkout@v2
+        # fetch unshallow to enable blame for Sonar
+      - name: Fetch unshallow
+        run: git fetch --prune --unshallow
       - name: Setup java
         uses: actions/setup-java@v1
         with:
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -24,6 +24,11 @@ jobs:
         uses: actions/setup-java@v1
         with:
           java-version: ${{ matrix.java }}
+      - name: Cache Sonar
+        uses: actions/cache@v1
+        with:
+          path: ~/.sonar/cache/
+          key: ${{ runner.os }}-sonar
       - name: Gradle build
         uses: eskatos/gradle-command-action@v1
         with:
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -28,6 +28,12 @@ jobs:
         uses: actions/setup-java@v1
         with:
           java-version: ${{ matrix.java }}
+      - name: Cache Gradle
+        uses: actions/cache@v1
+        with:
+          path: ~/.gradle/caches/
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*') }}
+          restore-keys: ${{ runner.os }}-gradle-
       - name: Cache Sonar
         uses: actions/cache@v1
         with:
@@ -56,6 +62,12 @@ jobs:
   #       uses: actions/setup-java@v1
   #       with:
   #         java-version: 8
+  #     - name: Cache Gradle
+  #       uses: actions/cache@v1
+  #       with:
+  #         path: ~/.gradle/caches
+  #         key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*') }}
+  #         restore-keys: ${{ runner.os }}-gradle-
   #     - name: Perform Release
   #       uses: eskatos/gradle-command-action@v1
   #       if:
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -3,7 +3,11 @@
 
 name: JUnit Pioneer Flow
 
-on: [push, pull_request]
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: '*'
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


.github/workflows/actions.yml
```diff
@@ -3,11 +3,7 @@
 
 name: JUnit Pioneer Flow
 
-on:
-  push:
-    branches: [ master ]
-  pull_request:
-    branches: '*'
+on: [push, pull_request]
```
aepfli, line 6: i do not think that we really need a build on each push to all branches. `master` alone is fine. Based on shipkit we also do not want build on tags i think.

.github/workflows/actions.yml
```diff
@@ -19,11 +15,26 @@ jobs:
         os: [ubuntu-latest, windows-latest, macos-latest]
     name: ${{ matrix.os }} - Java ${{ matrix.java }}
     steps:
-      - uses: actions/checkout@v2
+      - name: Check out repo
+        uses: actions/checkout@v2
+        # fetch unshallow to enable blame for Sonar
+      - name: Fetch unshallow
```
aepfli, line 21: not sure if we need an extra step we could also inline this into the checkout step - https://github.com/actions/checkout#fetch-all-history-for-all-tags-and-branches

.github/workflows/actions.yml
```diff
@@ -19,11 +15,26 @@ jobs:
...
         uses: actions/setup-java@v1
         with:
           java-version: ${{ matrix.java }}
+      - name: Cache Gradle
+        uses: actions/cache@v1
+        with:
+          path: ~/.gradle/caches/
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*') }}
+          restore-keys: ${{ runner.os }}-gradle-
+      - name: Cache Sonar
```
aepfli, line 33: sonar only runs on ubuntu and only on one JDK - maybe we should take this into consideration.

.github/workflows/actions.yml
beatngu13: AFAIK the first step does the actual checkout with a bunch of optimizations:

https://github.com/beatngu13/junit-pioneer/runs/551399103?check_suite_focus=true#step:2:1

Whereas the second step only fetches the history:

https://github.com/beatngu13/junit-pioneer/runs/551399103?check_suite_focus=true#step:3:1

I don't think we can use one without the other? Unless we do the checkout manually and avoid setting depth to 1 (like [here](https://github.com/beatngu13/junit-pioneer/runs/551399103?check_suite_focus=true#step:2:26)).

Probably related to this feature request:

https://github.com/actions/checkout/issues/204

.github/workflows/actions.yml
beatngu13: Good catch, thx! Will fix that.

.github/workflows/actions.yml
aepfli: maybe i miss understood the logic in the readme of the action
```
- uses: actions/checkout@v2
- run: |
    git fetch shallow--prune --un
```

i thought based on this we can just do

```
  - name: Check out repo
        uses: actions/checkout@v2
        run: |
             git fetch shallow--prune --un
```

But it also seams like i miss read yaml once again, and ignored the `-` which i just realized by writing this comment :)

.github/workflows/actions.yml
beatngu13: What is the downside of building on each push? If we don't touch the parallelization limit such that jobs are queued/blocked, I would personally find this helpful as I cannot locally verify the build matrix. As a contributor, I wish to get that feedback _before_ opening a PR.

Regarding Shipkit: unfortunately, I haven't really worked with it so far. This might be an issue because don't know how these CI skips work (neither in general nor in combination with GHA).

I would suggest to revert this commit and discuss that separately, maybe also in the course of #193.

aepfli: I have to admit, i did not take the contributor part into consideration, but, i am also thinking about resource usage etc. and i am not sure, if we really need a build on 3 jdks with 3 OS for a branch in development.

The thing we really should take care of, is to not build on a _tag_ push. Shipkit will create a tag and push it, hence that the build on master will trigger a another build, which is imho not needed at all. Hence that i would exclude tags.

`[ci skip]` is not working on GHA as it is a travis default
`[ci skip-release]` is something [offered by skipkit ](https://github.com/mockito/shipkit/blob/master/docs/features/merge-pull-requests.md#user-guide)

we could although configure our own "commit postfix" to prevent a release - to make it easier for @nicolaiparlog to not trigger a build :)

PR #218, opened 2020-04-04
Implement PioneerTestKit and PioneerEngineExecutionResults (#6)

### LINKED ISSUE #6: Replace Upstream Test Code With Dependency
#4 enabled tests to be written like in JUnit 5. Part of the implementation copy-pasted a bunch of classes from the project (see f613211 for reasons). As soon as JUnit publishes test artifacts (see junit-team/junit5#572) a proper dependency can be established and the class can be removed.

Bukama:
```
Up to now projects tests depends on copied internal JUnit Jupiter classes. In the past this was necessary but we always wanted to get rid of them. With the upgrade to JUnit 5.4.2 this is now possible, as the JUnit Jupiter team provides an public `EngineKit` as an API to those classes.

So with this PR, Pioneer gets its own `TestKit` and `EngineExectionResults` to wrap those of the JUnit Jupiter test engine as our tests always execute the `pioneer test engine`. This wrapper also provides more flexibility adding new functionality and/or easier access to things like the published `ReportEntries` of the executed test method.

PR: #218
closed #6 

ci skip-release
```
---


src/test/java/org/junitpioneer/jupiter/ReportEntryExtensionTest.java
```diff
@@ -83,26 +83,18 @@ void repeatedAnnotation_logEachKeyValuePairAsIndividualEntry() {
...
 	}
 
 	static class ReportEntriesTest {
 
+		private static AtomicInteger executionCount = new AtomicInteger(0);
+
 		@Test
 		@ReportEntry(key = "Crow2", value = "While I pondered weak and weary")
 		void explicitKey() {
+			System.out.println("HI THERE! " + executionCount.getAndIncrement());
```
Michael1993, line 97: Why is this here?

src/test/java/org/junitpioneer/jupiter/ReportEntryExtensionTest.java
```diff
@@ -83,26 +83,18 @@ void repeatedAnnotation_logEachKeyValuePairAsIndividualEntry() {
...
-				.collect(toList());
-	}
-
 	private static Map.Entry<String, String> entryOf(String key, String value) {
 		return new AbstractMap.SimpleEntry<>(key, value);
 	}
 
 	static class ReportEntriesTest {
 
+		private static AtomicInteger executionCount = new AtomicInteger(0);
```
Michael1993, line 92: Don't really see why this is here. Just for testing?

src/test/java/org/junitpioneer/platform/testkit/engine/PioneerEngineExecutionResults.java
```diff
@@ -0,0 +1,156 @@
...
+
+package org.junitpioneer.platform.testkit.engine;
+
+import static java.util.stream.Collectors.toList;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+import org.junit.platform.engine.TestDescriptor;
```
Michael1993, line 19: Test descriptor is 'unused' in the strictest sense because it is only mentioned in comments. I think we should use fully-qualified names in docs.

src/test/java/org/junitpioneer/platform/testkit/engine/PioneerEngineExecutionResults.java
```diff
@@ -0,0 +1,156 @@
...
+import org.junit.platform.engine.TestExecutionResult;
+import org.junit.platform.engine.discovery.DiscoverySelectors;
+import org.junit.platform.testkit.engine.EngineExecutionResults;
+import org.junit.platform.testkit.engine.EngineTestKit;
+import org.junit.platform.testkit.engine.Events;
+
+/**
+ * Pioneer own class to handle {@link EngineExecutionResults} of the JUnit-Jupiter-Engine.
```
Michael1993, line 27: 'Pioneers' to show possession. Sounds better to me.

src/test/java/org/junitpioneer/platform/testkit/engine/PioneerEngineExecutionResults.java
```diff
@@ -0,0 +1,156 @@
...
+	 * Returns the published report entries of all tests.
+	 *
+	 * @return published report entries of all tests
+	 */
+	public List<Map<String, String>> getPublishedTestReportEntries() {
+		return executionResults
+				.tests()
+				.reportingEntryPublished()
+				.stream()
+				.map(executionEvent -> executionEvent.getPayload(org.junit.platform.engine.reporting.ReportEntry.class))
```
Michael1993, line 149: You could import this (in which case the import is used over the `ReportEntry` in the same package).

src/test/java/org/junitpioneer/platform/testkit/engine/PioneerEngineExecutionResults.java
```diff
@@ -0,0 +1,156 @@
...
+				.filter(Optional::isPresent)
+				.map(Optional::get)
+				.map(org.junit.platform.engine.reporting.ReportEntry::getKeyValuePairs)
```
Michael1993, line 152: You could import this.

src/test/java/org/junitpioneer/platform/testkit/engine/PioneerTestKit.java
```diff
@@ -0,0 +1,40 @@
...
+	 * Returns the execution results of the given method of a given test class.
+	 *
+	 * @param testClass Name of the test class
+	 * @param testMethodName Name of the test method (of the given class)
+	 * @return The execution results
+	 */
+	public static PioneerEngineExecutionResults getExecutionResults(Class<?> testClass, String testMethodName) {
+		PioneerEngineExecutionResults results = new PioneerEngineExecutionResults(testClass, testMethodName);
+
+		return results;
```
Michael1993, line 37: You could inline this (i.e.: `return new Pioneer...`).

src/test/java/org/junitpioneer/platform/testkit/engine/PioneerTestKit.java
```diff
@@ -0,0 +1,40 @@
...
+	/**
+	 * Returns the execution results of the given test class.
+	 *
+	 * @param testClass Name of the test class, the results should be returned
+	 * @return The execution results
+	 */
+	public static PioneerEngineExecutionResults getExecutionResults(Class<?> testClass) {
+		PioneerEngineExecutionResults results = new PioneerEngineExecutionResults(testClass);
+
+		return results;
```
Michael1993, line 24: You could inline this (i.e.: `return new Pioneer...`).

src/test/java/org/junitpioneer/jupiter/ReportEntryExtensionTest.java
Bukama: Nicolai added it for basic tests after creating the basic wrapper after updating so see if it works. And I just didn't think about it anymore. Will remove it

src/test/java/org/junitpioneer/platform/testkit/engine/PioneerTestKit.java
Bukama: I know, but I normally don't do it, to have a better debug possibility and the costs for this variable are quite less in my opinion.

Also in this specific case (and the one below) the PioneerTestKit is a wrapper for future improving. We will then have to split it again.

Bukama:
Those rebased and forced pushes on master really screw me up. Regardless if I rebase a branch on updated master or merge master into it. All changed files from master are shown in the PR....

src/test/java/org/junitpioneer/platform/testkit/engine/PioneerTestKit.java
Bukama: See above

src/test/java/org/junitpioneer/platform/testkit/engine/PioneerEngineExecutionResults.java
```diff
@@ -0,0 +1,155 @@
...
+import org.junit.platform.testkit.engine.EngineTestKit;
+import org.junit.platform.testkit.engine.Events;
+
+/**
+ * Pioneers own class to handle {@link org.junit.platform.testkit.engine.EngineExecutionResults} of the JUnit-Jupiter-Engine.
+ *
+ */
+public class PioneerEngineExecutionResults {
+
+	EngineExecutionResults executionResults;
```
Michael1993, line 31: Is this package-private on purpose?

src/test/java/org/junitpioneer/platform/testkit/engine/PioneerEngineExecutionResults.java
```diff
@@ -0,0 +1,155 @@
...
+
+	public PioneerEngineExecutionResults(Class<?> testClass) {
+		executionResults = EngineTestKit
+				.engine("junit-jupiter")
+				.selectors(DiscoverySelectors.selectClass(testClass))
```
Michael1993, line 36: Consider static importing `DiscoverySelectors` methods. Would that be more readable?

src/test/java/org/junitpioneer/platform/testkit/engine/PioneerEngineExecutionResults.java
```diff
@@ -0,0 +1,155 @@
...
+				.execute();
+	}
+
+	public PioneerEngineExecutionResults(Class<?> testClass, String testMethodName) {
+		executionResults = EngineTestKit
+				.engine("junit-jupiter")
+				.selectors(DiscoverySelectors.selectMethod(testClass, testMethodName))
```
Michael1993, line 43: Consider a static import.

src/test/java/org/junitpioneer/platform/testkit/engine/PioneerEngineExecutionResults.java
Bukama: Yes I always use package private instead of private to acces fields and methods from tests

src/test/java/org/junitpioneer/platform/testkit/engine/PioneerEngineExecutionResults.java
Bukama: Personally I like regular imports in such cases more as it's clear that this messages does not belong to the class. But if the majority wants to use static imports I can switch.

src/test/java/org/junitpioneer/platform/testkit/engine/PioneerEngineExecutionResults.java
Michael1993: But in this case - do you access it? I'm not seeing the tests for this class.

src/test/java/org/junitpioneer/platform/testkit/engine/PioneerEngineExecutionResults.java
nicolaiparlog: I like static imports in general, but I prefer them in situations where a somewhat universal API is used that is frequently seen. `DiscoverySelectors::selectClass` doesn't quite qualify, but I'm still ok, either way.

Bukama: I would like to keep them. So I resolve this conversation.

Bukama:
Moved all extension tests to the new API, but now two tests of the TempDirExtension fail because it seems that they are not started anymore and I can't figure out why :(

Bukama:
> Moved all extension tests to the new API, but now two tests of the TempDirExtension fail because it seems that they are not started anymore and I can't figure out why :(

I mean: *NOW* I moved *ALL* extension tests to the new API and removed the superfluous classes.

But still two failing tests and I don't know f*** why 😢 If they would run, this would be mergeable.

Bukama:
> > Moved all extension tests to the new API, but now two tests of the TempDirExtension fail because it seems that they are not started anymore and I can't figure out why :(
> 
> I mean: _NOW_ I moved _ALL_ extension tests to the new API and removed the superfluous classes.
> 
> But still two failing tests and I don't know f*** why 😢 If they would run, this would be mergeable.

NAILED them 🥳

nicolaiparlog:
Great work, @Bukama! This was tough, but we made it. :tada: Let's hope we get a lot of green lights, so we can merge this.

Issue #217, opened 2020-04-03
Evalute 'code smell' usefulness


Michael1993:
As can be seen in #216 SonarCloud code smells can often result in "false positives". Should this particular feature be disabled?

Most common false positives are the empty tests we don't actually execute just record with `TestExecutionRecorder`.

Bukama:
At this particular: Definitly as it doesn't fit for our environment.

In General? I don't think so as we then can skip many things. When @aepfli confirms that he has finished setting up sonar then we should evaluate all active rules (not only code smells) and also fix violations of rules we want to keep

aepfli:
Sonar is setup ;)

Bukama:
> Sonar is setup ;)

Okay, I'll try to go through it. Maybe this evening then the others play Stellaris

Bukama:
> > Sonar is setup ;)
> 
> Okay, I'll try to go through it. Maybe this evening then the others play Stellaris

As this said, I'll assign the issue to me as long as I have a look at the checks.

Bukama:
From `src/main`
1. Generic `Exceptions` or `Thrownables` are used, this is correct as you should aways use a particular exception depending on your context, e.g. `IllegalArguementException` for invalid parameters
2. Several local variables don't follow the naming conventions (`^[a-z][a-zA-Z0-9]*$`, e.g. `__` - yeah such names are used! So we have to decide if want these naming convention. As they are the default convetions of Java I would vote to do so
3. `peek()` is used in streams, but offical javadoc says that its for debugging purpose and not to be used in production code
4. `Remove this use of "Test"; it is deprecated` - this is mentioned several times in the vintage extension, as the `@Test` annotation there is deprecated. The rule itself (don't use deprecated classes/methods) is very important in my opinion so we would need to exclude it for this class or change the implementation. This also belongs to the annotation itself (`Do not forget to remove this deprecated code someday`)


From `src/test`
5. `Add at least one assertion to this test case` we got this messages en mass. This is because we mostyl deal with annotations and not with tests itself. I vote for disable this check for pioneer.
6. `Remove this use of "Thread.sleep()"` This is used inside a test for the `TimeoutExtension`. I don't see a way to remove this warning.

I vote for cleaning up all the things from `src/main`. I already cleaned up some things (#221), where I didn't see the need of an discussion. I would do for the points 1 and 2. For 3 I need help as I'm really not into streams (this goes for another violation I don't get fixed, see the draf)

[...]

Issue #215, opened 2020-04-02
Clean up / align internal structure, names, etc.


Bukama:
The current milestone we work on is called ["Cleaner Pioneer"](https://github.com/junit-pioneer/junit-pioneer/milestone/6) and includes issues covering several different like infrastructure, refacotring, polish and more. Goal is to build a solid base for future work (goal is `version 1.0`). But in my opinion one aspect of the term "Cleaner Pioneer" is missing: The current state of the lib itselfs is a mess in terms of folders, naming. This issue will show some points which in my opinion should be cleaned up / aligned. Maybe there are even more.

## Definition  of annotations

**Current state**

Currently most annotations are declared in own interface-files, but not all (e.g. `@TempDirectory` is definied inside the `TempDirectory` class).
Repeatable annotations (here I mean the "plural" ones that hold the array, e.g. `@ReportEntries`) are also declared in own interface-files, regardless the fact, that they are never used as an annotation.

**Proposal**:

We should define a standard so all annotations the same way. Here we should go for defining the declaration inside own interface-files (e.g. `@ReportEntry`).

A special case may be repeatable annotations: I accidently stumbled over [this article](https://dzone.com/articles/repeatable-annotations-in-java-8-1) about declaration of repeatable annotations. After reading it I think this is a nice way to provide a more intuitive and cleaner API. It also makes it easier to compare that both annotation have the same `@Target` and `@Retention`. So I suggest to following the guidance in the articel and put the array holding annotations of repeatable annotations inside the corresponding one.  

## Location of annotations and other files

**Current state**

Currently annotations, their extensions and all other files (e.g. `ArgumentProvider`s, `InvocationContexts` or `Utils`) are declared inside the `org.junitpioneer.jupiter` package. A special case are the files or the parameter extensions, which were moved into a subpackage for better overview.

The JUnit API is declared in `org.junit.jupiter.api` and subpackages of this (e.g. `org.junit.jupiter.api.extension`). *Note: It's also declared in a own java module but for pioneer we have already decided (see #25) that we don't want to build a multi module project (at least not yet).*

**Proposal**:

As annotations are the API of pioneer and to get somehow aligned to JUnit I suggest to move:

* all annotations to `org.junitpioneer.jupiter.api` or sub packages of it, e.g. `params` for extensions / features with a lot of annotations.
* all extension implementations and their need classes (e.g. `ArgumentProvider`s, `InvocationContexts`) to `org.junitpioneer.jupiter.extension`
[...]

PR #221, opened 2020-04-06
Fix sonarqube violations

Bukama:
This is a draft an connected to and blocked by #215 and #217 .

---


src/main/java/org/junitpioneer/jupiter/params/RangeSourceProvider.java
```diff
@@ -55,6 +55,7 @@
 								.stream(annotations.annotationType().getAnnotationsByType(ArgumentsSource.class))
 								.anyMatch(annotation -> getClass().equals(annotation.value())))
 						.collect(Collectors.toList()))
+				// .filter(Optional::isPresent) <-- Can not accessed in a static context
```
Bukama, line 58: Can anyone help me how to implement this `isPresent` check (violation is, that the following `get()` is used without checking if the value is present)

aepfli: haha flatmapping optionals in later java versions is fixing that actually. not sure how to best tackle this

Michael1993: The simplest solution would be to just say `.orElse(Collections.emptyList())` and let it throw a `PreconditionViolationException`?

Michael1993: I mean, the `map` should have an empty list anyways(?), so (while somewhat redundant) this is basically the same?

Bukama: > .orElse(Collections.emptyList())

You mean to replace the `.get()` with `orElse(Collections.emptyList())` ?

Bukama:
I would like to merge this (if everybody accepts) so we get a more cleaner view on master branch. For further violations I would then spend time again. But at the moment I always get confused which violations I have fixed and which not 🤷‍♂️

src/main/java/org/junitpioneer/jupiter/TempDirectory.java
```diff
@@ -125,7 +125,8 @@
 		 *
 		 * @return the parent directory for all temporary directories
 		 */
-		Path get(ParameterContext parameterContext, ExtensionContext extensionContext) throws Exception;
+		Path get(ParameterContext parameterContext, ExtensionContext extensionContext) throws Exception; // NOSNAR
```
aepfli, line 128:
```suggestion
		Path get(ParameterContext parameterContext, ExtensionContext extensionContext) throws Exception; // NOSONAR
```

Bukama: Ops yes. And I think my other `// NOSONAR` where wrong too, but not sure. Anyway changed them to `//NOSONAR`

src/main/java/org/junitpioneer/jupiter/TempDirectory.java
```diff
@@ -345,6 +346,7 @@ private void tryToDeleteOnExit(Path path) {
 				path.toFile().deleteOnExit();
 			}
 			catch (UnsupportedOperationException ignore) {
+				// this exception shall be ignored by purpose
```
Michael1993, line 349: What purpose? Shouldn't that be included in the comment?

Bukama: > What purpose? Shouldn't that be included in the comment?

To be honest I'm not sure what the exactly purpose is, but as the method is only called in on method which creates an own exception, I worte that down

src/main/java/org/junitpioneer/jupiter/EnvironmentVariableExtension.java
```diff
@@ -92,8 +92,9 @@ private void storeOriginalEnvironmentVariables(ExtensionContext context, Collect
 
 	private void reportWarning(ExtensionContext context) {
 		boolean wasReported = REPORTED_WARNING.getAndSet(true);
-		if (!wasReported)
+		if (!wasReported) {
 			context.publishReportEntry(WARNING_KEY, WARNING_VALUE);
+		}
```
nicolaiparlog, line 97: I don't like having these superfluous parenthesis in our code. They add little except almost empty lines.

src/main/java/org/junitpioneer/jupiter/RepeatFailedTestExtension.java
```diff
@@ -59,8 +60,8 @@ private static FailedTestRepeater repeaterFor(ExtensionContext context) {
 		Method repeatedTest = context.getRequiredTestMethod();
 		return context
 				.getStore(NAMESPACE)
-				.getOrComputeIfAbsent(repeatedTest.toString(), __ -> FailedTestRepeater.createFor(repeatedTest),
-					FailedTestRepeater.class);
+				.getOrComputeIfAbsent(repeatedTest.toString(),
+					defaultCreator -> FailedTestRepeater.createFor(repeatedTest), FailedTestRepeater.class);
```
nicolaiparlog, line 64: `__` means "nobody cares about this variable" - I want to keep using that.

src/main/java/org/junitpioneer/jupiter/RepeatFailedTestExtension.java
```diff
@@ -72,8 +73,8 @@ private static FailedTestRepeater repeaterFor(ExtensionContext context) {
 
 		private FailedTestRepeater(int maxRepetitions) {
 			this.maxRepetitions = maxRepetitions;
-			this.repetitionsSoFar = 0;
-			this.exceptionsSoFar = 0;
+			repetitionsSoFar = 0;
+			exceptionsSoFar = 0;
```
nicolaiparlog, line 77: The `this.` are unnecessary, but IMO they improve readability by keeping the three assignments aligned for easy top-down reading.

src/main/java/org/junitpioneer/jupiter/params/RangeSourceProvider.java
```diff
@@ -55,7 +56,7 @@
 								.stream(annotations.annotationType().getAnnotationsByType(ArgumentsSource.class))
 								.anyMatch(annotation -> getClass().equals(annotation.value())))
 						.collect(Collectors.toList()))
-				.get();
+				.orElse(Collections.emptyList());
```
nicolaiparlog, line 59: It's good to avoid to just `get()`, but there's a comment a few lines above:

> // since it's a method annotation, the element will always be present

This justifies why the `Optional` can never be empty. I think if it is, some central assumption of the extension no longer holds and this _should_ explode. Probably with an `orElseThrow(IllegalStateException::new)`, though.

src/main/java/org/junitpioneer/jupiter/EnvironmentVariableExtension.java
aepfli: but i do not think they are sonar related - sonar is not complaining about those in the gui

Bukama: Maybe I just saw this while doing something else in the code and edited it then. For me (personally) they improve readability of the code. And I like them because I had to debug an massive amount of bugs where at the end it comes down that someone thought "na I don't need parantheses.." and broke the code.

But as @aepfli mentions there is no rule which enforces the one or the other, we can go with both as the author wants to use it.

src/main/java/org/junitpioneer/jupiter/RepeatFailedTestExtension.java
Bukama: Never saw this before and was wondering that it's even allowed. Had to google it up to know that it's a way to show that the variable is not needed.

src/main/java/org/junitpioneer/jupiter/RepeatFailedTestExtension.java
Bukama: I can follow your argumentation and suggest that @aepfli disables the rule

src/main/java/org/junitpioneer/jupiter/params/RangeSourceProvider.java
Bukama: I like the `orElseThrow(IllegalStateException::new)` thing! 👍

Bukama:
Will revert / change the two things @nicolaiparlog is unhappy with. For the other two @aepfli has to change the rules.

src/main/java/org/junitpioneer/jupiter/RepeatFailedTestExtension.java
aepfli: hmm, is there really a rule, i cant find a violation in sonarqube

src/main/java/org/junitpioneer/jupiter/RepeatFailedTestExtension.java
aepfli: adapted the regex to `^([a-z][a-zA-Z0-9]*|__)$`

src/main/java/org/junitpioneer/jupiter/EnvironmentVariableExtension.java
Bukama: I just checked it. It was added by my `save action` plugin. I have only a very few of its action enabled, but as written that I really like having paranthesis for blocks.

But as wished I disabled it to revert this one. So now at this if the paranthesis are gone again, but in a similar block a few lines below they are still there so we have an inconsistency there now (which I never had touched, so they are from the original author).

So you see were this ends? We should either define if we want to force using paranthesis or not for blocks and force them by a rule or we don't force the one or the other and let the authors use their personal like.

src/main/java/org/junitpioneer/jupiter/RepeatFailedTestExtension.java
Bukama: > adapted the regex to `^([a-z][a-zA-Z0-9]*|__)$`

As written only double underscore is not enough.

```
	public TempDirectory() {
		this((__, ___, dirPrefix) -> createDefaultTempDir(dirPrefix));
	}
``

File: src/main/java/org/junitpioneer/jupiter/RepeatFailedTestExtension.java (hunk 6 above)
Comment (Bukama) [reply src/main/java/org/junitpioneer/jupiter/RepeatFailedTestExtension.java:77]: Again save actions

File: src/main/java/org/junitpioneer/vintage/ExpectedExceptionExtension.java [hunk 8]
```diff
@@ -44,14 +44,14 @@ public void handleTestExecutionException(ExtensionContext context, Throwable thr
 				.filter(expected -> expected.isInstance(throwable))
 				.isPresent();
 
-		// in the `afterTestExecution` callback we have to pass or fail the test
-		// depending on whether the exception was thrown or not;
-		// to do that we need to register whether the exception was thrown;
```
Bukama, line 49: @nicolaiparlog You were wondering why this was changes and you thought it was because the lower case at the start of the line. No it's not. The `;` are the violation as sonarcube thinks it's outcommented code, see [here](https://sonarcloud.io/project/issues?id=junit-pioneer_junit-pioneer&open=AXDuy1qtYLFyXCsBD7Mm&resolved=false&severities=MAJOR#)

Bukama:
> 50.0% Coverage on New Code (is less than 80%)

thank you sonarqube....

src/main/java/org/junitpioneer/jupiter/RepeatFailedTestExtension.java
aepfli: meh than i will add more than 2 _

Bukama:
I'll check the failing test coverage. Does anyone has an idea why sonarcube plugin was not found on macos 14 ?

```
A problem occurred configuring root project 'junit-pioneer'.
> Could not resolve all artifacts for configuration ':classpath'.
   > Could not resolve org.sonarsource.scanner.api:sonar-scanner-api:2.14.0.2002.
     Required by:
         project : > org.sonarqube:org.sonarqube.gradle.plugin:2.8 > org.sonarsource.scanner.gradle:sonarqube-gradle-plugin:2.8
      > Could not resolve org.sonarsource.scanner.api:sonar-scanner-api:2.14.0.2002.
         > Could not parse POM https://plugins.gradle.org/m2/org/sonarsource/scanner/api/sonar-scanner-api/2.14.0.2002/sonar-scanner-api-2.14.0.2002.pom
            > Could not resolve org.sonarsource.scanner.api:sonar-scanner-api-parent:2.14.0.2002.
               > Could not resolve org.sonarsource.scanner.api:sonar-scanner-api-parent:2.14.0.2002.
                  > Could not parse POM https://plugins.gradle.org/m2/org/sonarsource/scanner/api/sonar-scanner-api-parent/2.14.0.2002/sonar-scanner-api-parent-2.14.0.2002.pom
                     > Could not resolve org.sonarsource.parent:parent:52.
                        > Could not resolve org.sonarsource.parent:parent:52.
                           > Could not get resource 'https://plugins.gradle.org/m2/org/sonarsource/parent/parent/52/parent-52.pom'.
                              > Could not HEAD 'https://plugins.gradle.org/m2/org/sonarsource/parent/parent/52/parent-52.pom'.
                                 > Operation timed out
```

sormuras:
https://plugins.gradle.org is sometimes not responding. We see that error from time to time as well.

src/main/java/org/junitpioneer/jupiter/TempDirectory.java
```diff
@@ -345,6 +346,9 @@ private void tryToDeleteOnExit(Path path) {
 				path.toFile().deleteOnExit();
 			}
 			catch (UnsupportedOperationException ignore) {
+				// this exception shall be ignored by purpose this method is only called in the
+				// stream of the method "createIOExceptionWithAttachedFailures".
+				// That method overrules this "UnsupportedOperationException" anyway.
```
nicolaiparlog, line 351:
```suggestion
				// If the `Path` can't be turned into a `File` (which throws the UOE),
				// it can't be registered to be deleted when the JVM terminates.
				// Because deleting on JVM termination is just a last ditch effort and
				// nicety towards the user, it is entirely optional and shouldn't affect
				// the extension's behavior.
```

src/main/java/org/junitpioneer/vintage/ExpectedExceptionExtension.java
```diff
@@ -44,14 +44,14 @@ public void handleTestExecutionException(ExtensionContext context, Throwable thr
...
-		// (NOTE that if no exception was thrown, NOTHING is registered)
+		// In the `afterTestExecution` callback we have to pass or fail the test
+		// depending on whether the exception was thrown or not.
+		// To do that we need to register whether the exception was thrown
+		// (NOTE that if no exception was thrown, NOTHING is registered).
 		if (throwableMatchesExpectedException) {
 			storeExceptionStatus(context, EXCEPTION.WAS_THROWN_AS_EXPECTED);
 		} else {
-			// this extension is not in charge of the throwable, so we need to rethrow;
+			// This extension is not in charge of the throwable, so we need to rethrow
```
nicolaiparlog, line 54:
```suggestion
			// this extension is not in charge of the throwable, so we need to rethrow
```

src/main/java/org/junitpioneer/vintage/ExpectedExceptionExtension.java
```diff
@@ -67,13 +67,14 @@ public void afterTestExecution(ExtensionContext context) throws Exception {
 							throw error;
 						});
 			case WAS_THROWN_AS_EXPECTED:
-				// the exception was thrown as expected so there is nothing to do
+				// The exception was thrown as expected so there is nothing to do
```
nicolaiparlog, line 70:
```suggestion
				// the exception was thrown as expected so there is nothing to do
```

Bukama:
I'm unable to pass the quality gate as the not covered part in the `ExpectedExceptionExtension` is not accessible so it will never get called as long nobody breaks the extension using reflection.

Same for the not covered code in `RepeatFailedTestExtension` as the inner class is made private static which makes is unable to access it with package private tests. @nicolaiparlog mentioned some weeks ago that he dislikes package private fields/methods for testing purpose, so I won't "open" the classes just for Sonar.

aepfli:
@Bukama  imho i think your PR is still fine to merge, although you do not pass the quality gate, it complains about 2 lines which are not tested, and in that kind of case, you will invest more time into writing test for that cases, as it is to just accept the fact. Sonar is a nice tool, but in the end, we as developers decide what is good or bad, and it that case everything is totally fine. so i say we can merge here for sure.

PR #223, opened 2020-04-07
Fixed a typo, minor cleanup of utility method

Michael1993:
---

### DIFF
--- a/src/test/java/org/junitpioneer/jupiter/TestUtils.java
+++ b/src/test/java/org/junitpioneer/jupiter/TestUtils.java
@@ -17,6 +17,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
+import org.junit.platform.engine.reporting.ReportEntry;
 import org.junit.platform.engine.test.event.ExecutionEvent;
 import org.junit.platform.engine.test.event.ExecutionEventRecorder;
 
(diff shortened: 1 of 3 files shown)


src/test/java/org/junitpioneer/jupiter/TestUtils.java
```diff
@@ -25,11 +26,10 @@
 	static List<Map<String, String>> reportEntries(ExecutionEventRecorder recorder) {
 		return recorder
 				.eventStream()
-				.filter(event -> event.getType().equals(ExecutionEvent.Type.REPORTING_ENTRY_PUBLISHED))
```
Michael1993, line 28: This line was deleted because it's not needed - the get payload ensures that we get ReportEntry or empty and we filter out empties. Another option is to remove the Optional::isPresent but that's a 'code smell', I think.

Bukama: But do we still get the correct ReportEntries (wondering about the `REPORTING_ENTRY_PUBLISHED`)?

Michael1993: Yes, because the only events in the recorder that have a ReportEntry.class type payload are the report entry events.

Michael1993: There is only the `REPORTING_ENTRY_PUBLISHED` type in `ExecutionEvent.Type` that's related to report entries, so the filtering does nothing, I believe.

Issue #233, opened 2020-04-22
RepeatFailedTest causes sporadically PreconditionViolationException


mp911de:
A failed test repeated through `@RepeatFailedTest` sporadically leads to:

```
org.junit.platform.commons.PreconditionViolationException: Illegal state: required test method is not present in the current ExtensionContext
```

Link to a failed build: https://travis-ci.org/github/spring-projects/spring-data-mongodb/jobs/678085125

Reproducer: https://github.com/spring-projects/spring-data-mongodb/blob/46ab6b4c945fa5eef1acc000c9d9735a715c8f31/spring-data-mongodb/src/test/java/org/springframework/data/mongodb/core/messaging/ChangeStreamTests.java#L201

nicolaiparlog:
Thanks for opening the issue, Mark. I'll look into it tomorrow.

nicolaiparlog:
Immediate solution is simple: Remove [superfluous `@Test` annotation](https://github.com/spring-projects/spring-data-mongodb/blob/46ab6b4c945fa5eef1acc000c9d9735a715c8f31/spring-data-mongodb/src/test/java/org/springframework/data/mongodb/core/messaging/ChangeStreamTests.java#L200).

I'll post a longer explanation as soon as I'm done writing it. :grin:

nicolaiparlog:
# What's going on?

The `@RepeatFailedTest` extension is mainly a [test template](https://junit.org/junit5/docs/current/user-guide/#writing-tests-test-templates), which means it becomes it's own container for the tests it generates and doesn't need an additional `@Test` annotation (just like `@ParameterizedTest`).

In the linked example, the test method is annotated with `@Test` _and_ `@RepeatFailedTest`, though, just like this example:

```java
@Test
@RepeatFailedTest(3)
void failsNever() { }
```

This means if the test is successful, it's executed twice:

![image](https://user-images.githubusercontent.com/6537432/80126340-40149980-8593-11ea-9cc8-f936993d44c1.png)

If it isn't successful, things get a bit more complicated. I said, this is extension is _mainly_ a test template. It's also a `TestExecutionExceptionHandler`, so it can catch exceptions and trigger an additional run:

```java
@Override
public void handleTestExecutionException(ExtensionContext context, Throwable throwable) throws Throwable {
	// this `context` (M) is a child of the context passed to `provideTestTemplateInvocationContexts` (T),
	// which means M's store content is invisible to T's store; this can be fixed by using T's store here
[...]

PR #236, opened 2020-04-23
Rename/clarify RepeatFailedTest (#233)

Bukama:
This draft emerges from #233

This draft improves the documentation and adds more tests for the `@RepeatFailedTest` to point out that `@Test` and `@TestTemplate` should not be combined.

I was trying to add a new method in `PioneerAnnotationUtils` like `isOnlyOneTestAnnotationPresent` which was not a problem. But I was unsure how to throw an exception as the method which checks if the test template must be applied or not, because this would break all other tests which should not executed by the extension.

I threw it away because I noticed that there is already a warning, when such misconfiguration is present (which seems to be printed for every annotation):

> Apr. 23, 2020 10:59:36 NACHM. org.junit.jupiter.engine.discovery.JavaElementsResolver logMultipleTestDescriptorsForSingleElement
WARNUNG: Possible configuration error: method [void org.junitpioneer.jupiter.RepeatFailedTestTests$RepeatFailedTestTestCase.invalidConfigurationWithTest()] resulted in multiple TestDescriptors [org.junit.jupiter.engine.descriptor.TestTemplateTestDescriptor, org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor]. This is typically the result of annotating a method with multiple competing annotations such as @Test, @RepeatedTest, @ParameterizedTest, @TestFactory, etc.
> Apr. 23, 2020 10:59:36 NACHM. org.junit.jupiter.engine.discovery.JavaElementsResolver logMultipleTestDescriptorsForSingleElement
WARNUNG: Possible configuration error: method [void org.junitpioneer.jupiter.RepeatFailedTestTests$RepeatFailedTestTestCase.invalidConfigurationWithTest()] resulted in multiple TestDescriptors [org.junit.jupiter.engine.descriptor.TestTemplateTestDescriptor, org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor]. This is typically the result of annotating a method with multiple competing annotations such as @Test, @RepeatedTest, @ParameterizedTest, @TestFactory, etc.

---


nicolaiparlog:
Proposed commit message:

```
Rename `RepeatFailedTest` to `RetryingTest`

#233 reported a problem that was eventually tracked down to
`@RepeatFailedTest` and `@Test` being both applied to the same
method. While the extension's documentation clearly advises against
that, we think the name contributed to the confusion.
`@RepeatFailedTest` may sound too much like needing to be applied to
a method that is already a test, i.e. already annotated with `@Test`.
We hope that `@RetryingTest` prevents that. 
```

nicolaiparlog:
:warning: We need to check whether the docs were updated as well.

Bukama:
Docs updated, build successful locally

aepfli:
we do have to be careful with renaming and deleting annotations. I do not think that it is a nice move, to just rename it. I think we do have a responsibility as a library maintainer to allow an easy migration. Hence that i am more a fan of deprecation, and dedicated information, when this annotation is gonna get deleted. so our users have time to react to this, instead of being forced to do this immediately. What is your take on that?

Bukama:
> we do have to be careful with renaming and deleting annotations. I do not think that it is a nice move, to just rename it. I think we do have a responsibility as a library maintainer to allow an easy migration. Hence that i am more a fan of deprecation, and dedicated information, when this annotation is gonna get deleted. so our users have time to react to this, instead of being forced to do this immediately. What is your take on that?

I see you point and in general I'm totally with you. But as already mentioned somewhere else (don't find it, maybe it was on Discord or Stream) we have not released 1.0 yet, so I think it's okay to rename the feature (as done my @nicolaiparlog )

nicolaiparlog:
I agree @aepfli, but that's why we're doing it now before releasing 1.0. After that, we definitely have to give users time to migrate.

PR #240, opened 2020-04-24
Fix enviornmentvariablesextension documentation

### LINKED ISSUE #237: EnvironmentVariable annotations/extension documentation problems
@nicolaiparlog  I noticed during watching a past recording of your stream that during the work on #174 the warnings went into the wrong file.

https://github.com/junit-pioneer/junit-pioneer/blob/master/src/main/java/org/junitpioneer/jupiter/ClearSystemProperty.java#L33

Should be in https://github.com/junit-pioneer/junit-pioneer/blob/master/src/main/java/org/junitpioneer/jupiter/ClearEnvironmentVariable.java

Same for 
https://github.com/junit-pioneer/junit-pioneer/blob/master/src/main/java/org/junitpioneer/jupiter/SetSystemProperty.java#L34

Should be in https://github.com/junit-pioneer/junit-pioneer/blob/master/src/main/java/org/junitpioneer/jupiter/SetEnvironmentVariable.java

Check also the `@since` from SystemProperty Extensions

Also https://github.com/junit-pioneer/junit-pioneer/blob/master/build.gradle.kts#L106 should reference the EnvironmentVariableExtension instead of the SystemPropertyExtension

https://github.com/junit-pioneer/junit-pioneer/blob/master/docs/enviroment-variables.adoc
`class MySystemPropertyTest {` should proably also be changed to `class MyEnvironmentVariableTest {`

Hancho2009:
fixes #237
---

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -104,7 +104,7 @@ tasks {
             includeTestsMatching("*Tests")
         }
         systemProperty("java.util.logging.manager", "org.apache.logging.log4j.jul.LogManager")
-        // `SystemPropertyExtension` uses reflection to change environment variables;
+        // `EnvironmentVariableExtension` uses reflection to change environment variables;
         // this prevents the corresponding warning (and keeps working on Java 8)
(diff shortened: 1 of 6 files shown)


nicolaiparlog:
Thanks for spotting and resolving this! :+1:

PR #242, opened 2020-04-25
Environmentvariableextension securitymanager tests

### LINKED ISSUE #241: No EnvironmentVariableUtils Tests if a SecurityManager is set
There are not tests for `org.junitpioneer.jupiter.EnvironmentVariableUtils` when a SecurityManager is set.

Hancho2009:
resolves #241
---


Bukama:
For documentation:

> Task :test FAILED
2020-04-25 14:51:44,487 pool-1-thread-1 ERROR Unable to unregister MBeans java.security.AccessControlException: access denied ("javax.management.MBeanServerPermission" "createMBeanServer")

src/test/java/org/junitpioneer/jupiter/EnvironmentVariableUtilsTests.java
```diff
@@ -0,0 +1,59 @@
...
+			assertThat(System.getenv("TEST")).isEqualTo("TEST");
+		});
+	}
+
+	/*
+	 * This needs to be done during the execution of the test method and cannot be moved to setup/tear down
+	 * because junit uses reflection and the default SecurityManager prevents that.
+	 */
+	private void executeWithSecurityManager(Runnable runnable) {
+		Policy.getPolicy().refresh();
```
Hancho2009, line 48: Does something like that needs to be called after the SystemProperty java.security.policy is reverted after the test?

Hancho2009:
I am very inexperienced with the SecurityManager. If someone has better ideas how to handle this please share it with me.
I do not really like the *.policy files and how I can only give a certain code bases permission but what I really want is to remove certain (or all) permissions from the code base I am testing and not effect the others like gradle and junit.
I thought about implementing a SecurityManager for that but it seems a little to much for this small case. it would be cool if there where an junit extension to control policy permissions without files in the contex of a test i.e. the @SetSystemProperty annotation.

aepfli:
sidenote: i think this security thingy quiet interesting, and i am curious, beside this usecase with the policy - do you/we think that this might be a possible junit extension? we could generate a ticket out of that, and evaluate options, but it looks like something which could be useful. just an idea i came up while reading through this

Hancho2009:
> sidenote: i think this security thingy quiet interesting, and i am curious, beside this usecase with the policy - do you/we think that this might be a possible junit extension? we could generate a ticket out of that, and evaluate options, but it looks like something which could be useful. just an idea i came up while reading through this

In general I could imagine such an extension but as I stated above I do not have much experience with the SecurityManager so I hope someone with more of a grasp on the topic can give some input. I think it would be good to create a ticket/issues for that to have a place for discussions if this PR will go away.

aepfli:
> To be honest I don't know anything about this topic and I don't know if I find time to read into it. So I ask @aepfli and @nicolaiparlog for their review.

haha me neither - i will try to check today

src/test/java/org/junitpioneer/jupiter/EnvironmentVariableUtilsTests.java
```diff
@@ -0,0 +1,59 @@
...
+		SecurityManager securityManager = System.getSecurityManager();
```
Michael1993, line 49: I'd have named this variable `original`. Everywhere it is used it is clear that it's a `SecurityManager`, so naming it `securityManager` is redundant - however `System.setSecurityManager(original)` clearly communicates that we reset (and not modify) the `SecurityManager`.
This is just a sidenote, don't feel the need to modify the PR just for this.

Bukama: > I'd have named this variable `original`. Everywhere it is used it is clear that it's a `SecurityManager`, so naming it `securityManager` is redundant - however `System.setSecurityManager(original)` clearly communicates that we reset (and not modify) the `SecurityManager`.
> This is just a sidenote, don't feel the need to modify the PR just for this.

I would *not* rename the variable! It only seems redundant because of Java 8. With 10+, using var at this section it would not be clear what a variable with the name `original` holds.

Michael1993: I mean, even with `var` it would look like this:
```
var original = System.getSecurityManager();
// ... other code ...
System.setSecurityManager(original);
```
Seems clear to me?

src/test/java/org/junitpioneer/jupiter/EnvironmentVariableUtilsTests.java
Michael1993: According to the [official Oracle Java SE documentation](https://docs.oracle.com/javase/8/docs/api/index.html?java/security/Policy.html)
> The `refresh` method causes the policy object to refresh/reload its data. This operation is implementation-dependent. For example, if the policy object stores its data in configuration files, calling `refresh` will cause it to re-read the configuration policy files. If a refresh operation is not supported, this method does nothing. Note that refreshed policy may not have an effect on classes in a particular ProtectionDomain. This is dependent on the Policy provider's implementation of the `implies` method and its PermissionCollection caching strategy.

Michael1993: If you want to test if resetting the policy works as expected, you should probably add a new test for that and add `@Order` to your suite. [Here is a blog post from Baeldung on that topic]()https://www.baeldung.com/junit-5-test-order.

nicolaiparlog:
Finally took some time to look at this, @Hancho2009. Sorry for letting it sit for nigh three months. Looks good to me! :smiley: 

Proposed commit message:

```
Add tests with enabled SecurityManager to env extension (#241 / #242)

The documentation mentions that without proper permissions an enabled
security manager will not give access to the internals we need to
change environment variables. These tests confirm that.

Closes: #241
PR: #242
```

Hancho2009:
@nicolaiparlog no problem

PR #244, opened 2020-04-26
Fix sonarqube violations (#217)

Bukama:
This PR replaces the two `peek()` calls with `map` as suggested by @Michael1993 in #217. Sonar cries about using `peek()` as it should be used for debugging only. 

This lefts us with only one violation (`Mind to remove this deprecated class in the future` for the vintage `@Test` annotation which has to be excluded in sonar in my opinion.

So I call this the last PR for the issue.

closes: #217

---


Bukama:
> [20:34] nipafx: Re peek, I understand Sonar's comment that "it should be used with caution" and that's what's happening here - caution was applied. :wink: It is perfectly ok in these cases and replacing it with map doesn't really change anything.
> [20:35] nipafx: We either need to deactivate that rule in general or add a comment that tells Sonar to shut up about it here.

So i close this PR

PR #245, opened 2020-04-28
Create assertions for Pioneer (#232)

### LINKED ISSUE #232: Provide pioneer own AssertJ assertions
When checking `ExcecutionResults` inside a test class for `Throwable`s you have to go through the results and filter them. While doing this, already assertions are made that there are such results. Furthermore to avoid repeated codes some assertions were moved into a method inside the test class and were so hided when only looking at the acutal test case.

We also have some private `assertThat...` methods in some extension test classes which can be used in other tests so. Moreover there are no good `assertThat(results)` methods in the default _AssertJ_ API. So in stream the idea came up to provide some own _AssertJ_ assertion methods, to do something like the following.

```
//   TODO: assertThat(results).hasSingleTestFailedWith(ExtensionConfigurationException.class);
//   TODON'T: result.hasThrowable().isInstanceOf().hasMessage()?
     results.assertTestFailedWithExtensionConfigurationException();
```

Another one: 
```
PioneerAnnotationUtilsTests:

		private void assertFailedTestHasMessage(ExecutionResults results, String... messages) {
			assertThat(results.numberOfFailedTests()).isEqualTo(1);
			assertThat(results.firstFailuresThrowableMessage()).isEqualTo(String.join(",", messages));
		}
```

Examples can be seen at this [tutorial](https://joel-costigliola.github.io/assertj/assertj-core-custom-assertions.html#custom-assertion-class)

Michael1993:
Created assertions for Pioneers `ExecutionResults` class with a (hopefully) intuitive fluent API.

Closes #232 

---


nicolaiparlog:
I really like what the refactored tests look like, this was exactly what I was hoping for: very clean and in the domain-specific language.

Some ideas on the implementation:

* since you have several classes, let's create a new package `org.junitpioneer.testkit.assert` and put'em there as their own top-level classes
* I like the fluent API with several interfaces and think it would be more clear cut if _all_ methods that belong to the API came from an interface
* then `PioneerAssertions::assertThat` can return an interface - more strictly, none of the methods should ever return a concrete class

You're definitely on the right path, so keep going. :+1:

src/test/java/org/junitpioneer/testkit/PioneerAssertContainer.java
```diff
@@ -0,0 +1,134 @@
...
+
+			List<Map<String, String>> reportEntries = actual.reportEntries();
+			Assertions.assertThat(reportEntries).hasSize(1);
+			Map<String, String> reportEntry = reportEntries.get(0);
+			Assertions.assertThat(reportEntry).hasSize(1);
+
+			return this;
+		}
+
+		public FailureAssert hasSingleFailedTest() {
```
aepfli, line 46: i might suggest to have one basic method like `hasNumberOfFailedTests(int number)` and than just use that within this method. as i think this assert might be usable for other cases in the future too.

src/test/java/org/junitpioneer/testkit/PioneerAssertContainer.java
```diff
@@ -0,0 +1,134 @@
...
+					.failed()
+					.stream()
+					.findFirst()
+					.flatMap(first -> first.getPayload(TestExecutionResult.class))
+					.flatMap(TestExecutionResult::getThrowable);
+
+		}
+
+		@Override
+		public ExceptionAssert andHasException(Class<? extends Throwable> exceptionType) {
```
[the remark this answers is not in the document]
Michael1993: My reasoning behind `andHasException` is that you can only get there if you already called a method that returns with the `FailureAssert` interface.

So it always looks like this:
```
assertThat(results).hasXY(...).andHasException();
```
In a purely grammatical context the `and` makes sense(?).

src/test/java/org/junitpioneer/testkit/PioneerAssertContainer.java
Bukama: I would like to keep the single methods and offer additional methods where you can pass the number, because we have plenty of tests where we expect exactly one. So why force the writer to pass magic numbers and make it less readable in those cases? Don't get me wrong. We need the basic methods for dynamic parameters, but this PR shows that the single ones makes sense too.

src/test/java/org/junitpioneer/testkit/PioneerAssertContainer.java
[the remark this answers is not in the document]
Bukama: To put an example to the discussion:

`assertThat(results).hasSingleFailedTest().andHasException().withMessageContaining("ne...")`

My first thought about naming was: We always compare `results` which is plural, so I asked myself if the methods should align with plural too. But I don't think so, because the default _AssertJ_ lib claims only a single result, independent if its a single value or a list or whatever. So we should go for singular method names too.

What I really like at @Michael1993 names is that he always has something in mind when choosing the names. Was the same at the publish condition of the `@ReportEntry` extension. And I can totally understand his thoughts here to. On the other hand I can see @aepfli point to align with the default _AssertJ_ which uses `withException" and I could imagine that this helps others people to use the API. While I see the use of `and` from grammatical context, I don't think we need it as there is always the `.` of the fluent API which says exactly this.

src/test/java/org/junitpioneer/testkit/PioneerAssertContainer.java
aepfli: i never said to skip that method, i said we should extract its functionality into an generic one, which we use within this method

Michael1993:
I was under the impression that this is test code and does not get bundled in JUnit Pioneer - so it is for internal use only. I agree that adding Javadoc is needed but why add examples to something users can not access?

src/test/java/org/junitpioneer/testkit/assertion/ExceptionAssert.java
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * http://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junitpioneer.testkit.assertion;
+
+/**
+ * Used to assert exceptions thrown by failed tests/containers.
+ */
+public interface ExceptionAssert {
```
aepfli, line 16: could we actually use https://github.com/joel-costigliola/assertj-core/blob/master/src/main/java/org/assertj/core/api/AbstractThrowableAssert.java instead of that assert - it does have already a lot of functionality and we would stick to the default

Michael1993: We *could* use it but it would look kind of broken:
```
assertThat(results)
    .hasSingleTest()
    .thatFailed()
    .withException()
    .hasMessageContainingAll(/* ... */);
```
Then again, this interface does not really do anything besides adding the word `that` in front of the methods in `AbstractThrowableAssert`

aepfli: but it would also provide additional methods, like message containing with varargs, etc.

Bukama:
> I was under the impression that this is test code and does not get bundled in JUnit Pioneer - so it is for internal use only. I agree that adding Javadoc is needed but why add examples to something users can not access?

For contributors. Or as @nicolaiparlog would say: `You. Me. And JUnit Pioneer". I like when tools are documented, so everybody knows about them and knows how to use them.

src/test/java/org/junitpioneer/testkit/assertion/ExceptionAssert.java
Michael1993: No need to reinvent the wheel, you are right.

Bukama:
I think we can turn this into a real PR, can we @Michael1993 ?

Bukama:
Thank you for turning it into a PR. Even that @aepfli and I had already approved I leave the PR open and assign it to @nicolaiparlog , because I'm nearly 100% sure that the wants to merge it on stream. We already had a look on this PR on stream and it's quite a real "core feature" of Pioneer, so perfect for stream.

So please don't be sad that you most probably won't get any update until then (I expect the next Pioneer stream to be in early June)

nicolaiparlog:
This is really cool and I love how far you've taken the simplistic "let's have AssertJ-esque assertions"-idea. I'm really looking forward to start using these assertions. :+1: 

That said, we agreed (on stream) that the `ExecutionResultAssert::hasNumberOfTests`/etc methods are a bit confusing. They appear to be asserting the number of tests, but they aren't. The problem described in `TestAssert` has the same underlying reason:

```
 * There is an inherent problem with the naming of the `thatStarted()` method that you should be aware of:
 * Imagine the following scenario: You have 4 tests, 3 should fail, one should succeed.
 * The logical way to test that would be to write
 * <p>
 *     assertThat(results).hasNumberOfTests(3).thatStarted().thenFailed();
 *     assertThat(results).hasSingleTest().thatStarted().thenSucceeded();
 * </p>
 * Looks good, does not work: it fails on the first line because there were actually 4 tests that started.
```

The only solution we found so far is to "flatten" the API and provide detailed assertion methods directly on `ExecutionResultAssert` or `TestAssert`. The latter would look as follows:

```
interface ExecutionResultAssert {
	
	TestAssert tests();

	// same for container (and report entries?)
	
}

interface TestAssert {
	
	TestAssert hasTests(int ex);
	
	TestAssert hasSuccessfulTests(int ex);
	
	FailedTestAssert hasFailedTests(int ex);
	
}

interface FailedTestAssert extends TestAssert  {

	AbstractThrowableAssert<?, ? extends Throwable> withException(Class<? extends Throwable> exceptionType);

	AbstractThrowableAssert<?, ? extends Throwable> withException();
	
}
```

This would bloat `TestAssert` by making all combinations of state/cardinality explicit methods. Yet, it seems to be the best solution.

nicolaiparlog:
Also, interface implementations should probably all be package-visible and I'm not convinced about `EntryPoint`. :monocle_face:

Bukama:
Please merge the current master into the branch, to remove the Gradle issues mentioned in #290 - thank you!

nicolaiparlog:
Looks good to me. Proposed commit message:

```
Create assertions for Pioneer (#232 / #245)

In 0b79db8 (#6 / #218), we moved all assertion-like methods onto the
`ExecutionResults` class even though that wasn't a good fit. The
intention was to collect all such methods to then more easily replace
them with proper AssertJ-like assertions that we needed to write
ourselves.

This change implements these methods. The API is pretty good already,
but we expect that after using it for a while the experience with it
as well as new use cases may lead to further changes.

Closes: #232
PR: #245
```

PR #248, opened 2020-04-30
Reduce duplication

### LINKED ISSUE #180: Reduce duplication
Once #173, #174, #187 are merged, there's some duplication in the code that we could remove it seems helpful to do that:

* `EnvironmentVariableExtension` and `SystemPropertyExtension` are basically identical
* `findRepeatableAnnotations` should be replaceable by the new method in `Util`

More?

beatngu13:
This is an early draft to fix #180.

In order to reduce the code duplication among `SystemPropertyExtension` and `EnvironmentVariableExtension`, I thought about an abstract base class for entry-based extensions, where entries (i.e. key-value pairs) shall be cleared or set.

As outlined above, this is a draft. There may be still some design flaws, ~the Javadoc is incomplete, no type wildcards~ … I just wanted to get some feedback if you think it is worth to further work on this, use a different approach or leave the extensions as they are. ✌️

Personally, I quite like it because when you look at [`SystemPropertyExtension`](https://github.com/junit-pioneer/junit-pioneer/blob/5a301100f5770ef57b2f1ed601e98f0f18dea700/src/main/java/org/junitpioneer/jupiter/SystemPropertyExtension.java) and [`EnvironmentVariableExtension`](https://github.com/junit-pioneer/junit-pioneer/blob/5a301100f5770ef57b2f1ed601e98f0f18dea700/src/main/java/org/junitpioneer/jupiter/EnvironmentVariableExtension.java) now, they are dead-simple.

---

### DIFF
--- a/src/main/java/org/junitpioneer/jupiter/AbstractEntryBasedExtension.java
+++ b/src/main/java/org/junitpioneer/jupiter/AbstractEntryBasedExtension.java
@@ -1,0 +1,170 @@
+/*
+ * Copyright 2015-2020 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junitpioneer.jupiter;
+
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.stream.Stream;
+
+import org.junit.jupiter.api.extension.AfterAllCallback;
+import org.junit.jupiter.api.extension.AfterEachCallback;
+import org.junit.jupiter.api.extension.BeforeAllCallback;
+import org.junit.jupiter.api.extension.BeforeEachCallback;
+import org.junit.jupiter.api.extension.ExtensionConfigurationException;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.extension.ExtensionContext.Namespace;
+
+/**
+ * An abstract base class for entry-based extensions, where entries (key-value
+ * pairs) can be cleared or set.
+ *
+ * @param <K> The entry's key type.
+ * @param <V> The entry's value type.
+ */
+abstract class AbstractEntryBasedExtension<K, V>
+		implements BeforeAllCallback, BeforeEachCallback, AfterAllCallback, AfterEachCallback {
+
+	private static final Namespace NAMESPACE = Namespace.create(AbstractEntryBasedExtension.class);
+	private static final String BACKUP = "Backup";
+
+	/**
+	 * @param context The current extension context.
+	 * @return <code>true</code> if one or more of the extension's annotations are present.
+	 */
+	protected abstract boolean isAnnotationPresent(ExtensionContext context);
+
+	/**
+	 * @param context The current extension context.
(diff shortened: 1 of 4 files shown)


Bukama:
Wow - nice duplicate code reduction so far! Like the approach and interested in how you go further

src/main/java/org/junitpioneer/jupiter/AbstractEntryBasedExtension.java
```diff
@@ -0,0 +1,170 @@
...
+		entriesToClear.forEach(clearer());
+	}
+
+	private void setEntries(Map<K, V> entriesToSet) {
+		entriesToSet.forEach(setter());
+	}
+
+	@Override
+	public void afterEach(ExtensionContext context) throws Exception {
+		if (isAnnotationPresent(context))
```
beatngu13, line 135: I wonder if this check is necessary? If not, we could also get rid of the abstract `isAnnotationPresent` method.

beatngu13:
I think I'm done for the most part. I would keep `AbstractEntryBasedExtension` internal (package private), but in the future this may become public if users want to extend it for similar purposes.

Note that subclasses of `AbstractEntryBasedExtension` are currently responsible for discovering their own annotations. In a second step, we might move additional code into `AbstractEntryBasedExtension`. However, for now I think the way it is implemented is already a good improvement.

Proposed commit message:

```
Reduce duplication in envar and sysprop extensions

This PR reduces the duplication among `EnvironmentVariableExtension` and `SystemPropertyExtension` by introducing an abstract base (`AbstractEntryBasedExtension`) class for entry-based extensions, where entries (key-value pairs) can be cleared or set.

Closes: #180
PR: #248
```

Bukama:
Will have a look, hope tomorrow

beatngu13:
Open question: As with most traditional refactorings, I didn't touch existing tests. But do we want to have some additional tests for the new base class?

Bukama:
> Open question: As with most traditional refactorings, I didn't touch existing tests. But do we want to have some additional tests for the new base class?

I vote for yes. And if there are existing tests of the old extension which, in core, test the same, because the functionality moved to the base class, than we should merge them into tests for the base class.

Bukama:
> Open question: As with most traditional refactorings, I didn't touch existing tests. But do we want to have some additional tests for the new base class?

I asked @beatngu13 if he wants to add tests for the base class and we wants so. So we wait for an update :)

nicolaiparlog:
When I saw that we need six abstract methods, I was worried that the abstraction would the very fragile, bit it looks pretty good. Thank you very much @beatngu13! :+1: I was a bit surprised by the methods returning functional interfaces and turned that into "normal" methods (0c72122) - hope that's ok with you.

From my perspective, this can be merged.

Bukama:
> From my perspective, this can be merged.

@beatngu13 wanted to add some additional tests. After that is done we can merge.

nicolaiparlog:
I looked at this for a few minutes, wondering what tests to add.

With abstract base classes, I usually try to come up with an abstract base class for the tests as well. The concrete classes can then be tested by concrete test classes that extend the base class. In that scenario, actual tests are usually only in the abstract test base class and concrete test classes only implement some factory methods. I don't think this will work here because I know of no way to handle the annotations (e.g. `@SetEnvironmentVariable`) - whether you try to put them into the abstract base class or the implementation, there are weird consequences.

I'm ok with just testing the implementation, i.e. just `EnvironmentVariableExtension` and `SystemPropertyExtension` without any tests for the abstract base class.

If you still want to add tests, keep in mind that we don't want to "overtest", i.e. if there's a bug in the abstract base class, we don't need its tests _and_ the implementations to fail if that just adds more failing tests without providing more insight.

beatngu13:
@nicolaiparlog makes sense, I'm fine with that. 👍

Just did a rebase to current `master`, so feel free to merge.

PR #249, opened 2020-05-01
Removed weird quotation marks.

Michael1993:
---


Bukama:
There are several more places where they are used inside the `ReportEntryExtensionTest`. The line I posted in Discord was only one of them.

And to note again: I don't why they cause trouble: The files are perfectly UTF-8 (as notepad says) and I think it's windows terminal, which causes those messages:

```
ReportEntryExtensionTests.java:377: error: unmappable character for encoding Cp1252
                                                                "And the only word there spoken was the whispered word, ÔÇ£Lenore?ÔÇ?",
```

But I also had them, when I started the build form IntelliJ. After I applied `spotlessJava`, which resulted in no changed files it works. Very strange to me :(

PR #253, opened 2020-05-02
Improve thread safety / parallel execution of extensions

### LINKED ISSUE #131: DefaultLocale/DefaultTimeZone - ExecutionMode
Both _DefaultLocale_ and _DefaultTimeZone_ can only be realistically used with [@Execution(SAME_THREAD)](https://junit.org/junit5/docs/current/api/org/junit/jupiter/api/parallel/ExecutionMode.html#SAME_THREAD) or a [ResourceLock](https://junit.org/junit5/docs/current/api/org/junit/jupiter/api/parallel/ResourceLock.html).

Maybe it should be enforced on a programmatic level?

Or their documentation should have pointers to Execution/ResourceLock.

See also:
https://junit.org/junit5/docs/current/user-guide/#writing-tests-parallel-execution-synchronization

Bukama:
```
To check extensions thread safety, the parallel execution of tests were enabled.
It turned out that several extensions are not thread safe and need to be 
executed in a single thread or need resource locks. The needed
annotations were added.

Also the execution via the TestKitEngine was set do single threaded.
Marc Philipp (team lead JUnit) said that it's not expected behaivor of the engine
to pick up the concurrency settings by default.

fixes #131
PR: #253


---

### CODE REVIEW

**Comment by NPException** [comment]
Can you include the recently added `StdIOExtension` in this PR? If I'm not mistaken it should be single-threaded as well.

**Comment by Bukama** [comment]
> Can you include the recently added `StdIOExtension` in this PR? If I'm not mistaken it should be single-threaded as well.

Branch updated

**Comment by Michael1993** [comment]
Weird, the tests are failing.

**Comment by Bukama** [comment]
> Weird, the tests are failing.

Locally the build is successful

> BUILD SUCCESSFUL in 27s

**Comment by nicolaiparlog** [comment]
Note: We should add some example on threads-safety to the docs.

**Comment by nicolaiparlog** [comment]
I executed the build twice and got two different errors: :sob: 

On Java 8, `DefaultLocaleTests.MethodLevelTests::setsLocaleViaLanguageTag` read "custom" locale (i.e. `TEST_DEFAULT_LOCALE`), not "zh-Hant-TW". I assume the reason is that another test that sets the locale was executed in parallel and the locale reset at the end happened between the failing test's before and its assertion.

I wonder whether it may be a good idea to compare the current default locale with the one the extension as configured to set before resetting it to the old value and throw an error or log a warning. If another process interfered, then we could discover that here and probably provide better information than the random piece of code that fails downstream.

On Java 11:

```
EnvironmentVariable extension > ReportWarningTests > shouldReportWarningExactlyOnce() FAILED
    java.lang.AssertionError: 
    Expected size:<1> but was:<0> in:
    <[]>
        at org.junitpioneer.jupiter.EnvironmentVariableExtensionTests$ReportWarningTests.shouldReportWarningExactlyOnce(EnvironmentVariableExtensionTests.java:273)
```

**Comment by nicolaiparlog** [comment]
Proposed commit message:

```
Make extensions thread-safe (#131 / #253)

Pioneer's extensions should be thread-safe. To have a better chance
to discover where they aren't and to test whether they are, this
change switches our test execution to full parallel mode (see
src/test/resources/junit-platform.properties and `ExecutionResults`).
It also evaluates, improves, and documents actual thread-safety.

Closes: #131
References: #274, #276, #277
PR: #253
```

**Comment by Bukama** [comment]
Locally the build runs, on GithubAction it fails with permission problems (as far as I understand the log)

> 2020-07-20 11:32:12,131 pool-1-thread-1 ERROR Could not unregister MBeans for org.apache.logging.log4j2:type=6d06d69c,component=Loggers,name=* java.security.AccessControlException: access denied ("javax.management.MBeanPermission" "-#-[-]" "queryNames")

**Comment by nicolaiparlog** [comment]
This is a tough nut to crack. I observed at least three different behaviors:

* 537e6ca [failed](https://github.com/junit-pioneer/junit-pioneer/runs/790564921) because of an assertion (something was 0 that should've been 1 or the other way around)
* 86bcf46 [failed](https://github.com/junit-pioneer/junit-pioneer/runs/895237111) for some weird "Log4J2 + MBean"-thingy
* same commit blocks for minutes on my machine, reliably on `DefaultTimeZoneTests$MethodLevelTests`

At the moment, I'm convinced that the last two can be explained by the merger from `master`, which pulled in tests that configure a stricter security manager. Tests running concurrently with those tests may run into security manager issues. I don't know a good solution to that. Quite the opposite, I wonder whether this is indicative of a larger problem:

1. There seems to be no simple way to express that, despite all the parallelism that was enabled for the other tests, the security-manager tests should be run on their own. The only solution seems to be resource locks that need to be applied to *all* tests.

2. The security manager tests behave similarly to the many extensions we have that change global state (e.g. default locale) and arguably, the same problem can arise there: While we can use resource locks to keep tests using our extensions from treading on one another, we can't prevent "unextended" tests from running concurrently to extended ones. The user would have to apply respective annotations. Which sucks.

Right now it feels like this is a serious problem for all extensions that touch global state that unextended tests may also use: There is no way to run these concurrently without a adding annotations to tests that are seemingly unrelated to the extended ones. Worse, it may be very hard to even find the ones that need to be forced into single threads.

I wonder whether an [`ExecutionMode`](https://junit.org/junit5/docs/5.3.0/api/org/junit/jupiter/api/parallel/ExecutionMode.html) `SingleThreaded` would make sense...

**Comment by nicolaiparlog** [comment]
Dear @junit-pioneer/junit-5-team, could you please take a look at this? Start with the previous comment - ideally, there's nothing else you need to look at in this PR.

**Comment by marcphilipp** [comment]
Have you thought about executing the tests that use a different security manager to a separate Gradle test task, e.g. using tags? I could help with the wiring, if need be.

**Comment by nicolaiparlog** [comment]
Yes, we thought about that. We could create a second, non-parallelized test phase (different `junit-platform.properties`?) where all tests that need to run in a single thread would get dumped into. 

It feels like a cop-out, though. Why should a random property of some tests raise to the level of importance where they change the build script? And not just ours, either: If I didn't miss anything above, this can just as well happen to users (of Jupiter and of Pioneer) if some of their tests mutate global state. I'm under the impression that Jupiter would be the best place to solve this problem at the root by allowing a mostly parallelized test suite where some tests still run on their own.

**Comment by marcphilipp** [comment]
There's an issue for that: https://github.com/junit-team/junit5/issues/2142 😉

**Comment by nicolaiparlog** [comment]
Great, we'll chime in there with our experience. Thanks for pointing us to it.

**Comment by nicolaiparlog** [comment]
Thanks @Michael1993 for pointing me into the right direction on the env var test failures. For the record, this is my current understanding of how locking in parallel tests works:

* each engine instance has its own thread pool
* each engine instance uses its own set of locks

That means if a parallel test suite (like ours, where methods in a test are parallelized) launches several Jupiter engine instances at the same time (like those in `EnvironmentVariableExtensionTests.ConfigurationFailureTests` and in `EnvironmentVariableExtensionTests.ReportWarningTests`), then _within each engine_ no two tests can get write access to the same lock, but _across several engines_ they can.

In this instance, we fixed that (in d685bd4) by making sure that a misconfigured env var extension (as tested by `ConfigurationFailureTests`) does not interact with properly configured extensions by racing to report a warning.

**Comment by nicolaiparlog** [comment]
As it stands, at least three tests occasionally fail:

* [Default Time Zone](https://github.com/junit-pioneer/junit-pioneer/pull/253/checks?check_run_id=940542388)
* [Default Locale](https://github.com/junit-pioneer/junit-pioneer/pull/253/checks?check_run_id=940414365)
* [EnvVars vs SecurityManager](https://github.com/junit-pioneer/junit-pioneer/pull/306/checks?check_run_id=940516709)

**Comment by marcphilipp** [comment]
> Did I get this right?

Yes, that's right.

**Comment by nicolaiparlog** [comment]
Each of the three occasionally failing tests listed above were addressed. The last two ([Default Locale](https://github.com/junit-pioneer/junit-pioneer/pull/253/checks?check_run_id=940414365) and [EnvVars vs SecurityManager](https://github.com/junit-pioneer/junit-pioneer/pull/306/checks?check_run_id=940516709)) can be executed tens of thousands of times without failure. (Thanks to @Michael1993 for letting me know [how to do that](https://twitter.com/nipafx/status/1290553157748166656).)

[Default Time Zone](https://github.com/junit-pioneer/junit-pioneer/pull/253/checks?check_run_id=940542388) also never fails, but deadlocks every few thousand executions. I have no idea why. @junit-pioneer/junit-5-team Could you take a look?

**Comment by nicolaiparlog** [comment]
"Every few thousand executions" ~> [this execution](https://github.com/junit-pioneer/junit-pioneer/pull/253/checks?check_run_id=944254563)

**Comment by nicolaiparlog** [comment]
Since everybody already signed off on this, I will merge it.

No way will we come back to this PR in the future if flaky tests rear their head. No way, I say!

File: src/test/java/org/junitpioneer/jupiter/DefaultTimeZoneTests.java [hunk 1]
```diff
@@ -35,16 +34,15 @@
...
-		// the system's time zone; in that case it uses GMT
+		// we set UTC as test time zone unless it is already
+		// the system's time zone; in that case we use UTC+12
 		DEFAULT_TIMEZONE_BEFORE_TEST = TimeZone.getDefault();
 		TimeZone utc = TimeZone.getTimeZone("UTC");
-		TimeZone gmt = TimeZone.getTimeZone("GMT");
-		if (DEFAULT_TIMEZONE_BEFORE_TEST.equals(utc)) {
-			TimeZone.setDefault(gmt);
-		} else {
+		TimeZone utcPlusTwelve = TimeZone.getTimeZone("GMT+12:00");
```
NPException, line 41: @nicolaiparlog Shouldn't this technically be "UTC+12:00" if you want UTC+12? There can be a tiny difference between UTC and GMT if I remember correctly. Probably not relevant for the tests here, but I thought I mention it anyway.

Here's a video by Tom Scott that mentions said difference: https://www.youtube.com/watch?v=h0AMaW4XRCI
(It's entirely possible that I'm misinterpreting the things said in the video though...)

nicolaiparlog:
```
jshell> TimeZone.getTimeZone("UTC+12:00")
$1 ==> sun.util.calendar.ZoneInfo[id="GMT",offset=0,dstSavings=0,useDaylight=false,transitions=0,lastRule=null]
```

:man_shrugging: 

That's why I picked `TimeZone.getTimeZone("GMT+12:00")` instead.

src/main/java/org/junitpioneer/jupiter/StdIoExtension.java
```diff
@@ -33,7 +33,7 @@
 
 	private static final String SYSTEM_IN_KEY = "StdIo_System_In";
 	private static final String STD_IN_KEY = "StdIo_Std_In";
-	private static final String OUT_KEY = "StdIo_Std_Out";
+	private static final String SYSTEM_OUT_KEY = "StdIo_Std_Out";
```
NPException, line 36: I'd expect the value for this to be `"StdIo_System_Out"` if it's used to store `System.out`.

src/test/java/org/junitpioneer/jupiter/DefaultTimeZoneTests.java
NPException: I see. That's odd. But at least you tried :sweat_smile:

src/main/java/org/junitpioneer/jupiter/StdIoExtension.java
nicolaiparlog: Good point!

nicolaiparlog:
Since @NPException seems to be looking this over, I will let it sit for some more time.

Proposed commit message:

```
Improve thread safety / parallel execution of extensions (#131 / #253)

It must be safe to use Pioneer's extensions in a test suite that is
executed in parallel. This change enables that for all extensions
except TempDirectory (see #227). It does that in various ways.

# Resource locks

For extensions touching global state (like default locales or
environment variables), we've chosen the following approach:

* the extension acquires a read/write lock to the global resource
  (this prevents extended tests from running in parallel)
* we offer a `@Writes...` annotation that does the same thing, so
  users can annotate their tests that write to the same resource and
  prevent them from running in parallel with each other and with
  extended tests
* we offer a `@Reads...` annotation that acquires read access to the
  same lock, so users can make sure such tests do not run in parallel
  with tests that write to the same resource (they can run in
  parallel with one another, though)

# Verify, then modify

Most extensions verify their configuration at some point. It helps
with writing parallel tests for them if they do not change global
state until the configuration is verified. That particularly applies
to "store in beforeEach - restore in afterEach"-extensions! If they
fail after "store", they will still "restore" and thus potentially
create a race condition with other tests.

# Testing

To have a better chance to discover threading-related problems in our
extensions, we parallelize our own tests (configured in
`junit-platform.properties`). Ideally, we'd like to run them in
parallel _across_ and _within_ top-level classes, but unfortunately,
this leads to problems[1] when some test setups change global state
(like the security manager) that other tests rely on. As we see it,
the solution would be to force such tests onto a single thread, but
Jupiter has no such feature, yet[2]. While a homegrown solution is
possible[3], we wait for the discussion to resolve. We hence do not
parallelize across top-level classes - just within.

Closes: #131
Relates to #274
PR: #253

[1]: https://github.com/junit-pioneer/junit-pioneer/pull/253#issuecomment-665235062
[2]: https://github.com/junit-team/junit5/issues/2142
[3]: https://github.com/junit-team/junit5/issues/2142#issuecomment-668409251
```

src/test/java/org/junitpioneer/jupiter/DefaultTimeZoneTests.java
NPException: Looks like UTC is technically not a timezone, which is why `TimeZone.getTimeZone("UTC+12:00")` will return GMT as a default instead. From https://www.timeanddate.com/time/gmt-utc-time.html:

> UTC is not a time zone, but a time standard that is the basis for civil time and time zones worldwide. This means that no country or territory officially uses UTC as a local time.

Today I learned.

nicolaiparlog:
Just for the record, I forgot to mention @Michael1993's great help by trying out and pushing his fixes to #306.

PR #256, opened 2020-05-02
Trigger Release with tag 'releaseTrigger'

aepfli:
This changes will trigger a build on master,
when there is a tag 'releaseTrigger' created.

It will notify our normal GitHub Actions, and
retrigger a build. The payload will tell the
our normal build, that this is a release, and
it will trigger the build.

Downside, this will only work in MASTER for now.
If we want to support other branches, we need
to use our combined brain power, how to retrieve
a branch name from a git tag. It is not such
a trivial problem.

Questions:
1. is it okay that we can only release on master?
2. does this work for everyone with the tagging?
3. shall we move the deletion of the tag also into the triggering?

---


Bukama:
> 
> Downside, this will only work in MASTER for now.
> If we want to support other branches, we need
> to use our combined brain power, how to retrieve
> a branch name from a git tag. It is not such
> a trivial problem.

Tags are not related to branches, but to commits.


> 
> Questions:
> 
> 1. is it okay that we can only release on master?

Sure, same as of now

> 2. does this work for everyone with the tagging?

That made me think - can everyone create a tag and start a release? Or only maintainers, who have write permissions on the repo? 

> 3. shall we move the deletion of the tag also into the triggering?

In my opinion it should be the last step of the process, so when the release was successful the tag is gone.

aepfli:
@beatngu13 as you are also deeper into this build pipeline issues, what do you think about this approach, splitting deployment into own yaml and waiting of all other suites which are github actions, to finish?

.github/workflows/releaseTrigger.yml
```diff
@@ -0,0 +1,46 @@
...
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    name: releasing
+    steps:
+      - name: Wait for build to succeed
+        uses: jitterbit/await-check-suites@v1
+        with:
+          appSlugFilter: 'github-actions'
+          timeoutSeconds: '300'
```
nicolaiparlog, line 20: This means we wait for the build in `actions.yml` to complete successfully? Is that job uniquely identified, i.e. what happens if we ever add a third `yml` for whatever reason? Could this step here then be completed by that other job workflow?

.github/workflows/releaseTrigger.yml
```diff
@@ -0,0 +1,46 @@
...
+      - name: Setup java
+        uses: actions/setup-java@v1
+        with:
+          java-version: 8
+      - name: Perform Release
+        uses: eskatos/gradle-command-action@v1
+        env:
+          TRAVIS_BRANCH: master
+          TRAVIS_COMMIT_MESSAGE: <no commit message, not needed, we do not care about them, if we trigger or not>
+          TRAVIS_PULL_REQUEST: false
```
nicolaiparlog, line 31: The `TRAVIS_` is necessary because of ShipKit, correct? If so, can you add a comment here that clarifies that?

.github/workflows/releaseTrigger.yml
```diff
@@ -0,0 +1,46 @@
...
+          GH_WRITE_TOKEN: ${{ secrets.GH_WRITE_TOKEN }}
+          BINTRAY_USER_NAME: ${{ secrets.BINTRAY_USER_NAME }}
+          BINTRAY_API_KEY: ${{ secrets.BINTRAY_API_KEY }}
+          NEXUS_TOKEN_USERNAME: ${{ secrets.NEXUS_TOKEN_USERNAME }}
+          NEXUS_TOKEN_PASSWORD: ${{ secrets.NEXUS_TOKEN_PASSWORD }}
+        with:
+          arguments: ciPerformRelease
+      - name: Delete Tag
+        run: git push --delete origin releaseTrigger
```
nicolaiparlog, line 40: This means a broken release requires manual cleanup of the tag. Would that be a reason to perform this step much earlier, i.e. as the first one?

.github/workflows/releaseTrigger.yml
aepfli: > This means we wait for the build in actions.yml to complete successfully?

yes

> Is that job uniquely identified, i.e. what happens if we ever add a third yml for whatever reason?

No currently we wait for all test suites which are running on the same git SHA to be finished, not one particular, but all checks from github actions

> Could this step here then be completed by that other job workflow?

Not completed but it will be a necessary step to wait for

.github/workflows/releaseTrigger.yml
aepfli: that is a really good question -> we also should think ob some kind of notification in the error case. Or we can also provide a flag, which will continue the job, even if it failed. That is something hard to tell, which we should define. I do not have any special preferences regarding that :)

.github/workflows/releaseTrigger.yml
nicolaiparlog: That's good to know (can you add a short comment explaining that?). I wondered whether we should try to limit this to the actual checks that are currently configured, but I think waiting for all to be successful (they need to be successful, right?) is safer. This way, when we add a new job, the worst that can happen is the release not getting triggered.

.github/workflows/releaseTrigger.yml
nicolaiparlog: I think it's bad enough that we abuse tags this way, we shouldn't go further and build an entire protocol based on them. :wink: The check marks next to each commit will show how the build went. Additionally, you can configure Bintray to send you an email for each release.

beatngu13: I think we can use this feature:

https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#job-status-check-functions

Per default, steps only run "when none of the previous steps have failed or been canceled". Using `if: ${{ always() }}` seems to be a good fit for clean-up tasks like this.

(Sorry, haven't found time for a full review so far.)

aepfli: > Per default, steps only run "when none of the previous steps have failed or been canceled". Using `if: ${{ always() }}` seems to be a good fit for clean-up tasks like this.

added

.github/workflows/releaseTrigger.yml
aepfli: done

.github/workflows/releaseTrigger.yml
aepfli: added comment

.github/workflows/releaseTrigger.yml
beatngu13: @aepfli could it be that you forgot the condition in your commit 254583d?

aepfli: you are right, sorry my bad -.-

.github/workflows/releaseTrigger.yml
```diff
@@ -0,0 +1,51 @@
+# This workflow will build a Java project with Gradle
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-gradle
+
+name: Release
+
+on:
+  create:
```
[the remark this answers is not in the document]
aepfli: No there is no need for a trigger on master. We do not want to publish on master pushes, but as soon as the tag is created. So we can trigger a deployment without caring about commit messages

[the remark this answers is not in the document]
beatngu13: I always forget that tags are not bound to branches, but commits. 🤦‍♂️

.github/workflows/releaseTrigger.yml
```diff
@@ -0,0 +1,51 @@
+# This workflow will build a Java project with Gradle
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-gradle
```
aepfli, line 2:
```suggestion
# This workflow will build a Java project with Gradle
# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-gradle
```

.github/workflows/releaseTrigger.yml
```diff
@@ -0,0 +1,51 @@
...
+  create:
+    tags:
+      - 'releaseTrigger'
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    name: releasing
+    steps:
+      - name: Wait for build to succeed
```
aepfli, line 16:
```suggestion
      - name: Wait For Other Builds To Succeed
```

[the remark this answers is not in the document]
aepfli: renamed

PR #257, opened 2020-05-05
Fix sorting of 2020's contributors

mureinik:
Fix the sorting of 2020's contributors so they are actually sorted in lexicographical order (of first names, then surnames), as documented.

---


Bukama:
Thank you!

PR #261, opened 2020-05-10
Introduce Issue-Extension with XML-Report (#135 / #261)

### LINKED ISSUE #135: Add declarative annotations for "issue" type support and infrastructure

Bukama:
This PR shows an implementation of an issue extension (#135), which writes the results to a XML-report

```
Introduce Issue-Extension with XML-Report (#135 / #261)

This PR provides the @Issue-Annotation and its extension,
which allows users to show that tests belong to a special
issue (e.g. requirement).

While this information is not needed during test execution
(e.g. like changing the time zone via @DefaultTimeZone-
annotation), but an information needed in a report to check
which tests of which issues have passed/failed this PR also
introduces the IssueExtensionReportListener.
This listener creates an additional test report. For the
moment this report contains the issues and the related test
cases with their result.

For example a report entry for an test with the issue
annotation may look like this (line breaks in test case cause
of commit message):

<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<issueReport>
    <issues>
        <issue issueId="bishue 135">
            <testCases>
                <testCase name="[engine:junit-jupiter]/
                [class:org.junitpioneer.jupiter.IssueTests]/
                [method:writeToReport()]" status="SUCCESSFUL"/>
            </testCases>
        </issue>
    </issues>
</issueReport>

The new/own report is needed, because the default JUnit Jupiter report
can't be extended and the JUnit Jupiter teams don't want to provide
a way to add custom information until a default test format for all
test frameworks and IDE is defined (see 
https://github.com/ota4j-team/opentest4j/issues/9 for more information).

With providing such a report further tools may import the report
and extract information out of it (and e.g. to provide them
within a CI-build).

The report is created and validated as a XML-File using the default java
mechanism JAXB for this. As this libary is not part of Java 11+ JDK
anymore, there was the need to add additional depencies .
In core this violates the JUnit Pioneer guidelines, but as it's a default
java functionality and no "classic" 3rd party dependency it was considered
to be okay.

closed #135
PR: #261

```

closed #135
---


src/main/java/org/junitpioneer/jupiter/Issue.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+@Target({ ElementType.METHOD })
+@Retention(RetentionPolicy.RUNTIME)
+@ExtendWith(IssueExtension.class)
+public @interface Issue {
+
+	/**
+	 * The id of the issue as defined by the issue-tracker, e.g. "REQ-123".
+	 *
+	 * @since 0.6
```
Michael1993, line 46: Same thing here, I think this should be (at least) 0.6.1.

src/main/java/org/junitpioneer/jupiter/IssueExtension.java
```diff
@@ -0,0 +1,71 @@
...
+	/**
+	 * Reads the {@code @Issue} value from the annotation.
+	 * @param context The Extensions context
+	 * @return The read value
+	 */
+	String readIssueIdFromAnnotation(ExtensionContext context) {
+		//@formatter:off
+		return AnnotationSupport
+				.findAnnotation(context.getElement(), Issue.class)
+				.map(IssueExtension::convertIssueIdValueAsString)
```
Michael1993, line 57: This is confusing, why can't this be just `Issue::value`?

src/main/java/org/junitpioneer/jupiter/IssueExtension.java
```diff
@@ -0,0 +1,71 @@
...
+		context.publishReportEntry(KEY, issueId);
+	}
+
+	/**
+	 * Reads the {@code @Issue} value from the annotation.
+	 * @param context The Extensions context
+	 * @return The read value
+	 */
+	String readIssueIdFromAnnotation(ExtensionContext context) {
+		//@formatter:off
```
Michael1993, line 54: Does spotless break this method so much that this becomes necessary? Nicolai worked hard to remove these from the project. :disappointed:

src/main/java/org/junitpioneer/jupiter/package-info.java
```diff
@@ -6,6 +6,7 @@
  *     <li>{@link org.junitpioneer.jupiter.ClearSystemProperty} and {@link org.junitpioneer.jupiter.SetSystemProperty}</li>
  *     <li>{@link org.junitpioneer.jupiter.ClearEnvironmentVariable} and {@link org.junitpioneer.jupiter.SetEnvironmentVariable}</li>
  *     <li>{@link org.junitpioneer.jupiter.DefaultLocale} and {@link org.junitpioneer.jupiter.DefaultTimeZone}</li>
+ *     <li>{@link org.junitpioneer.jupiter.Issue}</li>
```
Michael1993, line 9: Nice, I always forget to update the `package-info.java`. Good work. :sweat_smile:

src/main/java/org/junitpioneer/jupiter/report/PioneerReport.java
```diff
@@ -0,0 +1,166 @@
...
+public class PioneerReport {
+
+	protected Properties properties;
+	protected List<Testcontainer> testcontainer;
+	protected List<Testcase> testcase;
+	@XmlAttribute(name = "name")
+	protected String name;
+
+	public PioneerReport() {
+		this.testcontainer = new ArrayList<Testcontainer>();
```
Michael1993, line 56: Sonar says this can be replaced with the diamond operator `<>` and I have to agree.

src/main/java/org/junitpioneer/jupiter/report/PioneerReport.java
```diff
@@ -0,0 +1,166 @@
...
+	 *
+	 *
+	 * <p>
+	 * Objects of the following type(s) are allowed in the list
+	 * {@link Testcontainer }
+	 *
+	 *
+	 */
+	public List<Testcontainer> getTestcontainer() {
+		if (testcontainer == null) {
```
Michael1993, line 107: Isn't this *literally impossible*? You initialize this value in the no-args constructor and this field doesn't have a setter. :question:

src/main/java/org/junitpioneer/jupiter/report/PioneerReport.java
```diff
@@ -0,0 +1,166 @@
...
+	 *
+	 *
+	 * <p>
+	 * Objects of the following type(s) are allowed in the list
+	 * {@link Testcase }
+	 *
+	 *
+	 */
+	public List<Testcase> getTestcase() {
+		if (testcase == null) {
```
Michael1993, line 136: Same thing, this should never be `null` unless I'm missing something?

src/main/java/org/junitpioneer/jupiter/report/PioneerReport.java
```diff
@@ -0,0 +1,166 @@
...
+ * </pre>
+ *
+ *
+ */
+@XmlAccessorType(XmlAccessType.FIELD)
+@XmlType(name = "", propOrder = { "properties", "testcontainer", "testcase" })
+@XmlRootElement(name = "pioneerReport")
+public class PioneerReport {
+
+	protected Properties properties;
```
Michael1993, line 49: Why are these fields `protected` instead of `private`?

src/main/java/org/junitpioneer/jupiter/report/PioneerReportListener.java
```diff
@@ -0,0 +1,220 @@
...
+ *
+ * <p>The report contains test containers and its test methods. For each object properties, based on published
+ * report entries are reported. This allows extension to report additional values in the report.
+ * for test cases the result of the test is (of course) reported.</p>
+ *
+ * @see <a href="https://junit.org/junit5/docs/current/api/org.junit.platform.launcher/org/junit/platform/launcher/TestPlan.html">JUnit Docs</a>
+ */
+public class PioneerReportListener implements TestExecutionListener {
+
+	TestPlan storedTestPlan;
```
Michael1993, line 56: Why are these fields `package-private` instead of `private`?

src/main/java/org/junitpioneer/jupiter/report/PioneerReportListener.java
```diff
@@ -0,0 +1,220 @@
...
+		writeReport(pioneerReport);
+	}
+
+	/**
+	 * Converts an {@code TestIdentifier} object into a {@code Testcontainer} for the report.
+	 *
+	 * @param container Object to be converted into {@code Testcontainer}
+	 * @return Created object
+	 */
+	Testcontainer processContainer(TestIdentifier container) {
```
Michael1993, line 112: Shouldn't this method be `private`?

src/main/java/org/junitpioneer/jupiter/report/PioneerReportListener.java
```diff
@@ -0,0 +1,220 @@
...
+		return testcontainer;
+	}
+
+	/**
+	 * Converts an {@code TestIdentifier} object into a {@code Testcase} for the report.
+	 *
+	 * @param test Object to be converted into {@code Testcase} element
+	 * @return Created object
+	 */
+	Testcase processTest(TestIdentifier test) {
```
Michael1993, line 142: Shouldn't this method be `private`?

src/main/java/org/junitpioneer/jupiter/report/PioneerReportListener.java
```diff
@@ -0,0 +1,220 @@
...
+		return testcase;
+	}
+
+	/**
+	 * Retrieves all cached properties for an {@code TestIdentifier} by its unique id.
+	 *
+	 * @param testIdentifierUniqueId Unique id the properties should be loaded for.
+	 * @return Wrapper which contains all cached properties or an empty list.
+	 */
+	Properties getProperties(String testIdentifierUniqueId) {
```
Michael1993, line 162: Shouldn't this method be `private`?

src/main/java/org/junitpioneer/jupiter/report/PioneerReportListener.java
```diff
@@ -0,0 +1,220 @@
...
+
+		return allProperties;
+	}
+
+	/**
+	 * Creates the report and validates it against the XSD.
+	 *
+	 * @param report Report to be created.
+	 */
+	void writeReport(PioneerReport report) {
```
Michael1993, line 177: Shouldn't this method be `private`? (This is just copy-paste at this point. Feel free to address these concerns in a single comment.)

src/main/java/org/junitpioneer/jupiter/report/PioneerReportListener.java
```diff
@@ -0,0 +1,220 @@
...
+	}
+
+	/*
+	 Implementation note:
+	 Package private for test purpose, because the listener overwrites its own file.
+	 When testing the listener itself the report is created the first time (by the tested listener)
+	 and then overwritten by the registered listener after all tests are done.
+	 A system property to configure the report name from outside was seen as too much overhead by the author.
+	 */
+	PioneerReport getPioneerReport() {
```
Michael1993, line 216: This could be `public`. Is there a reason to 'hide' this?

src/test/java/org/junitpioneer/jupiter/IssueTests.java
```diff
@@ -0,0 +1,80 @@
...
+import org.junitpioneer.testkit.ExecutionResults;
+import org.junitpioneer.testkit.PioneerTestKit;
+
+public class IssueTests {
+
+	@Issue("bishue 135")
+	@Test
+	void writeToReport() {
+		// This tests should get covered by the registered PioneerReportListener.
+		// In the report should then be an property with key "issue" and value "bishue 135" for this test
```
Michael1993, line 27: "a property"

src/test/java/org/junitpioneer/jupiter/IssueTests.java
```diff
@@ -0,0 +1,80 @@
...
+
+		/*
+		    <testcase name="writeToReport()" status="SUCCESSFUL">
+		        <properties>
+		            <property name="Issue" value="bishue 135"/>
+		        </properties>
+		    </testcase>
+		 */
+
+		// The report can't be accessed within this test as the file is written after this (and all others)
```
Michael1993, line 37: "(and all other)"

src/test/java/org/junitpioneer/jupiter/IssueTests.java
```diff
@@ -0,0 +1,80 @@
...
+		// tests are finished
+		assertThat(true).isTrue();
+	}
+
+	@Test
+	void checkMethodNotAnnotated() {
```
Michael1993, line 43: This makes it sound like you check if the method is annotated or not. I think this should be called `checkNotAnnotatedMethod` and have a `@DisplayName` for clarity's sake.

src/test/java/org/junitpioneer/jupiter/IssueTests.java
```diff
@@ -0,0 +1,80 @@
...
+		ExecutionResults results = PioneerTestKit
+				.executeTestMethod(IssueTests.IssueDummyTestClass.class, "testNoAnnotation");
+
+		assertThat(results.numberOfSucceededTests()).isEqualTo(1);
+		assertThat(results.reportEntries()).isEmpty();
+	}
+
+	@Test
+	void checkMethodIsAnnotated() {
```
Michael1993, line 52: See above: this should be renamed and *could* have a `@DisplayName`.

src/test/java/org/junitpioneer/jupiter/report/PioneerReportListenerTest.java
```diff
@@ -0,0 +1,158 @@
...
+		PioneerReport report = listener.getPioneerReport();
+		assertThat(report.getProperties()).isNull();
+		assertThat(report.getTestcontainer().isEmpty()).isTrue();
+		assertThat(report.getTestcase().isEmpty()).isTrue();
+	}
+
+	@Test
+	void reportingCorrectCounts() {
+
+		ReportEntry succContainerEntry = ReportEntry.from("key: succContainer", "value: succContainer");
```
Michael1993, line 49: (Are you aware that there succ is a slang term? Just saying...)

src/test/java/org/junitpioneer/jupiter/report/PioneerReportListenerTest.java
```diff
@@ -0,0 +1,158 @@
...
+		assertThat(report.getProperties()).isNull();
+
+		// Verify test containers
+		assertThat(report.getTestcontainer().size()).isEqualTo(4);
+		assertThat(report.getTestcontainer().get(0).getName()).isIn("c1", "c2", "c3", "c4");
+
+		for (Testcontainer testcontainer : report.getTestcontainer()) {
+			if (testcontainer.getName().equals("c1")) {
+				Property property = testcontainer.getProperties().getProperty().get(0);
+				assertThat(property.getName()).isEqualTo("key: succContainer");
```
Michael1993, line 102: Could you move the assertions out from the loop? I think they would look better.

Michael1993: Also if you move them out this can be replaced with a `Stream`.

src/main/java/org/junitpioneer/jupiter/Issue.java
```diff
@@ -0,0 +1,50 @@
...
+ * <p>{@code @Issue} can be used on the method level only.
+ * This is done by purpose, so that the test class can contain tests
+ * which cover the professional requirements but also tests which are only
+ * exist to ensure technical functionality / implementation details.</p>
+ *
+ * <p>{@code @Issue} can only be used once per method.
+ * This is done by purpose because a test case should only cover exactly
+ * one aspect of a method.</p>
+ *
+ * @since 0.6
```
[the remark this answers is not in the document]
Bukama: I'll remove it, since I don't know if at all or at version this PR will get accepted and merged.

src/main/java/org/junitpioneer/jupiter/IssueExtension.java
Bukama: Because I don't know streams ;)

src/main/java/org/junitpioneer/jupiter/IssueExtension.java
Bukama: This was written several months ago, long before the formatting update. 👼 Updated it 👍

src/main/java/org/junitpioneer/jupiter/report/PioneerReport.java
Bukama: As the report files are generated I suggest to exclude all of them.

src/main/java/org/junitpioneer/jupiter/report/PioneerReportListener.java
Bukama: Because I rarely create private fields, so tests are able to access them.

src/main/java/org/junitpioneer/jupiter/report/PioneerReportListener.java
Bukama: I don't create private methods, so tests are able to access them.

src/main/java/org/junitpioneer/jupiter/report/PioneerReportListener.java
Bukama: I don't create private methods, so tests are able to access them.

src/main/java/org/junitpioneer/jupiter/report/PioneerReportListener.java
Bukama: I don't create private methods, so tests are able to access them.

src/main/java/org/junitpioneer/jupiter/report/PioneerReportListener.java
Bukama: I don't create private methods, so tests are able to access them.

src/main/java/org/junitpioneer/jupiter/report/PioneerReportListener.java
Bukama: This method only exists for tests and if it would be possible I won't have created it at all.

src/test/java/org/junitpioneer/jupiter/IssueTests.java
Bukama: Will rename it and add a `@DisplayName`

src/test/java/org/junitpioneer/jupiter/report/PioneerReportListenerTest.java
Bukama: I'm not, but I always write _successful_ wrong (mixing single s and double f instead of correct double s and single f).

Renamed

src/main/java/org/junitpioneer/jupiter/report/PioneerReport.java
Bukama: I changed the diamond operators, but I won't change the "code duplication", because they are no duplication.

src/test/java/org/junitpioneer/jupiter/report/PioneerReportListenerTest.java
Bukama: Updated

build.gradle.kts
```diff
@@ -23,9 +23,13 @@ repositories {
...
     implementation(group = "org.junit.jupiter", name = "junit-jupiter-params", version = "5.4.2")
+    implementation(group = "org.junit.platform", name = "junit-platform-launcher", version = "1.4.2")
+    // For Java 8 provided would be enough as its part of the JRE,
+    // but Java 11 and 14 need this dependency cause of Jigsaw
+    implementation(group = "javax.xml.bind", name ="jaxb-api", version = "2.3.1")
+    implementation(group = "com.sun.xml.bind", name ="jaxb-impl", version = "2.3.3")
 
     testImplementation(group = "org.junit.jupiter", name = "junit-jupiter-engine", version = "5.4.2")
-    testImplementation(group = "org.junit.platform", name = "junit-platform-launcher", version = "1.4.2")
     testImplementation(group = "org.junit.platform", name = "junit-platform-testkit", version = "1.4.2")
```
aepfli, line 33: i would not split the dependencies on junit platfrom. i think , that same packages should stick together. never the less, if they are used for test or implementation. It makes upgrading easier, because the version which needs to be changed are closer to each other

src/main/java/org/junitpioneer/jupiter/report/PioneerReportListener.java
aepfli: i am not sure i do think that this is a good practice. i know it makes writing tests easier, but it also allows manipulation of the fields, which can cause unintended behaviour.

build.gradle.kts
Bukama: I moved the dependency, because it has another scope.

Bukama:
> i am not sure i do think that this is a good practice. i know it makes writing tests easier, but it also allows manipulation of the fields, which can cause unintended behaviour.

If I want to be able to access the things I have no other options, except module system I think.

Bukama:
IssueExtension has 92.9 percent code coverage, ReportEntryListener 93.9 percent.

I vote for excluding/ignoring the generated report classes, because I don't see a benefit to write tests for simple default constructors, setters and getters.

src/main/java/org/junitpioneer/jupiter/report/PioneerReportListener.java
```diff
@@ -0,0 +1,213 @@
...
+	TestPlan storedTestPlan;
+	PioneerReport pioneerReport;
+
+	ConcurrentHashMap<String, List<Property>> propertyCache = new ConcurrentHashMap<>();
+	ConcurrentHashMap<String, String> testStatusCache = new ConcurrentHashMap<>();
+
+	@Override
+	public void reportingEntryPublished(TestIdentifier testIdentifier, ReportEntry entry) {
+		String id = testIdentifier.getUniqueId();
+
```
Michael1993, line 65: You are always going to add all of the properties, so why not do it like this:
```
String id = testIdentifier.getUniqueId();
propertyCache.putIfAbsent(id, new ArrayList<>());

entry.getKeyValuePairs().forEach((key, value) -> propertyCache.get(id).add(new Property(key, value)));
```

src/main/java/org/junitpioneer/jupiter/report/Properties.java
```diff
@@ -0,0 +1,79 @@
...
+ *
+ *
+ */
+@XmlAccessorType(XmlAccessType.FIELD)
+@XmlType(name = "", propOrder = { "property" })
+@XmlRootElement(name = "properties")
+public class Properties {
+
+	@XmlElement(required = true)
+	protected List<Property> property;
```
Michael1993, line 47: This should be called propertyList, at the very least. Right now it is _very_ confusing.

src/main/java/org/junitpioneer/jupiter/report/PioneerReportListener.java
```diff
@@ -0,0 +1,213 @@
...
+	 * @return Created object
+	 */
+	Testcase processTest(TestIdentifier test) {
+		String status = testStatusCache.getOrDefault(test.getUniqueId(), "UNKNOWN");
+		Testcase testcase = new Testcase(test.getDisplayName(), status);
+
+		// Add properties, if there are any
+		Properties allProps = getProperties(test.getUniqueId());
+
+		if (!allProps.getProperty().isEmpty()) {
```
Michael1993, line 149: Is this check necessary? What would happen if you just set the properties without this check?

src/main/java/org/junitpioneer/jupiter/report/PioneerReportListener.java
```diff
@@ -0,0 +1,213 @@
...
+	/**
+	 * Retrieves all cached properties for an {@code TestIdentifier} by its unique id.
+	 *
+	 * @param testIdentifierUniqueId Unique id the properties should be loaded for.
+	 * @return Wrapper which contains all cached properties or an empty list.
+	 */
+	Properties getProperties(String testIdentifierUniqueId) {
+		Properties allProperties = new Properties();
+
+		if (propertyCache.containsKey(testIdentifierUniqueId)) {
```
Michael1993, line 165: You could use `getOrDefault` here.

src/main/java/org/junitpioneer/jupiter/report/PioneerReportListener.java
```diff
@@ -0,0 +1,213 @@
...
+		return testcase;
+	}
+
+	/**
+	 * Retrieves all cached properties for an {@code TestIdentifier} by its unique id.
+	 *
+	 * @param testIdentifierUniqueId Unique id the properties should be loaded for.
+	 * @return Wrapper which contains all cached properties or an empty list.
+	 */
+	Properties getProperties(String testIdentifierUniqueId) {
```
Michael1993, line 162: Maybe this could be an `Optional<Properties>` to indicate that you might _not_ always get back `Properties`?
E.g.:
```
Optional<Properties> getProperties(String id) {
    if (!propertyCache.containsKey(id)) {
        return Optional.empty();
    }
    Properties allProps = new Properties();
    allProps.getPropertyList().addAll(propertyCache.get(id));
    return Optional.of(allProps);
}
```

src/main/java/org/junitpioneer/jupiter/report/PioneerReportListener.java
Michael1993: If you take the alternative approach with `Optional` (see below), this could look like this:
```
if (allProps.isPresent()) {
    testcase.setProperties(allProps.get(());
}
```

Michael1993: This looks a bit better to me because you don't break Law of Demeter with it?

src/main/java/org/junitpioneer/jupiter/report/Testcase.java
```diff
@@ -0,0 +1,131 @@
...
+ *   &lt;/complexContent&gt;
+ * &lt;/complexType&gt;
+ * </pre>
+ *
+ *
+ */
+@XmlAccessorType(XmlAccessType.FIELD)
+@XmlType(name = "", propOrder = { "properties" })
+@XmlRootElement(name = "testcase")
+public class Testcase {
```
Michael1993, line 43: Shouldn't this be `TestCase`?

src/main/java/org/junitpioneer/jupiter/report/Testcontainer.java
```diff
@@ -0,0 +1,165 @@
...
+ *   &lt;/complexContent&gt;
+ * &lt;/complexType&gt;
+ * </pre>
+ *
+ *
+ */
+@XmlAccessorType(XmlAccessType.FIELD)
+@XmlType(name = "", propOrder = { "properties", "testcontainer", "testcase" })
+@XmlRootElement(name = "testcontainer")
+public class Testcontainer {
```
Michael1993, line 47: Shouldn't this be `TestContainer`?

docs/issue.adoc
```diff
@@ -0,0 +1,31 @@
+:page-title: @Issue
+:page-description: JUnit Jupiter extensions to add information about a related issue
+
+== Introduction
+
+The main reason tests are written to ensure the functionality of a requirement of to avoid technical problems.
```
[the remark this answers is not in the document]
Michael1993: Also, I'm not quite sure what you are trying to say here. Requirements don't have functionality, I think? Maybe you meant feature?

src/main/java/org/junitpioneer/jupiter/report/Property.java
```diff
@@ -0,0 +1,103 @@
...
+ *   &lt;/complexContent&gt;
+ * &lt;/complexType&gt;
+ * </pre>
+ *
+ *
+ */
+@XmlAccessorType(XmlAccessType.FIELD)
+@XmlType(name = "")
+@XmlRootElement(name = "property")
+public class Property {
```
Michael1993, line 40: Do you need to have the empty constructor and the setters even if you only write to (and never read from) XML?

src/main/java/org/junitpioneer/jupiter/IssueExtension.java
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * This class implements the functionality for the {@code @Issue} annotation.
+ *
+ * @see Issue
+ */
+@DisplayName("Issue extension")
+class IssueExtension implements BeforeEachCallback {
+
+	static final Namespace NAMESPACE = Namespace.create(IssueExtension.class);
```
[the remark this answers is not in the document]
Bukama: Leftover from the first try to implement the extension. Removed after next push.

src/main/java/org/junitpioneer/jupiter/IssueExtension.java
```diff
@@ -0,0 +1,61 @@
...
+		if (PioneerAnnotationUtils.isAnyAnnotationPresent(context, Issue.class)) {
+			storeIssueId(context);
+		}
+	}
+
+	/**
+	 * Reads the {@code @Issue} value from the annotation and publishes it in the extension context.
+	 * @param context The Extensions context
+	 */
+	void storeIssueId(ExtensionContext context) {
```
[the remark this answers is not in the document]
Bukama: Name was also a leftover - good spot!
Yeah with the new implementation a separate method only for publishing is overkill and doesn't improve readability (like it was before). So I removed the `storeIssueId` method. I want to keep the method to read the value from the annotation, because for me the code is more readable when its in an own method.

src/main/java/org/junitpioneer/jupiter/report/PioneerReportListener.java
Bukama: Thank you. Another point that streams are strange (the stream skips - whyever -  the EntrySet. *shrug* Updated the code.

src/main/java/org/junitpioneer/jupiter/report/Properties.java
Bukama: I won't rename the (generated) field, because this would result in an invalid XML (without chaning the XSD).

> Ungültiger Content wurde beginnend mit Element 'propertyList' gefunden. '{property}' wird erwartet.

Translated to english ~

> Invalid content found starting with element 'propertyList' found. '{property}' expected.


And I don't want to rename the element in the XSD / report, but to keep it like this

```
<properties>
    <property name="Issue" value="bishue 135"/>
</properties>

File: src/main/java/org/junitpioneer/jupiter/report/PioneerReportListener.java (hunk 25 above)
Comment (Bukama) [reply src/main/java/org/junitpioneer/jupiter/report/PioneerReportListener.java:149]: > Is this check necessary? What would happen if you just set the properties without this check?

Yes it is, because otherwise there would be a non null `Properties` element in the `Testcase` / `Testcontainer` class, which results in empty `<properties> elements for each object in the report, which I want to avoid.

```
<testcase name="an empty list, if no properties are stored" status="SUCCESSFUL">
    <properties/>
</testcase>

Bukama: > If you take the alternative approach with `Optional` (see below), this could look like this:

`Optional` is not serializeable. Even if , I would expect that it results in an empty `properties` element in the report.

src/main/java/org/junitpioneer/jupiter/report/Testcase.java
Bukama: > Shouldn't this be `TestCase`?

Looks wired in the report with the camel case elements but after looking the term up inside the ISTQB glossary `test case` is the right way to write it, so I renamed it (and `TestContainer` too).

src/main/java/org/junitpioneer/jupiter/report/Property.java
Bukama: As far as I know (and you seem to confirm it) the default constructors are needed when unmarshalling an XML file. As of now I don't do it (but I'm thinking about writing an additional test to check the `writeReport` method, where I would need it. Aside from that the class is generated and I want to keep them for future use.

docs/issue.adoc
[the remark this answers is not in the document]
Bukama: > Also, I'm not quite sure what you are trying to say here. Requirements don't have functionality, I think? Maybe you meant feature?

No I mean requirements. Functional requirements describe all professional functionality of an application, while non-functional requirements describe addional needs, like throughput, respone times, accessibility and so on. Requirements are definied by the customer (based on their use cases) and are the core of everything.

Features are, at least for me, addional functionalities which makes life easier or more compfortable.  Let's take the `forEach` loop as an example:

Example 1:

> Customer: I need an option to loop through a list. (Requirement)
> Engineer: Here you have a `while` loop. (Implemented requirement)

Requirement fullfilled, right? So let's provide a feature:

> Engineer: Because we are kind / you cry so much about the usability of the `while` loop, we added the `forEach` loop to make your like easier (Feature)

Example 2:

The customer definied the requirement more precice:

> Customer: I need an option to loop through a list of objects as long as there are elements and I want to have an object of the current iterated element so I can work with it directly, without accessing helper variables to access the object in e.g. an Array. (Requirement)
> Engineer: Here you have a `forEach` loop. (Implemented requirement)

Feature here? Hmm, well maybe `Stream`-API.

In my opinion the term `feature` is mostly used in fields of tools, because many people don't think of a customer relationship there.

This said the core of writing a test is to ensure a requirement or to avoid bugs, like `NullPointerException` or `IndexOutOfBoundException`.

src/main/java/org/junitpioneer/jupiter/report/PioneerReportListener.java
Michael1993: `Optional<Properties>` returned by `getProperties(String id)` does not leak into serialization because you unwrap it with `get()`.

I'm advocating for `Optional` because that would communicate clearly that `getProperties` (unlike a normal getter) might _or might not_ return with a value. It also makes the code a bit shorter.

src/main/java/org/junitpioneer/jupiter/report/PioneerReportListener.java
```diff
@@ -0,0 +1,203 @@
...
+		String uniqueId = container.getUniqueId();
+		Properties allProps = getProperties(uniqueId);
+
+		if (!allProps.getProperty().isEmpty()) {
+			testcontainer.properties = allProps;
+		}
+
+		Set<TestIdentifier> allChildren = storedTestPlan.getChildren(uniqueId);
+
+		for (TestIdentifier identifier : allChildren) {
```
Michael1993, line 115: There is a bit of duplication here - Can you write this method in a way that `testPlanExecutionFinished `can just call this instead of a for-each loop? Or extract a method you can call from both `testPlanExecutionFinished `and `processContainer`?

src/main/java/org/junitpioneer/jupiter/report/PioneerReportListener.java
Bukama: I understand what you mean, but I'm not sure as I understand `Optional` is to be used to avoid NULL. And I never return NULL here.

src/main/java/org/junitpioneer/jupiter/report/PioneerReportListener.java
Bukama: I'm not sure as the result is stored in different variables. When I get time (and motivation :angel: ) I'll have a look at time

docs/issue.adoc
```diff
@@ -0,0 +1,31 @@
+:page-title: @Issue
+:page-description: JUnit Jupiter extensions to add information about a related issue
+
+== Introduction
+
+The main reason tests are written is to ensure the functionality of a requirement or to avoid technical problems.
+The `@Issue` allows to mark tests with the related issue, like a requirement or a bugfix, of an issue-tracker like JIRA or Redmine.
```
Michael1993, line 7: What do you think of this:
"The `@Issue` annotation allows marking tests with a String, referencing a related issue (like a requirement or a bugfix) of an issue-tracker (like JIRA or Redmine)."

docs/issue.adoc
```diff
@@ -0,0 +1,31 @@
...
+
+It's a JUnit Jupiter extension to show that tests exist to cover an specific issue.
```
Michael1993, line 9: I'd use "for showing" instead of "to show", I think.

src/main/java/org/junitpioneer/jupiter/report/PioneerReportListener.java
Bukama: Don't know how I should do it. After further investigation I don't think this is real duplication.

src/main/java/org/junitpioneer/jupiter/report/Property.java
Bukama: It's needed, otherwise `Caused by: com.sun.xml.bind.v2.runtime.IllegalAnnotationsException` are thrown, which claim that the defaultconstructor do not exist.

Bukama:
Removed the comments in the generated classes, reducing the "duplicated code" to 4,x percent, which is still too much to pass the quality gate. So another hint for ignoring these classes.

src/main/java/org/junitpioneer/jupiter/report/PioneerReport.java
```diff
@@ -0,0 +1,65 @@
...
+import javax.xml.bind.annotation.XmlRootElement;
+import javax.xml.bind.annotation.XmlType;
+
+/**
+ * Root element of the Pioneer report.
+ */
+@XmlAccessorType(XmlAccessType.FIELD)
+@XmlType(name = "", propOrder = { "properties", "testContainer", "testCase" })
+@XmlRootElement(name = "pioneerReport")
+public class PioneerReport {
```
Michael1993, line 28: This class could just extend `TestContainer` and have an empty body - in which case you'd need to remove `propOrder`. This would solve the code duplication issue with Sonar.

src/main/java/org/junitpioneer/jupiter/report/PioneerReportListener.java
Michael1993: How about the following:
```
void processIdentifiers(TestContainer container, Set<TestIdentifier> identifiers, TestPlan plan) {
	for (TestIdentifier id : identifiers) {
		if (id.isContainer()) {
			final TestContainer testContainer = new TestContainer(id.getDisplayName());

			String uniqueId = id.getUniqueId();

			// Add properties, if there are any
			getProperties(uniqueId).ifPresent(testContainer::setProperties);

			processIdentifiers(testContainer, plan.getChildren(uniqueId), plan);
			container.getTestContainers().add(testContainer);
		} else if (id.isTest()) {
			container.getTestCases().add(processTest(id));
		}
	}
}
```

Then instead of the original `for` loop, you could write:
```
processIdentifiers(pioneerReport, testPlan.getRoots(), testPlan);
```
This has the added benefit that you'd no longer need to "store" the `testPlan` in a field.

src/main/java/org/junitpioneer/jupiter/report/PioneerReportListener.java
```diff
@@ -0,0 +1,199 @@
...
+	 *
+	 * @param container Object to be converted into {@code Testcontainer}
+	 * @return Created object
+	 */
+	TestContainer processContainer(TestIdentifier container) {
+		TestContainer testcontainer = new TestContainer(container.getDisplayName());
+
+		// Add properties, if there are any
+		String uniqueId = container.getUniqueId();
+		Properties allProps = getProperties(uniqueId);
```
Michael1993, line 103: If `getProperties` returned with `Optional<Properties>`, this whole thing could look like this:
```
String uniqueId = container.getUniqueId();
// Add properties, if there are any present
getProperties(uniqueId).ifPresent(testContainer::setProperties);
```

src/main/java/org/junitpioneer/jupiter/report/Properties.java
Michael1993: Just rename the variable and modify the `propOrder`. You can still satisfy the XSD/Report with explicit naming i.e.: `@XmlElement(required = true, name = "property")`

src/main/java/org/junitpioneer/jupiter/report/TestContainer.java
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * Represents a test container in the Pioneer report.
+ */
+@XmlAccessorType(XmlAccessType.FIELD)
+@XmlType(name = "", propOrder = { "properties", "testContainer", "testCase" })
+@XmlRootElement(name = "testContainer")
+public class TestContainer {
+
+	protected Properties properties;
+	protected List<TestContainer> testContainer; //NOSONAR renaming would produce invalid XML
```
Michael1993, line 31: You could just rename this field, modify `propOrder` and have explicit naming with `@XmlElement(name = "testContainer")`.

src/main/java/org/junitpioneer/jupiter/report/TestContainer.java
```diff
@@ -0,0 +1,71 @@
...
+	protected List<TestCase> testCase;
+	@XmlAttribute(name = "name")
+	protected String name;
+
+	public TestContainer() {
+		// Needed for marshalling
+	}
+
+	public TestContainer(String name) {
```
Michael1993, line 40: I'd rewrite this to be:
```
public TestContainer() {
	this.testContainers = new ArrayList<>();
	this.testCases = new ArrayList<>();
}

public TestContainer(String name) {
	this();
	this.name = name;
}
```

src/main/java/org/junitpioneer/jupiter/report/PioneerReportListener.java
```diff
@@ -0,0 +1,199 @@
...
+			File xsdFile = new File(ClassLoader.getSystemResource("xsd/pioneerreport.xsd").toURI());
+			Schema schema = schemaFactory.newSchema(xsdFile);
+			Validator validator = schema.newValidator();
+			validator.setProperty(XMLConstants.ACCESS_EXTERNAL_DTD, "");
+			validator.setProperty(XMLConstants.ACCESS_EXTERNAL_SCHEMA, "");
+			validator.validate(xmlFileSource);
+		}
+		catch (Throwable t) { //NOSONAR
+			// Throw PioneerReportException regardless of checked or unchecked exceptions during execution
+			throw new PioneerReportException("Error while creating the pioneer report", t); //NOSONAR
```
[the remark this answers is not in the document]
Bukama: It's bad style to throw (or catch) generic exceptions like `RuntimeException`, `Exception` or `Throwable`.

Bukama:
I merged master today and now the test runs forever until I terminate them...

Bukama:
After feeling not fine with this PR for a long time, I turned it into a draft again to exclude the general report and provide some context-related thing for the issues.

Bukama:
Most important feedback for me is the addition of the two java libs for unmarshaling and validating. If this is a no-go, feel free to close the draft and I will made it a private extension on my own (as I need the feature). Want to set up a maven plugin for working with the report after it.

Michael1993 (changes requested):
Please fix the code smell, it's very minor.

Bukama:
I again turned this into a draft, because @Michael1993 brought up a discussion about the focus of the PR. Before I get to the arguments I want to give a short view at what I aim for, let's call it "my personal requirements".

In the long run I want to have a report, which I can pass e.g. to the QA-lead to show which issues/requirements are covered by tests (and their results) and which are not. So I need the possibility to

* label tests with a certain issue / requirement
* log the results of the test
* pass a list of issues / requirements for comparison
* generate the final report

When I implemented the current state of the extension I clearly had in mind that the last to steps (taking a list of issues and generate the final report) is not task of the pioneer xp (but of some private/others lib) and I wanted to give people the possibility to do their own things with the result of the first steps (which everyone has to do). As published `ReportEntries` don't show up anyway where you can "pick them up", I decided to create a (schema based) report as some kind of interface contract, where people can progress on. In my case take this report, combine it with a list of issues and generate an report (didn't decide yet, most probably XML in the first step and then use XSLT to transform it to HTML or PDF and link the issues to a concrete issue tool instance).

I knew from the start on that creating the report would add new dependencies which at its core is something we don't want to do. Sure I could create the XML by hand (like the JUnit legacy reporter does) but than I still have to validate it against a schema, so I need the dependency anyway.

To come back to the start: @Michael1993 suggest to slice the current implementation to "just" provide some kind of "abstract extension" and don't create the report, but give people the possibility to work with the collected results of the labeled tests. Or maybe even not this, but only provide the annotation. Aside from that he suggested to also provide a sample implementation of one (read: my) concrete usage / needs in another repository beneath the pioneer organization.

I like the idea of writing a somehow "abstract" extension to avoid the dependency, but I can't (yet) imagine where/ how to slice. I mean just providing an extension which does nothing (or which only publishes `ReportEntries`) is easy. But as far as I read the two issues in the JUnit5 repository and our repository everyone needs / wants to know the outcome of `@Issue` labeled tests. So I see the need that the extensions already collects them (in my draft with the listener) and doesn't force everyone to implement that on their own. But I 'm somehow stuck how to "publish" those gathered information than.

When I cut this mental knot, I'll work again on this and change the current implementation.

Michael1993:
The way I imagine the extension would look like:
The `IssueExtension` has 3 parts.

The first one is the annotation, `@Issue` which takes a `String` value. This references the issue the test is linked to, basically a unique ID.

Second is a `TestExecutionListener` that gathers the information about the tests annotated (or meta-annotated!) with `@Issue`. It stores this information. Maybe we could re-use JUnit's [`Events`](https://junit.org/junit5/docs/5.5.0-M1/api/org/junit/platform/testkit/engine/Events.html) class and store it in a `Map<String, Events>` where the key is the `value()` of the annotation. Alternatively, it could be a `Map<String, Map<String, Events>>` if we want to store an `Events` object for each individual test but also group them based on the issue.

The third part would be an abstract class/interface that whoever uses the extension has to implement themselves. We could load this in with a `ServiceLoader` or let the user define their own annotations:
```java
@ExtendWith(ConcreteImplementationExtension.class)
@Issue("GitHub")
public @interface GitHub {
    String value();
}
```
Best option is to do both, I think.

I am not entirely certain about this abstract class - maybe it could implement a default generic comparison logic but how would that even look like? I'm almost certain people would need to overwrite it anyways to better fit their needs.

After we implement these three parts, we could create a separate project under the Pioneer umbrella that showcases examples that implement/extend this abstract thing.

Bukama:
As I don't expect that this solutions gets any approves I did not spent much time to investigate how to use JAXB on Java 11+.

nipafx:
As documented in the README, we prefer to keep our run-time dependencies to an absolute minimum, which for now is zero. I don't like the idea of adding JAXB "just" for a single extension. After discussing this, you opened #361, so we can close this one.

Issue #164, opened 2020-01-21
Require configuration failiure tests?


nicolaiparlog:
Some extensions have tests that check whether broken configurations are handled as expected (e.g. `DefaultLocaleTests`). Should all extensions do that? (E.g. `DefaulTimeZoneTests`, which doesn't at the moment.)

Bukama:
I my opinion it's important that a functioniality checks all of its parts. That's why I tried to check the annotation when I was trying to create the `@Issue` annotation. So I vote for _Yes they should do it_

Bukama:
As nobody argues against this I remove the `in-discussion` label and set `up for grabs`.

Bukama:
Note: Aside from the opened PR in #272 I would like to create a list of all extension which lack of such tests, so we have an overview before starting (and maybe doing double work).

As I don't have much time at the moment, feel free to have a look at the code and add such a list here - thank you.

Bukama:
If I have not overseen anything this should be the extensions which as of today (don't) check if the configuration is correct:

Checks config:
* RangeSourceArgumentsProvider
* DefaultLocaleExtension
* EnvironmentVariableExtension
* IssueExtension
* RepeatFailedTestExtension
* ReportEntryExtension
* StdIOExtension
* SystemPropertyExtension
* TempDirectoryExtension

Does not check / do they need to?

* DefaultTimeZoneExtension
* TimeoutExtension


@Michael1993 , @beatngu13 @NPException Did I oversee anything?

Bukama:
> Does not check / do they need to?
> 
> * DefaultTimeZoneExtension
> * TimeoutExtension

For the `DefaultTimeZoneExtension´ @Michael1993 has opened PR #272 .

The other is still up for grabs.

nicolaiparlog:
Closed by #272 and #280.

PR #272, opened 2020-06-04
Added ExtensionConfigurationException to DefaultTimeZoneExtension

Michael1993:
Also includes a small documentation fix.
Progresses #164 - which extensions should also be updated this way?

---

### DIFF
--- a/src/main/java/org/junitpioneer/jupiter/DefaultTimeZoneExtension.java
+++ b/src/main/java/org/junitpioneer/jupiter/DefaultTimeZoneExtension.java
@@ -43,13 +43,21 @@ class DefaultTimeZoneExtension implements BeforeAllCallback, BeforeEachCallback,
 
 	private void setDefaultTimeZone(Store store, DefaultTimeZone annotation) {
 		storeDefaultTimeZone(store);
-		TimeZone configuredTimeZone = TimeZone.getTimeZone(annotation.value());
+		String timeZoneId = annotation.value();
+		TimeZone.setDefault(getTimeZone(timeZoneId));
+	}
+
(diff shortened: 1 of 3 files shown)


src/main/java/org/junitpioneer/jupiter/DefaultTimeZoneExtension.java
```diff
@@ -43,6 +44,10 @@ public void beforeEach(ExtensionContext context) {
 	private void setDefaultTimeZone(Store store, DefaultTimeZone annotation) {
 		storeDefaultTimeZone(store);
 		TimeZone configuredTimeZone = TimeZone.getTimeZone(annotation.value());
+		if (configuredTimeZone.equals(TimeZone.getTimeZone("GMT")) && !annotation.value().equals("GMT")) {
```
Bukama, line 47: I ask you to to add a code comment here (and in the documentation (`.adoc` file)) , because I struggled some time to understand this line. It got clear, when I read the Javadoc of `TimeZone.getTimeZone(...)` which points out

> returns the specified TimeZone, or the GMT zone if the given ID cannot be understood.

src/main/java/org/junitpioneer/jupiter/RepeatFailedTest.java
```diff
@@ -62,7 +62,7 @@
 public @interface RepeatFailedTest {
 
 	/**
-	 * Specifies how often the test is executed at most.
+	 * Specifies how many times the test is executed at most.
```
Bukama, line 65: Please revert this change as it has nothing to do with the issue/PR (and will most probably result in merge conflicts of #236)

PR #273, opened 2020-06-09
Reinstate rule that requires issue and PR number in title

nicolaiparlog:
Proposed commit message:

```
Require issue and PR number in commit title (#264 / #273)

This was the rule until #79 / #86 changed it. We now consider that a
bad decision and want to start doing it again. An important factor
was that IDEs can make issue and PR numbers clickable links, which
adds additional value to making them easily accessible.
```

---

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -212,7 +212,7 @@ To make the single commit expressive, its message must be detailed and [good]((h
 Furthermore, it must follow this structure:
 
 ```
-${action}
+${action} (${issues} / ${pull-request})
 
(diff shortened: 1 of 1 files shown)


nicolaiparlog:
As soon as the first other maintainer gives thumbs up, the other one can merge this. :)

aepfli:
actually i do like the refernce block at the bottom. not particularly for closes/fixes - but it could also be used for referencing other topics to give more context. Also the manual part @Bukama is talking about, is something we should verify, if this is also the case for ticket numbers in the title

PR #275, opened 2020-06-19
Revamp/improvement of StdIoExtension

### LINKED ISSUE #274: Improvements to StdIOExtension
I had a look at the `StdIOExtension` and I think it's pretty neat. Some details that could be improved:

# Class Name

We never settled this anywhere, but I consider abbreviations as a single word, when it comes to capitalization, so I'd rename `StdIOExtension` to `StdIoExtension` (lower case "o"). For comparison: `DVDReader` vs `DvdReader` and `ASMRMic` vs `AsmrMic`.

# Reset standard reader/writer

The extension should reset the `System.in` and `System.out` after each test.

# `@ExtendWith`

I consider `@ExtendWith` to be a low-level annotation that I'd rather avoid for users. Why not apply it to the already existing `@StdInSource` and then create a new `@StdOut` annotation.

# Unexpected behavior

The documentation says:

> Some readers read all lines from `@StdInSource` eagerly (e.g.: `BufferedReader`) which can lead to unexpected behavior.

Can you go into a little more detail on that and provide an example?

# Small things

There are also a few small things that I already edited. I will put them on a branch.

Michael1993:
Closes #274

---


src/main/java/org/junitpioneer/jupiter/StdIoExtension.java
```diff
@@ -61,11 +90,34 @@ private Object getOut() {
...
+			if (!params.contains(StdIn.class) && !params.contains(StdOut.class)) {
+				throw new ExtensionConfigurationException(
+					format("Method is annotated with %s but no %s or %s parameters were found.", StdIo.class.getName(),
+						StdIn.class.getName(), StdOut.class.getName()));
+			}
+			if (!String.join("", method.get().getAnnotation(StdIo.class).value()).isEmpty()
+					&& !params.contains(StdIn.class)) {
+				throw new ExtensionConfigurationException(format(
+					"Method has no %s parameter but input sources were provided in the %s annotation (Did you forget to add a test parameter?).",
+					StdIn.class.getName(), StdIo.class.getName()));
```
Bukama, line 118: I don't get this. The if tests if the value array is *not* empty and if the parameter does *not* contain `StdIn`, but the error messages says that there are parameter but `StdIn` was * not* provided. Which is totally fine, if `StdOut` was provided and the values are the expected written values.

Did you miss to add the check for `StdOut`, because you want to check if there is a) an annotation with values but b) neither `StdIn` or `StdOut` parameter?

src/test/java/org/junitpioneer/jupiter/StdIoExtensionTests.java
```diff
@@ -10,84 +10,115 @@
...
-	@ExtendWith(StdIOExtension.class)
-	class ExtendWithTests {
+	@DisplayName("with standard configuration ")
+	class StdIoTests {
 
 		@Test
 		@DisplayName("catches the output on the standard out as lines")
+		@StdIo
 		void catchesOut(StdOut out) {
 			app.write();
```
Bukama, line 124: The method gets no values provided, so the expected result is that `out.capturedLines()` is empty.

Where do the expected lines come from? (I know - from the code itself, but you get what I mean?)

src/test/java/org/junitpioneer/jupiter/StdIoExtensionTests.java
```diff
@@ -122,10 +153,55 @@ void catchesBoth(StdIn in, StdOut out) throws IOException {
 
 	}
 
-	@ExtendWith(StdIOExtension.class)
-	static class StdIOExtensionConfigurations {
+	@Nested
+	@TestMethodOrder(OrderAnnotation.class)
```
Bukama, line 157: Just a note:

This nested class should be thread safe due the use of `@TestMethodOrder` regarding to [JUnit 5 User Guide](https://junit.org/junit5/docs/current/user-guide/#writing-tests-parallel-execution)

> The default execution mode is applied to all nodes of the test tree with a few notable exceptions, namely test classes that use the Lifecycle.PER_CLASS mode or a MethodOrderer (except for Random).

src/main/java/org/junitpioneer/jupiter/StdIo.java
```diff
@@ -0,0 +1,44 @@
...
+ * Marks a method to be extended with {@link StdIoExtension} and provided with
+ * {@link StdIoExtension.StdIn StdIn} or {@link StdIoExtension.StdOut StdOut}.
+ * The annotated method must take one or both parameters or an
+ * {@link org.junit.jupiter.api.extension.ExtensionConfigurationException} will be thrown.
+ *
+ * Provide values that the {@link StdIoExtension extension} will read instead of reading the
+ * standard input ({@code System.in}).
+ * If values are provided but there is no {@link StdIoExtension.StdIn} parameter
+ * an {@link org.junit.jupiter.api.extension.ExtensionConfigurationException} will be thrown.
+ * The opposite is not true to enable testing for empty inputs.
```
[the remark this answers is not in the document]
Michael1993: Your last statement is incorrect. This is only for `StdIn`.

src/main/java/org/junitpioneer/jupiter/StdIo.java
```diff
@@ -0,0 +1,44 @@
...
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.METHOD)
+@ExtendWith(StdIoExtension.class)
+public @interface StdIo {
+
+	/**
+	 * Provides the intercepted standard input with values.
+	 * If this is not blank, the annotated method has to
+	 * have a {@link StdIoExtension.StdIn} parameter.
+	 */
```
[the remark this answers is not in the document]
Michael1993: No, that's not correct.

src/main/java/org/junitpioneer/jupiter/StdIoExtension.java
Michael1993: If the array is not empty, then you provided values for `StdIn` - which is incorrect if you don't have and `StdIn` parameter.

src/main/java/org/junitpioneer/jupiter/StdIoExtension.java
```diff
@@ -20,21 +20,38 @@
...
 
+	private static final Namespace NAMESPACE = Namespace.create(DefaultLocaleExtension.class);
+
+	private static final String IN_KEY = "StdIo_In";
+	private static final String OUT_KEY = "StdIo_Out";
+
+	@Override
+	public void afterEach(ExtensionContext context) {
+		System.setIn(context.getStore(NAMESPACE).get(IN_KEY, InputStream.class)); //NOSONAR resetting input
+		System.setOut(context.getStore(NAMESPACE).get(OUT_KEY, PrintStream.class));
```
NPException, line 52: Could this break if only one of `StdIn` or `StdOut` is used?
`System.in` or `System.out` may end up being `null` if one of these `.get(...)` calls returns null.

src/main/java/org/junitpioneer/jupiter/StdIoExtension.java
```diff
@@ -20,21 +20,38 @@
...
  * <p>For details and examples, see
  * <a href="https://junit-pioneer.org/docs/standard-input-output/" target="_top">the documentation on <code>Standard input/output</code></a>
  * </p>
  */
-public class StdIOExtension implements ParameterResolver {
+public class StdIoExtension implements ParameterResolver, BeforeTestExecutionCallback, AfterEachCallback {
 
 	private static final String SEPARATOR = System.getProperty("line.separator");
 
+	private static final Namespace NAMESPACE = Namespace.create(DefaultLocaleExtension.class);
```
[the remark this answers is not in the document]
Michael1993: Oops. :sweat_smile:

src/main/java/org/junitpioneer/jupiter/StdIoExtension.java
Michael1993: Yeah, should probably check if it even needs to be reset first.

src/test/java/org/junitpioneer/jupiter/StdIoExtensionTests.java
Michael1993: You don't provide values to `System.out` - you want to test that your SUT writes to `System.out` what you expect it to.

docs/standard-input-output.adoc
```diff
@@ -1,59 +1,113 @@
...
+To use a parameter, you have to use the annotation and if you use the annotation, you have to use a parameter.
 The extension provides the following two parameters:
 
-StdIn(`org.junitpioneer.jupiter.StdIOExtension.StdIn`)::
+StdIn(`org.junitpioneer.jupiter.StdIoExtension.StdIn`)::
 This is for simulating input from the standard in.
-This parameter also requires that the method is annotated with `@StdInSource`.
-The `@StdInSource` annotation takes a String array which holds the values that the test will "read" instead of waiting for input on the standard input.
+The `StdIo` annotation takes a String array which holds the values that the test will "read" instead of waiting for input on the standard input.
+This is not required - you can simulate an empty input if you don't supply your own values.
```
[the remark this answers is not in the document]
Michael1993: I changed the default value and added a test so this behaviour is more apparent but the documentation is correct here.

docs/standard-input-output.adoc
```diff
@@ -1,59 +1,113 @@
...
+
+        consoleReader.readLines();
+
+        assertEquals(in.capturedLines(), "line1", "line2");
+    }
+
+}
+----
+
+This fails!
```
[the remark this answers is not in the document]
Michael1993: Yeah, alright, fair enough. Changed.

src/main/java/org/junitpioneer/jupiter/StdIoExtension.java
Michael1993: I added more tests to cover this case and changed the behaviour.

src/test/java/org/junitpioneer/jupiter/StdIoExtensionTests.java
Michael1993: The expected lines come from the `app.write()` method. In it, there are calls to `System.out.println()` - which are redirected to StdOut by the extension.

nicolaiparlog:
I think this is ready to be merged. Would love for somebody else to take a look, though.

src/main/java/org/junitpioneer/jupiter/StdIoExtension.java
```diff
@@ -0,0 +1,117 @@
...
+	private void swapAndStoreIn(ExtensionContext context, StdIn stdIn) {
+		context.getStore(NAMESPACE).put(SYSTEM_IN_KEY, System.in); //NOSONAR never reading from System.in, only storing it
+		context.getStore(NAMESPACE).put(STD_IN_KEY, stdIn);
+		System.setIn(stdIn); //NOSONAR required to redirect output
+	}
+
+	@Override
+	public void beforeTestExecution(ExtensionContext context) {
+		final Method method = context.getRequiredTestMethod();
+		if (!method.isAnnotationPresent(StdIo.class))
```
Michael1993, line 98: The `@StdIo` annotation _must_ be present. The StdIoExtension itself is not public and can only be registered with the annotation.

We could delete this condition and exception because this never happens and could be confusing.

src/main/java/org/junitpioneer/jupiter/StdIoExtension.java
```diff
@@ -0,0 +1,117 @@
...
+				//@formatter:off
+				return extensionContext
+						.getStore(NAMESPACE)
+						.getOrComputeIfAbsent(
+								STD_IN_KEY,
+								__ -> createSwapStoreStdIn(extensionContext, source),
+								StdIn.class);
+				//@formatter:on
+		}
+		throw new ParameterResolutionException(format("Could not resolve parameter of type %s.", parameterType));
```
Michael1993, line 65: This exception is never thrown. Could be deleted.
From the [ParameterResolver::resolveParameter documentation](https://junit.org/junit5/docs/5.0.1/api/org/junit/jupiter/api/extension/ParameterResolver.html#resolveParameter-org.junit.jupiter.api.extension.ParameterContext-org.junit.jupiter.api.extension.ExtensionContext-):
> This method is only called by the framework if supportsParameter(org.junit.jupiter.api.extension.ParameterContext, org.junit.jupiter.api.extension.ExtensionContext) previously returned true for the same ParameterContext and ExtensionContext.

nicolaiparlog: Without this line, the compiler complains because the method does not return/throw on all paths. I prefer an exception over `return null`.

src/main/java/org/junitpioneer/jupiter/StdIoExtension.java
nicolaiparlog: I switched annotation detection to our utils (so meta-annotations are found as well), which forces us to consider the absent case because it returns an `Optional`.

Michael1993:
LGTM

PR #280, opened 2020-06-20
Fix @Test timeout to actually time out

nicolaiparlog:
---

### DIFF
--- a/docs/vintage-test.adoc
+++ b/docs/vintage-test.adoc
@@ -1,7 +1,7 @@
 :page-title: Vintage @Test
 :page-description: A drop-in replacement for JUnit 4's @Test annotation, including expected and timeout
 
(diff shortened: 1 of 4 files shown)


nicolaiparlog:
Proposed commit message:

```
Let vintage @Test timeout actually time out (#10 / #280)

With the update to JUnit 5.5, we can now use the
`InvocationInterceptor` extension point to push test execution onto
a separate thread that we can abandon when the specified timeout runs
out. Somewhat surprisingly, using Jupiter's
`Assertions::assertTimeoutPreemptively` worked really well for that.

The extension now also throws `ExtensionConfigurationException` if
the specified timeout is negative.

Closes: #10 
References: #164
PR: #280 
```

nicolaiparlog:
Because of the added tests for negative timeouts, this closes #164 as well.

PR #285, opened 2020-06-21
Provide additional PioneerTestKit methods to create ExecutionResults (283)

### LINKED ISSUE #283: Refactor `PioneerTestKit::executeTestMethodWithParameterTypes` to accept `Class<?>[]`
The `PioneerTestKit` allows creation of `ExecutionResults` for test methods with parameters. At the moment, `executeTestMethodWithParameterTypes` takes the parameter types as a single, concatenated string. This is painful to use. Can't we just refactor to provide a `Class<?>[]` instead?

Bukama:
```
Provide additional PioneerTestKit methods to create ExecutionResults (#283 / #285)

Refactored one method of the PioneerTestKit to create ExecutionResults,
allowing to pass classes instead of strings containing the fully
qualified class name.
Tests for the PioneerTestKit were also added.

Closes #283
PR: #285
```

---


src/test/java/org/junitpioneer/testkit/ExecutionResults.java
```diff
@@ -55,6 +57,20 @@
 				.execute();
 	}
 
+	ExecutionResults(Class<?> testClass, String testMethodName, Class[] methodParameterTypes) {
+
+		String allTypNames = Arrays
```
Michael1993, line 62: Is this a typo? `allTypNames` instead of `allTypeNames`

src/test/java/org/junitpioneer/testkit/ExecutionResults.java
```diff
@@ -55,6 +57,20 @@
 				.execute();
 	}
 
+	ExecutionResults(Class<?> testClass, String testMethodName, Class[] methodParameterTypes) {
```
Michael1993, line 60: This constructor seems unnecessary - you could just join the canonical class names in the factory method and call the constructor that takes a `String`?

src/test/java/org/junitpioneer/testkit/PioneerTestKit.java
```diff
@@ -46,4 +46,30 @@ public static ExecutionResults executeTestMethodWithParameterTypes(Class<?> test
...
 
+	/**
+	 * Returns the execution results of the given method of a given test class.
+	 *
+	 * @param testClass Name of the test class
+	 * @param testMethodName Name of the test method (of the given class)
+	 * @param methodParameterTypes Array containing the types of for the parameters
+	 * @return The execution results
+	 */
+	public static ExecutionResults executeTestMethodWithParameterTypes(Class<?> testClass, String testMethodName,
```
Michael1993, line 57: These two methods should be merged by using the `Class<?>...` as the last parameter type. Which should throw an exception if no last parameter is passed (or maybe just call `executeTestMethod` as an attempt to "fix" the bad call?).

Alternatively, you could make it like this:
```
public static ExecutionResults executeTestMethodWithParameterTypes(Class<?> testClass, String testMethodName, Class<?> firstParameter, Class<?>... additionalParameters)
```
This ensures at least one class object is always passed to the method but is much more cumbersome. Method signature might be confusing too.

src/test/java/org/junitpioneer/testkit/ExecutionResults.java
Bukama: > This constructor seems unnecessary - you could just join the canonical class names in the factory method and call the constructor that takes a `String`?

Its explicit requested by @nicolaiparlog . I could not find a place where I could use it, but as he requested it he most probably has something/somewhere in mind to use it.

Bukama:
Did some refactoring as requested. Local build successful.

src/test/java/org/junitpioneer/testkit/PioneerTestKit.java
```diff
@@ -38,12 +41,21 @@ public static ExecutionResults executeTestMethod(Class<?> testClass, String test
 	 *
 	 * @param testClass Name of the test class
 	 * @param testMethodName Name of the test method (of the given class)
-	 * @param methodParameterTypes Full qualified types names of the parameters (e.g. "java.nio.file.Path")
+	 * @param methodParameterTypes Class type(s) of for the parameters
```
Michael1993, line 44: It's either `of` or `for` but never both.

src/test/java/org/junitpioneer/testkit/PioneerTestKit.java
```diff
@@ -38,12 +41,21 @@ public static ExecutionResults executeTestMethod(Class<?> testClass, String test
...
 	 * @return The execution results
+	 * @throws IllegalArgumentException when methodParameterTypes is null
```
Michael1993, line 46: But why only for that one parameter?

src/test/java/org/junitpioneer/testkit/PioneerTestKitTests.java
```diff
@@ -0,0 +1,105 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+public class PioneerTestKitTests {
```
Michael1993, line 22: But who tests our tests? (Is this necessary? I'm not sure...)

aepfli: well, good question actually - but i do think it is worth testing, we never know where we end up with this, and maybe people start using our PioneerTestKit in their extensions :D

Michael1993: It's in the test root - do we ship it? I guess the code is open-source but I don't think `PioneerTestKit` is part of our "artifact"... or is it?

src/test/java/org/junitpioneer/testkit/PioneerTestKit.java
```diff
@@ -38,12 +41,21 @@ public static ExecutionResults executeTestMethod(Class<?> testClass, String test
...
 	 */
 	public static ExecutionResults executeTestMethodWithParameterTypes(Class<?> testClass, String testMethodName,
-			String methodParameterTypes) {
-		return new ExecutionResults(testClass, testMethodName, methodParameterTypes);
+			Class<?>... methodParameterTypes) {
+
+		if (null == methodParameterTypes) {
```
Michael1993, line 51: Did you consider `Objects.requireNonNull(T value, String message)`? I think it conveys this intent really nicely.

src/test/java/org/junitpioneer/testkit/PioneerTestKitTests.java
aepfli: good question

src/test/java/org/junitpioneer/testkit/PioneerTestKit.java
```diff
@@ -38,12 +41,21 @@ public static ExecutionResults executeTestMethod(Class<?> testClass, String test
 	 *
 	 * @param testClass Name of the test class
 	 * @param testMethodName Name of the test method (of the given class)
-	 * @param methodParameterTypes Full qualified types names of the parameters (e.g. "java.nio.file.Path")
+	 * @param methodParameterTypes Class type(s) of for the parameters
 	 * @return The execution results
+	 * @throws IllegalArgumentException when methodParameterTypes is null
 	 */
 	public static ExecutionResults executeTestMethodWithParameterTypes(Class<?> testClass, String testMethodName,
```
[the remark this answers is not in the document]
Bukama: Yes because its replaceable by the generic one (and you suggested in your last review to do so)

src/test/java/org/junitpioneer/testkit/PioneerTestKit.java
Bukama: Because the other ones are handled by the Jupiter TestKit and I want to avoid NPE in Pioneers code (accessing the get name in the stream).

src/test/java/org/junitpioneer/testkit/PioneerTestKit.java
Bukama: Changed, thanks

src/test/java/org/junitpioneer/testkit/PioneerTestKit.java
Bukama: Yes, but it throws an `NullPointerException` and not an `IllegalArgumentException`. A NPE has a total different meaning than a IAE.

src/test/java/org/junitpioneer/testkit/PioneerTestKitTests.java
Bukama: > But who tests our tests? (Is this necessary? I'm not sure...)

Who watchs the watchers? JUnit Jupiter is tested by JUnit Jupiter.

> It's in the test root - do we ship it? I guess the code is open-source but I don't think `PioneerTestKit` is part of our "artifact"... or is it?

Content in `src/test` is not shiped.

src/test/java/org/junitpioneer/testkit/PioneerTestKitTests.java
```diff
@@ -74,8 +74,9 @@ void executeTestMethodWithParameterTypesSingleParameterTypeIllegalArgumentExcept
 				});
 
 				String expectedMessage = "methodParameterTypes must not be null";
-				assertThat(thrown).isInstanceOf(IllegalArgumentException.class);
-				assertThat(thrown.getMessage()).isEqualTo(expectedMessage);
+
+				org.assertj.core.api.Assertions.assertThat(thrown).isInstanceOf(IllegalArgumentException.class);
+				org.assertj.core.api.Assertions.assertThat(thrown.getMessage()).isEqualTo(expectedMessage);
```
Michael1993, line 79: This could just be `Assertions.assertThat` if you do `import org.assertj.core.api.Assertions`

Bukama:
Updated some docs to persist the discussed things (e.g. about the `IllegalArgumentException`) in the code.

Bukama:
@aepfli *tapp tapp on your shoulder* 👼

src/test/java/org/junitpioneer/testkit/PioneerTestKit.java
Michael1993: I think a `NullPointerException` conveys that a certain value can not (or should not) be `null`.

Bukama: > I think a `NullPointerException` conveys that a certain value can not (or should not) be `null`.

No it indicates that a certan value *is* NULL, according to [documentation of NullPointerException](https://docs.oracle.com/javase/7/docs/api/java/lang/NullPointerException.html) (highlight by me):

> Thrown when an application **attempts to use null in a case where an object is required**. These include:
> 
> - Calling the instance method of a null object.
> - Accessing or modifying the field of a null object.
> - Taking the length of null as if it were an array.
> - Accessing or modifying the slots of null as if it were an array.
> - Throwing null as if it were a Throwable value.
> 
> Applications should throw instances of this class to indicate other illegal uses of the null object. NullPointerException objects may be constructed by the virtual machine as if suppression were disabled and/or the stack trace was not writable.

But the [documentation of IllegalArgumentExceptiuon](https://docs.oracle.com/javase/7/docs/api/java/lang/IllegalArgumentException.html) cleary show that this exeception is to be used in cases like mine:

> Thrown to indicate that a method has been passed an illegal or inappropriate argument.

Michael1993: ### My interpretation:

Both exceptions are correct because 
 1) it is an illegal argument
 2) it is illegal because **_it is null but an object is required_**.

Using `Objects.requireNonNull` is already used to validate arguments inside Pioneer (`TempDirectoryExtension`).
[Its documentation says](https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#requireNonNull-T-) it was designed with this purpose in mind:
> Checks that the specified object reference is not null. This method is designed primarily for doing parameter validation in methods and constructors

Anyways, if you feel strongly about this, I'll leave it alone. It fails fast either way, so what specific exception to throw and how is up to personal preference.

src/test/java/org/junitpioneer/jupiter/StdIoExtensionTests.java
```diff
@@ -215,7 +215,7 @@ void correctConfigurations() {
 		@DisplayName("without input but StdIn parameter, an exception is thrown")
 		void withoutInputWithStdInParameter() {
 			ExecutionResults results = executeTestMethodWithParameterTypes(IllegalConfigurations.class,
-				"noInputButStdIn", "org.junitpioneer.jupiter.StdIn");
+				"noInputButStdIn", StdIn.class);
```
nicolaiparlog, line 218: So much better :heart_eyes: , thanks for the change!

src/test/java/org/junitpioneer/testkit/PioneerTestKit.java
```diff
@@ -38,12 +41,25 @@ public static ExecutionResults executeTestMethod(Class<?> testClass, String test
...
-			String methodParameterTypes) {
-		return new ExecutionResults(testClass, testMethodName, methodParameterTypes);
+			Class<?>... methodParameterTypes) {
+
+		// throw IllegalArgumentException for a `null` array instead of NPE
+		// (hence no use of Objects::requireNonNull)`
+		if (methodParameterTypes == null) {
+			throw new IllegalArgumentException("methodParameterTypes must not be null");
+		}
```
nicolaiparlog, line 57: I removed the politics from the comment. My 2ct: `IllegalArgumentException` is the better exception, but nobody does it, so... :shrug: ... use `Optional`, I guess. :grin:

src/test/java/org/junitpioneer/testkit/PioneerTestKitTests.java
```diff
@@ -0,0 +1,103 @@
...
+		}
+
+		@Test
+		@DisplayName("without parameter results in IllegalArgumentException")
+		void executeTestMethodWithParameterTypes_singleParameterIsNull_IllegalArgumentException() {
+			assertThatThrownBy(() -> PioneerTestKit
+					.executeTestMethodWithParameterTypes(DummyParameterClass.class, "single", (Class<?>[]) null))
+							.isInstanceOf(IllegalArgumentException.class)
+							.hasMessage("methodParameterTypes must not be null");
+		}
```
nicolaiparlog, line 80: There are now two tests to clearly distinguish `null` as a single type from `null` as the type array. Is this the intended behavior?

src/test/java/org/junitpioneer/testkit/PioneerTestKitTests.java
```diff
@@ -0,0 +1,103 @@
...
+							.isInstanceOf(NullPointerException.class);
+		}
+
+		@Test
+		@DisplayName("without parameter results in IllegalArgumentException")
+		void executeTestMethodWithParameterTypes_singleParameterIsNull_IllegalArgumentException() {
+			assertThatThrownBy(() -> PioneerTestKit
+					.executeTestMethodWithParameterTypes(DummyParameterClass.class, "single", (Class<?>[]) null))
+							.isInstanceOf(IllegalArgumentException.class)
+							.hasMessage("methodParameterTypes must not be null");
```
nicolaiparlog, line 79: This is how to test assertions with pure AssertJ (i.e. no Jupiter assertions).

src/test/java/org/junitpioneer/jupiter/StdIoExtensionTests.java
nicolaiparlog: I mean, thanks to this PR, the code looks so much better and more readable.

src/test/java/org/junitpioneer/testkit/PioneerTestKitTests.java
nicolaiparlog: Quick conversation with @Bukama confirmed that this is indeed the intended behavior.

nicolaiparlog:
Talked to @Bukama, he allowed me to merge. :)

Issue #290, opened 2020-06-27
Builds brake due caching problems


Bukama:
Since a few days all builds break due caching problems, like this [build of the Pioneer assertions](https://github.com/junit-pioneer/junit-pioneer/pull/245/checks?check_run_id=808150118)

It seems that there are issues with the GithubAction for caching. @Michael1993 found this [issue about the Gradle / GithubActions caching](https://github.com/eskatos/gradle-command-action/issues/7).

beatngu13:
Haven't had time to read through the linked issue with Gradle/GHA, but I can easily remove the caching for now? Probably better to have a slower but working build.

Bukama:
> Haven't had time to read through the linked issue with Gradle/GHA, but I can easily remove the caching for now? Probably better to have a slower but working build.

Yes. Make it run, then make it fast.

beatngu13:
Had to close #292 because it looks like the caching issue stems from the Gradle command action:

https://github.com/eskatos/gradle-command-action/issues/24

Will try out if [disabling the wrapper cache](https://github.com/eskatos/gradle-command-action#caching) helps.

beatngu13:
Seems like #292 fixes the build problems. However, I would keep this issue open until the Gradle command action is fixed, then we can revert the PR and re-activate the Gradle wrapper caching.

Bukama:
Thank you @beatngu13 ! As @aepfli approved the PR, I just merged it. Let's hope that we get back on track until this here gets fixed (-> blocked)

Bukama:
Maybe related to #325

PR #293, opened 2020-07-01
New behaviour for RetryingTest for Assumptions

### LINKED ISSUE #291: RepeatFailedTest fails if assumptions are invalid
I have a project that I am converting from JUnit 4 to JUnit 5 and it runs tests both with and without the property `java.awt.headless == false`.

To make this work, we have a number of tests with this assumption as the first line of the test:
```
org.junit.Assume.assumeFalse(java.awt.GraphicsEnvironment.isHeadless());
```

That assumption works as intended to skip the test in JUnit 4 or JUnit 5 when annotated with `@Test`, but when annotated with `@RepeatFailedTest(2)`, that assumption causes a failure.

I would think that assumptions would not cause a failure.

Michael1993:
Closes #291.

Proposed commit message:
```
Skip test execution on @RetryingTest when Assumption fails (#291)

Tests that do not meet the Assumptions of those tests should be skipped.
Ensures that RetryingTest tests are skipped after the first test executes if 
an assumption is not met.

Issue: #291
```
---

### DIFF
--- a/docs/retrying-test.adoc
+++ b/docs/retrying-test.adoc
@@ -48,8 +48,9 @@ void aborted() {
 ```
 
 If a test is aborted (e.g.: because of a failed Assumption) `@RetryingTest` won't try again after that.
-The test `aborted` is executed once (because it is aborted).
-The test suite is marked as aborted/skipped, containing the underlying cause.
+The test `aborted` is aborted before (or during) its first execution.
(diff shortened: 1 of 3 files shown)


Michael1993:
Local build successful - please ignore that I accidentally committed twice with the same message :cry:

Bukama:
Please merge the current master into the branch, to remove the Gradle issues mentioned in #290 - thank you!

Issue #179, opened 2020-03-03
Improve ReportEntry Extension


nicolaiparlog:
Once #134 is fixed by #173 we could add more features to the extension. Random ideas:

* introduce variables, e.g. for test name or arguments
* introduce switch when to log, e.g. before or after or on failure
* should there be a way to specify multiple key/value pairs in a single annotation, so they can be put into a map and published as a single report entry?

Michael1993:
Yeah, sure, I would like to try this.

I'm not really keen on the third option but the other two is sorely needed for the extension to feel more 'complete'.

nicolaiparlog:
The prerequisite #134/#173 are done. Let me know, when you (or anybody else) starts working on this.

Michael1993:
I have something in the works but it might take some time as I'm figuring things out. I don't want to make another half-cooked PR and dumb hours of work on you. :worried:

nicolaiparlog:
On the next iteration you (or somebody else) has to add your contribution to the README.

Bukama:
To give a short information what was implemented and what is open for discussion / implementation from my personal point of view:

> * introduce variables, e.g. for test name or arguments

This seems quite useful to me. For arguments I understand something like this:

```
String[] params = {"A", "B", "C"};
...
@ReportEntry(value = "Some entry {0}, {1} and {2}")
```
and for names something like `{displayName} or `{classDisplayName}`, which would result in @ReportEntry(value = "Some entry for {displayName}")`


> * introduce switch when to log, e.g. before or after or on failure

Was talked about yesterday at stream: @Michael1993 has implemented a publish condition to choose when a report entry should be published. Options are : always, on success, on failure and when test was aborted. This resolved one part of this point as the entries are published always at the end of the test execution. I personally see this feature done.


> * should there be a way to specify multiple key/value pairs in a single annotation, so they can be put into a map and published as a single report entry?

Not sure if I understand it correctly but when I don't see a use case here, because the annotation is repeatable. Only case may be grabbing the parameters of a `@ParameterizedTest` - if this is possible at all.

nicolaiparlog:
I'm still very interested in this, @Michael1993, and since we down to a single-digit of open PRs, you might as well open a new one, if you have anything to show. :)

Michael1993:
Not sure about test names. Should that be added too?

Bukama:
> Not sure about test names. Should that be added too?

[...]

PR #294, opened 2020-07-09
Added variables to ReportEntry

Michael1993:
Refer to test method arguments with numbers between curly braces, starting with {0}.

---


Michael1993:
Progresses (possibly will close) #179

nicolaiparlog:
I like the change and I appreciate that the syntax aligns with [Jupiter's parameter-in-display-name syntax](https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests-display-names). Unfortunately, the documentation didn't get updated yet and so we can't merge this. Quoting @aepfli and @Bukama  "DENIED" "NEXT!" :P

docs/report-entries.adoc
```diff
@@ -110,4 +110,31 @@ The extension will publish your report entry if the test fails.
...
+
+Stream<Arguments> params() {
+    return Stream.of(
+            Arguments.of("Hello", 1),
+            Arguments.of("World", 2);
+    );
+}
+
+----
+Then the extension will publish `{1} - Hello` and `{2} - World`.
```
Bukama, line 138: Including the numbers?

Michael1993: I mucked it up again, dang.

Michael1993: Fixed.

docs/report-entries.adoc
```diff
@@ -110,4 +110,31 @@ The extension will publish your report entry if the test fails.
...
 The extension will publish your report entry if the test was aborted (for example because of an https://junit.org/junit5/docs/current/user-guide/#writing-tests-assumptions[Assumption]).
 
-Entries will always be published after the test finished.
\ No newline at end of file
+Entries will always be published after the test finished.
+
+=== Publishing test parameters
+
+You can access `@ParameterizedTest` parameter values and publish them with the `@ReportEntry` annotation.
+Add a number (starting from zero) between curly braces to reference the parameter with that index.
```
Bukama, line 118: I would add an hint like "Inside the report entry the parameter values can be reported in any order".

Question from my side: Can they be reported multiple times, e.g. `{1} - {1} - {0} - {0}`, ?

src/main/java/org/junitpioneer/jupiter/ReportEntryExtension.java
```diff
@@ -16,26 +16,62 @@
...
+
+	private static void verifyParameterCount(ExtensionContext context, ReportEntry entry) {
+		Matcher matcher = Pattern.compile("\\{([0-9])+}").matcher(entry.value());
+		int variableCount = 0;
+		while (matcher.find())
+			variableCount++;
+
+		if (context.getRequiredTestMethod().getParameterCount() < variableCount) {
+			String message = "Report entry contains unresolved variable(s): { key=\"%s\" value=\"%s\" }";
+			throw new ExtensionConfigurationException(format(message, entry.key(), entry.value()));
```
Bukama, line 71: This seems that this message only works if the method has one parameter less than the ReportEntry? What if there are multiple missing?  This misconfiguration should also be mentioned in the docs, as the error messages is only clear (to me) when knowing the  context.

src/main/java/org/junitpioneer/jupiter/ReportEntryExtension.java
```diff
@@ -60,16 +96,35 @@ public void testAborted(ExtensionContext context, Throwable cause) {
...
 				.filter(entry -> Arrays.asList(conditions).contains(entry.when()))
-				// we filter for empty keys/values because this is called if the test failed -
-				// even if it's due to bad extension configuration (but we don't publish for those)
-				.filter(entry -> !entry.key().isEmpty() && !entry.value().isEmpty())
-				.forEach(entry -> context.publishReportEntry(entry.key(), entry.value()));
+				.forEach(entry -> context.publishReportEntry(entry.key(), parseVariables(entry.value(), context)));
+	}
+
+	private String parseVariables(String value, ExtensionContext context) {
+		if (!value.matches(".*\\{.*}.*"))
```
Bukama, line 111: Shouldn't there be a start of line and end of line symbol in the regEx?

Bukama: I don't understand the regex. The `\\` means you check for `\` (first backslash is to escape second baskslash), but why even check for a baskslash? Then inside the curly braces you check for any character which may or may not be there (there can be no character between the curly braces). But I think (look below in line 117) you search for numeric (interger) values. So why you don't check for them inside the regex?

docs/report-entries.adoc
Michael1993: Yes, they can be reported multiple times, I'll update the tests (and the documentation) to reflect that.

src/main/java/org/junitpioneer/jupiter/ReportEntryExtension.java
Michael1993: - We don't really care where we find the regex inside the `String` so there is no need for [boundary matchers](https://docs.oracle.com/javase/tutorial/essential/regex/bounds.html).
- `{` and `}` are 'special' characters in regexes (quantifiers?). For example `a{5}` means _exactly_ five `a` characters. Because of this, the opening `{` needs to be "escaped" in the regex. Normally this would be done with `\{` however, in Java, `\` is an escape character already (e.g.: `\n` is newline). So, we have to put `\\` in the regex - first, to escape from the String, then to escape from the regex. Worse than Alcatraz, really.
- This is an early return, checking for a negative - if there are no curly braces in the value, we return immediately because there are no variables in the `@ReportEntry` value. If there are, we try to parse them and throw an exception if they cannot be parsed. An argument could be made that empty curly braces, i.e.: `{}` should not throw an exception - if you think that should be changed, let me know.

src/main/java/org/junitpioneer/jupiter/ReportEntryExtension.java
Michael1993: Why does it seem that way? I don't see it. Could you elaborate?

Bukama: Because there is one key/value reported and an exception thrown, so no more are tested. Or do I have a mind bug?

src/main/java/org/junitpioneer/jupiter/ReportEntryExtension.java
Bukama: I should stop reviewing, of course `\\{` is the regex and not `\{` 🤦

src/main/java/org/junitpioneer/jupiter/ReportEntryExtension.java
Michael1993: No, I see what you mean. This is incorrect. Will fix shortly.

Michael1993: Should be fixed now.

docs/report-entries.adoc
```diff
@@ -110,4 +110,32 @@ The extension will publish your report entry if the test fails.
...
```
nicolaiparlog, line 118: Can you mention and link to the format you're using here? (Is it `String::format` or `MessageFormat`?)

docs/report-entries.adoc
```diff
@@ -110,4 +110,32 @@ The extension will publish your report entry if the test fails.
...
+Inside the report entry, the parameters can be referenced in any order and multiple times.
+For example, if you have a `@ParameterizedTest` like the following:
+
+[source, java]
+----
+@ParameterizedTest
+@MethodSource("params")
```
nicolaiparlog, line 125: For small parameter sets, `@CsvSource` is more succinct and easier to read. Can you switch to it?

docs/report-entries.adoc
```diff
@@ -110,4 +110,32 @@ The extension will publish your report entry if the test fails.
...
+
+Stream<Arguments> params() {
+    return Stream.of(
+            Arguments.of("Hello", 21),
+            Arguments.of("World", 42);
+    );
+}
+
+----
+Then the extension will publish `21 - Hello - 21` and `42 - World - 42`.
```
nicolaiparlog, line 139:
```suggestion
Then the extension will publish `"21 - Hello - 21"` and `"42 - World - 42"`.
```

src/main/java/org/junitpioneer/jupiter/ReportEntryExtension.java
```diff
@@ -16,33 +16,78 @@
...
 	}
 
+	private static void verifyReportEntry(ExtensionContext context, ReportEntry entry) {
+		verifyParameterCount(context, entry);
+		verifyKeyValueAreNotBlank(entry);
+		verifyKeyNotParameterized(entry);
+	}
+
+	private static void verifyParameterCount(ExtensionContext context, ReportEntry entry) {
+		boolean valueReferencesNumberedParameter = entry.value().matches(".*\\{[0-9]+}.*");
```
nicolaiparlog, line 57: Maybe extract `matches(".*\\{[0-9]+}.*")` into a method?

docs/report-entries.adoc
Michael1993: Changed to `@CsvSource`

src/main/java/org/junitpioneer/jupiter/ReportEntryExtension.java
Michael1993: Agree, extracted and slightly refactored.

docs/report-entries.adoc
Bukama: As far as I see - this is still open right?

Michael1993: Did I misunderstand? I did mention that the format of the variables is in line with JUnit's `@ParameterizedTest`

Bukama: The way I understood Nicolais question is that he wanted to see you mention `String::format` or `MessageFormat`, but maybe I missunderstood him.

Issue #296, opened 2020-07-13
Multiversion build with different JUnit versions


aepfli:
Currently our extension is built for version 5.5.2, but what if somebody with version 5.5.0 or 5.6.2 wants to use JUnit Pioneer. Do we know if this is possible? We should think of experimental builds to cover most JUnit5 Versions out there. So we know, if our application at least builds with those versions, and if the tests are passing.

As soon as there is a new version, we should also add it to our build pipeline, So we can ensure backwards and forward compatibility. Ideally we run this just on master, but we can also think of doing this for pull requests, but for pull requests, i do not see much benefit for this. What do you think @junit-pioneer/junit-pioneer-team ?

aepfli:
https://github.com/junit-pioneer/junit-pioneer/actions/runs/167772059

 as there is an pr open from myside already, because i found it interesting there are also the action above to take a look at :)

nicolaiparlog:
First of all, very good point, we should strive to be compatible with as many JUnit versions as possible. Thinking of it, I think the docs don't even say as much. A policy like "be compatible with all versions in the range from [required; newest]" would make a lot of sense. Maybe this is a good time to discuss and possibly document.

Regarding your specific questions, I'm not sure on which branches to build this because:

1. building for many versions takes a while I assume, so on PRs this would make feedback unacceptably slow
2. since we're only developing in branches, PRs should be built against all versions, so we can spot problems before we merge them
3. since any new JUnit 5 versions could break Pioneer, we need to run master against newest versions as soon as they are released

Yes, this is a nice hodgepodge of contradictions. Not sure what the best course of action would be.

Bukama:
I like the idea and we could use this argument also to upgrade to the newest Pioneer version also (I personal don't see so much obstacles to upgrade a JUnit Jupiter version in comparison to something like Hibernate where you are bound to JPA-Specc etc.) - but that's another topic.

For me it's okay to run every PR against all other JUnit versions (but should do this only on LTS java versions). If we are honest to ourself: the only time where the duration of builds is somehow "annoying" is when you are "live coding" like in the streams, but that's not the regular thing.

> 3. since any new JUnit 5 versions could break Pioneer, we need to run master against newest versions as soon as they are released

That only means we have to update the matrix in the script after there's an JUnit release. I think that's fine.

aepfli:
as we all agree here, i remove the label in discussion, and i will continue with my implementation

PR #297, opened 2020-07-13
Adding experimental build for multiple JUnit5 versions (#296)

aepfli:
There are still some things to discuss, or things which imho are not that great within that solution.

1. JUnitVersion actually is just the patch.minor version and not the whole version. This is due to the fact that JUnit platform and JUnit jupiter are using the same patch.minor version, but sadly not the major one. We could think of changing this to some logic in the build.gradle.kts. Nevertheless it will be a good solution in any case, in my opinion. Providing two properties is also not a good idea.
2. When to run, is it overkill to run this on pull request? i think we do not gain a lot by doing so. but it could be quiet interesting :)

Commit Message:
```
Adding experimental build for multiple JUnit5 versions (#296)

Adding additional GitHub Action Workflow to check our compatibility with past and future
versions of JUnit5. Currently we are supporting 5.5.0, 5.5.1, 5.5.2, 5.6.0, 5.6.1 and 5.6.2
in this workflow.

closes: #296 
```

---

### DIFF
--- a/.github/workflows/experimentalJUnitVersion.yml
+++ b/.github/workflows/experimentalJUnitVersion.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Setup java
         uses: actions/setup-java@v1
         with:
-          java-version: 14
+          java-version: 11
       - name: Cache Gradle
         uses: actions/cache@v2
         with:
@@ -36,5 +36,5 @@ jobs:
       - name: Gradle build
         uses: eskatos/gradle-command-action@v1
         with:
-          arguments: --refresh-dependencies --stacktrace --scan clean build -PjunitVersion=${{ matrix.junit5 }}
+          arguments: --refresh-dependencies --stacktrace --scan clean build -PjunitMinorVersion=${{ matrix.junit5 }}
           wrapper-cache-enabled: false
--- a/.github/workflows/experimentalJUnitVersion.yml
+++ b/.github/workflows/experimentalJUnitVersion.yml
@@ -18,8 +18,8 @@ jobs:
     continue-on-error: true
     strategy:
       matrix:
-        junit5: [ 5.5.0, 5.5.1, 5.5.2, 5.6.0, 5.6.1, 5.6.2 ]
-    name: build with ${{ matrix.junit5 }}
+        junit5: [ 5.0, 5.1, 5.2, 6.0, 6.1, 6.2 ]
+    name: build with 5.${{ matrix.junit5 }}
     steps:
       - name: Check out repo
         uses: actions/checkout@v2
--- a/.github/workflows/experimentalJUnitVersion.yml
+++ b/.github/workflows/experimentalJUnitVersion.yml
@@ -18,8 +18,8 @@ jobs:
     continue-on-error: true
     strategy:
       matrix:
-        junit5: [ '5.0', '5.1', '5.2', '6.0', '6.1', '6.2' ]
-    name: with 5.${{ matrix.junit5 }}
+        junit5Minor: [ '5.0', '5.1', '5.2', '6.0', '6.1', '6.2' ]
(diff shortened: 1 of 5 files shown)


.github/workflows/experimentalJUnitVersion.yml
```diff
@@ -0,0 +1,40 @@
...
+      matrix:
+        junit5: [ '5.0', '5.1', '5.2', '6.0', '6.1', '6.2' ]
+    name: with 5.${{ matrix.junit5 }}
+    steps:
+      - name: Check out repo
+        uses: actions/checkout@v2
+      - name: Setup java
+        uses: actions/setup-java@v1
+        with:
+          java-version: 14
```
Bukama, line 29: Why 14? Default-version if Jupiter and Pioneer is 8!

src/main/java/org/junitpioneer/jupiter/params/DisableIfDisplayName.java
```diff
@@ -16,7 +16,6 @@
 import java.lang.annotation.Target;
```
Bukama, line 16: Please revert those changes - they have nothing to do with the topic of this PR!

aepfli: actually they have :D because DisplayIf was removed in 5.6.0 and due to that the build fails - i thought i will just fix this with this build

.github/workflows/experimentalJUnitVersion.yml
aepfli: the target version is 8 - does not mean you can not build it with newer versions, that is just a minor detail, but the latest LTS is 14 so i would go with our experimental topics with a recent version

Bukama: Latest LTS is 11, next will be 17 :)

14 is just current supported version until September (?) when 15 comes out :)

src/main/java/org/junitpioneer/jupiter/params/DisableIfDisplayName.java
Bukama: Oh okay :)

.github/workflows/experimentalJUnitVersion.yml
aepfli: wait isnt it every third release an lts... did i miss again something?

aepfli: my bad... i somehow had in my mind that every 3rd release is an LTS (not sure where this is coming from) - but than 11 would make more sense, in my opinion

Bukama: I had the same misinformation in my mind for a long time too, but it's every 3rd year (= sixth release), not every 3rd release :)

nicolaiparlog:
Relates to #296

gradle.properties
```diff
@@ -0,0 +1 @@
+junitVersion=5.2
```
nicolaiparlog, line 1: This is confusing because it sounds like we're building against 5.2. Calling it `junitMinorVersion` and adding a comment showing how the variable is used should fix that.

.github/workflows/experimentalJUnitVersion.yml
```diff
@@ -0,0 +1,40 @@
...
+  pull_request:
+    branches: '*'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    strategy:
+      matrix:
+        junit5: [ '5.0', '5.1', '5.2', '6.0', '6.1', '6.2' ]
```
nicolaiparlog, line 21: `junit5Minor` plus comment (see below)

Bukama:
A bit curious that two jobs don't start?

Issue #265, opened 2020-05-25
Documentation: Personal or formal style


Bukama:
@aepfli brought up this documentation topic inside a PR, which I cut out into this issue, as we have to decide, have to update the `CONTRIBUTING.md` file and maybe several documentation files.

> i just want to raise a discussion here, about our documentation style, or more about unifying it - and maybe find a guideline on how to write docs. Currently it seems like we have two different approaches:
> 
> 1. the personal/informal one, in which we address the developer and his work directly with you and your tests (like here)
> 2. the formal one, in which it seems like we try to avoid such constructs (eg. RangeSource)
> 
> For a good comprehensive documentation i think we should agree on one style to use. As it makes reading the docs easier to grasp and shows some kind of consistency. The questions are
> 
> * is this something which only bothers me? than lets continue, as i did not say anything :)
> * what would be the general preference. I tend to use the formal one, simply because it feels more 'technical' to me.
> 
> What is the general take on this?

Bukama:
I always thought that I'm the only one that tries to write formal style, but nice to see that I'm not alone.

The formal one has several advantages in my point of view:
* More serious
* Better focus on technical description, which brings up "better" text (with personal one you easily fall into a more conversational style)
* No one feels "touched personally"

This said I vote for keeping our documentation the formal style.

Michael1993:
This is probably a non-issue because it seems I'm the only one who wrote informal style documentation so far - in [the report entries documentation](../tree/master/docs/report-entries.adoc) and in #227.

If @nicolaiparlog agrees that formal documentation everywhere is the way to go, I'll fix these.

aepfli:
i do not have a dedicated opinion on that, i just prefer consistency :D but as it seems like we do have an agreemet, we should at least use this ticket to persist it some how in the documentation and in the end close this topic

Bukama:
I had a short talk about this with @nicolaiparlog and he will write his opinion down when he finds the time. But he mentioned that there are other things to consider, e.g. Mannings (the publisher of his book) wants the authors to address their readers directly.

nicolaiparlog:
While I tend to write somewhat formal (judging by the amount of passive voice that is used (see what I did there?)), I was told by my publisher that an active voice and addressing the reader with "you" is preferred. If you read carefully, you may see that this style is indeed more engaging and thus better to transport information.

[...]

PR #316, opened 2020-08-07
Clean up GitHub Actions scripts

beatngu13:
This PR does clean up the Pioneer GitHub Actions scripts by several means:

* Avoid unshallow fetch during release since Sonar is not involved and, therefore, full history is not needed.
* Do unshallow fetch during build via new `actions/checkout` option.
* Consistently name steps, jobs, workflows (using sentence case) and the corresponding files (using kebab case).

Proposed commit message:

```
Clean up GitHub Actions scripts (#316)

* Avoid unshallow fetch during release since Sonar is not involved
and, therefore, full history is not needed.
* Do unshallow fetch during build via new `actions/checkout` option.
* Consistently name steps, jobs, workflows (using sentence case) and
the corresponding files (using kebab case).
```

---
**PR checklist**

Documentation (general)

Documentation (new extension)

Code

Contributing

---


.github/workflows/junit5-build.yml
```diff
@@ -18,11 +15,11 @@ jobs:
     strategy:
       matrix:
         junit5Minor: [ '5.0', '5.1', '5.2', '6.0', '6.1', '6.2' ]
-    name: with 5.${{ matrix.junit5Minor }}
+    name: Build with JUnit 5.${{ matrix.junit5Minor }}
```
aepfli, line 18: i want to start a discussion here, i think this `build` is redundant as it is part of the name of the action, and i think it has a nice reading in the gui without it. I just want to check in what you think about - but that was my initial thought when i removed it.

image
image

beatngu13: Yeah, totally agree! Saw it in the UI after the PR. So yes, I will remove that redundancy.

aepfli: i also thought about to rename it like, build with Java 8 on ubuntu - because i think naming wise it is nicer to have the first information aligned rather than the last one

aepfli: maybe we should also name the gradle wrapper validation gha as `validate` and the job `Gradle Wrapper` - so we could theoretically validate more in there, and would also have the same naming scheme

beatngu13: Removed the redundancy and now included the OS everywhere, I think this plays well with the workflow name.

Not sure about your last comment. Usually, I'm in favor of validating (in the sense of `mvn validate`) as soon as possible in the main build pipeline, so that the build aborts early in case of a failure.

However, I'm open in this regard. What do you think?

aepfli: what i meant is that the gradle validation could also be a just a file called `validation.yml` instead of `gradle-wrapper-validation.yml` - the name of the GHA could be `Validate` and the name of the (currently only) job could be `gradle wrapper` compared to now where the GHA is named `Gradle Wrapper Validation` and the job is named `validate`. to be consistent with the other namings, and to allow us to add other validations into this GHA :)

PR #321, opened 2020-08-14
CartesianProductTestExtension from JUnit examples

Michael1993:
Opened to let us move forward with #68 

---
**PR checklist**

Documentation (general)
* [X] There is documentation (Javadoc and site documentation; added or updated)
* [X] There is implementation information to describe _why_ a non-obvious source code / solution got implemented
* [X] Site documentation has its own `.adoc` file in the `docs` folder, e.g. `docs/report-entries.adoc`
* [X] Only one sentence per line (especially in `.adoc` files)
* [X] Javadoc uses formal style, while sites documentation may use informal style (see #265)

Documentation (new extension)
* [X] The `docs/docs-nav.yml` navigation has an entry for the new extension
* [X] The `package-info.java` contains information about the new extension

Code
* [X] There are checks which validate correct / false usage / configuration of a functionality and there are tests to verify those checks (see #164)
* [X] Tests use [AssertJ](https://joel-costigliola.github.io/assertj/) or our own [PioneerAssert](https://github.com/junit-pioneer/junit-pioneer/blob/master/CONTRIBUTING.md#assertions) (which are based on AssertJ)

Contributing

---


Michael1993:
Things to consider:
* [ ] Explicit factory method option (i.e. `@CartesianProductTest(factory = "getSources")`)
* [ ] Supporting `@ArgumentsSource` annotations (which are **not** repeatable, apparently) -> [relevant JUnit 5 issue](https://github.com/junit-team/junit5/issues/1101)
* [ ] What should this throw, `AssertionError` or `ExtensionConfigurationException`?

src/main/java/org/junitpioneer/jupiter/CartesianProductTestExtension.java
```diff
@@ -0,0 +1,99 @@
...
+
+		return invokeSetsFactory(testMethod, factoryMethod).getSets();
+	}
+
+	private CartesianProductTest.Sets invokeSetsFactory(Method testMethod, String factoryMethodName) {
+		Class<?> declaringClass = testMethod.getDeclaringClass();
+		Method factory = PioneerUtils
+				.findMethodCurrentOrEnclosing(declaringClass, factoryMethodName)
+				.orElseThrow(() -> new AssertionError("Method `CartesianProductTest.Sets " + factoryMethodName
+						+ "()` not found in " + declaringClass + "or any enclosing class"));
```
Bukama, line 64: If I understand this right the declaration of the `CartesianProductTest.Sets` must be in the same test class or a parent class. Why force the users to use such test data structure instead allowing them to define the test data in a complete different class?

src/main/java/org/junitpioneer/jupiter/PioneerUtils.java
```diff
@@ -44,4 +48,32 @@ private PioneerUtils() {
...
+	 * Find the first {@link Method} of the supplied class or interface that
+	 * meets the specified criteria, beginning with the specified class or
+	 * interface and traversing its enclosing classes until such a method is
+	 * found or the top level class is reached.
+	 *
+	 * <p>The algorithm does not search for methods in {@link java.lang.Object}.
+	 *
+	 * @param clazz the class or interface in which to find the method; never {@code null}
+	 * @param methodName the name of the method to find; never {@code null} or empty
+	 * @param parameterTypes the types of parameters accepted by the method, if any;
```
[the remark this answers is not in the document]
Bukama: According `never null`: You can't ensure that the user does not pass `null`. The underlying methods may check this and throw an exception. So remove this in the parameter description and a `@throws` section where you describe the exception which is thrown when one of the parameters is `null`.

src/test/java/org/junitpioneer/jupiter/CartesianProductTestExtensionTests.java
```diff
@@ -0,0 +1,185 @@
...
+
+	static CartesianProductTest.Sets nFold() {
+		return new CartesianProductTest.Sets()
+				.add("Alpha", "Omega")
+				.add(Runnable.class, Comparable.class, TestInfo.class)
+				.add(TimeUnit.DAYS, TimeUnit.HOURS);
+	}
+
+	static CartesianProductTest.Sets supplyValues() {
+		return new CartesianProductTest.Sets().add("War", "Peace").add(TimeUnit.SECONDS, TimeUnit.DAYS);
```
Bukama, line 74: Linebreaks for each `.add` like in `nFold`.

src/test/java/org/junitpioneer/jupiter/CartesianProductTestExtensionTests.java
```diff
@@ -0,0 +1,185 @@
...
+
+		@CartesianProductTest({ "0", "1" })
+		void threeBits(String a, String b, String c) {
+			int value = Integer.parseUnsignedInt(a + b + c, 2);
+			Assertions.assertThat(value).isBetween(0b000, 0b111);
+		}
+
+		@CartesianProductTest
+		@DisplayName("S ⨯ T ⨯ U")
+		void nFold(String string, Class<?> type, TimeUnit unit, TestInfo info) {
```
Bukama, line 50: What does `nFold` mean?

src/test/java/org/junitpioneer/jupiter/CartesianProductTestExtensionTests.java
```diff
@@ -0,0 +1,185 @@
...
+
+	static CartesianProductTest.Sets supplyValues() {
+		return new CartesianProductTest.Sets().add("War", "Peace").add(TimeUnit.SECONDS, TimeUnit.DAYS);
+	}
+
+	@Nested
+	class BadConfigurationTests {
+
+		@Test
+		@DisplayName("Test fails when there is no factory method")
```
Bukama, line 81: You could put the `Test fails, when` into a `@DisplayName` of the `@Nested` class.

src/test/java/org/junitpioneer/jupiter/CartesianProductTestExtensionTests.java
```diff
@@ -0,0 +1,185 @@
...
+		}
+
+		@Test
+		@DisplayName("Test fails if the factory does not produce enough parameters")
+		void throwsForTooFewFactoryMethodParameters() {
+			ExecutionResults results = PioneerTestKit
+					.executeTestMethodWithParameterTypes(BadConfigurationTest.class, "incompleteFactory", int.class,
+						String.class, TimeUnit.class);
+
+			assertThat(results).hasNumberOfDynamicallyRegisteredTests(3).hasNumberOfFailedTests(3);
```
Bukama, line 123: Can a message be provided (and asserted) in such case?

src/test/java/org/junitpioneer/jupiter/CartesianProductTestExtensionTests.java
```diff
@@ -0,0 +1,185 @@
...
+		@DisplayName("Test fails if the factory produces too much parameters")
+		void throwsForTooManyFactoryMethodParameters() {
+			ExecutionResults results = PioneerTestKit
+					.executeTestMethodWithParameterTypes(BadConfigurationTest.class, "bloatedFactory", int.class,
+						String.class);
+
+			assertThat(results)
+					.hasSingleFailedContainer()
+					.withExceptionInstanceOf(AssertionError.class)
+					.hasMessageContaining("must register values for each parameter");
```
Bukama, line 136: Why is it possible to generate a message when there a too much parameters, but not too less (see test above)?

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,122 @@
...
+There are a couple conditions this method has to fulfill to qualify:
+
+- must have the same name as the test method or its name must be specified via `factory()`
+- must be `static`
+- must have **no** parameters
+- must return `CartesianProductTest.Sets` ^1^
+- must register values for every parameter exactly once ^2^
+
+1: This is a helper class, specifically for creating sets for `@CartesianProductTest`.
+The way to use it: instantiate it (`new CartesianProductTest.Sets()`), then use the `add()` method to register values for the parameters (see <<_static-factory-example, above>>).
```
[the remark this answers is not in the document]
Michael1993: You can't provide `CartesianProductTest.Sets` as an annotation parameter, so I'm not sure what you meant here.

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,122 @@
...
+This is most likely unwanted and will probably lead to the following exception/failure message:
+
+> org.junit.jupiter.api.extension.ParameterResolutionException:
+> No ParameterResolver registered for parameter [...]
+
+This is because `@Test` does not know what to do with the parameter(s) of the `@CartesianProductTest`.
+
+== Thread-Safety
+
+This extension is safe to use during https://junit.org/junit5/docs/current/user-guide/#writing-tests-parallel-execution[parallel test execution].
```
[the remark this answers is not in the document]
Michael1993: [This is the proper way to format links in Asciidoc](https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#links).

src/main/java/org/junitpioneer/jupiter/CartesianProductTest.java
```diff
@@ -0,0 +1,68 @@
...
+public @interface CartesianProductTest {
+
+	/**
+	 * Specifies {@code String} values for all inputs simultaneously.
+	 */
+	String[] value() default {};
+
+	String factory() default "";
+
+	class Sets {
```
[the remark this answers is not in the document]
Michael1993: I don't know what you mean. `Sets` is an interface member - automatically `public` and `static`.

src/main/java/org/junitpioneer/jupiter/PioneerUtils.java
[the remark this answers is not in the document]
Michael1993: This is copy-pasted from JUnit 5s `ReflectionSupport`, it means that the method will throw an exception if `null` is supplied as a parameter.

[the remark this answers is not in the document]
Michael1993: Your first suggestion is not in-line with other classes in Pioneer. [For example, `EnvironmentVariableUtils`](https://github.com/junit-pioneer/junit-pioneer/blob/master/src/main/java/org/junitpioneer/jupiter/EnvironmentVariableUtils.java#L52).

src/test/java/org/junitpioneer/jupiter/CartesianProductTestExtensionTests.java
Michael1993: Spotless. 🤷 
Would you prefer I did `@formatter:off`?

src/test/java/org/junitpioneer/jupiter/CartesianProductTestExtensionTests.java
Michael1993: Even though all tests fail, the container is considered successful. No clue why.

src/test/java/org/junitpioneer/jupiter/CartesianProductTestExtensionTests.java
Michael1993: Don't know how to test for too few parameters, unless we check for auto-injected parameters (like `TestReporter` or `TestInfo`).

src/test/java/org/junitpioneer/jupiter/CartesianProductTestExtensionTests.java
Bukama: To be honest: Yes, Think it's more readable then and consistent like in `nFold`

Michael1993: Alright. Fair enough.

src/main/java/org/junitpioneer/jupiter/CartesianProductResolver.java
```diff
@@ -0,0 +1,37 @@
...
+
+import java.util.List;
+
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.extension.ParameterContext;
+import org.junit.jupiter.api.extension.ParameterResolver;
+
+public class CartesianProductResolver implements ParameterResolver {
+
+	private final List<?> parameters;
```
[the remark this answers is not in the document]
Michael1993: I'm not sure, since `CartesianProductTest.Sets` is just a wrapper around a `List<List<?>>`. Or maybe I misunderstood what you meant... again. 😅

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,122 @@
...
+| 6th test   | "Alpha"           | Predicate.class  | TimeUnit.HOURS
+| 7th test   | "Omega"           | Runnable.class   | TimeUnit.DAYS
+| 8th test   | "Omega"           | Runnable.class   | TimeUnit.HOURS
+| 9th test   | "Omega"           | Cloneable.class  | TimeUnit.DAYS
+| 10th test  | "Omega"           | Cloneable.class  | TimeUnit.HOURS
+| 11th test  | "Omega"           | Predicate.class  | TimeUnit.DAYS
+| 12th test  | "Omega"           | Predicate.class  | TimeUnit.HOURS
+|===
+
+There are a couple conditions this method has to fulfill to qualify:
```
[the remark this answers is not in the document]
Michael1993: Thank you, I rewrote this section entirely.

docs/cartesian-product.adoc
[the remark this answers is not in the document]
Michael1993: I agree that `CartesianProductTest.Sets`should be introduced earlier in the docs, so I moved it up. Do you think it needs more explanation than this?

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,122 @@
...
+| 7th test  | 1            | 1            | 0
+| 8th test  | 1            | 1            | 1
+|===
+
+== Writing a static factory method for the parameters
+
+`@CartesianProductTest` can have a static factory method supplying the test parameters.
+By default, this method must have the same name as the test method, but you can specify a different name with the `factory()` annotation parameter.
+[#_static-factory-example]
+```java
```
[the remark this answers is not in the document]
Michael1993: Um, whoops? Thank you, updated. 😅

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,122 @@
...
+- must have **no** parameters
+- must return `CartesianProductTest.Sets` ^1^
+- must register values for every parameter exactly once ^2^
+
+1: This is a helper class, specifically for creating sets for `@CartesianProductTest`.
+The way to use it: instantiate it (`new CartesianProductTest.Sets()`), then use the `add()` method to register values for the parameters (see <<_static-factory-example, above>>).
+
+2: If you register too few, too many, or conflicting ^3^ parameters, you will get an `org.junit.jupiter.api.extension.ParameterResolutionException`.
+
+3: For example, if your test method has a `TestInfo` or `TestReporter` (or any other parameter that is automatically injected by JUnit) and you register too much sets in your factory method:
```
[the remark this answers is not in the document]
Michael1993: Good idea, did just that.

src/main/java/org/junitpioneer/jupiter/CartesianProductTestExtension.java
```diff
@@ -0,0 +1,99 @@
...
+
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.extension.TestTemplateInvocationContext;
+import org.junit.jupiter.api.extension.TestTemplateInvocationContextProvider;
+
+class CartesianProductTestExtension implements TestTemplateInvocationContextProvider {
+
+	@Override
+	public boolean supportsTestTemplate(ExtensionContext context) {
+		return findAnnotation(context.getTestMethod(), CartesianProductTest.class).isPresent();
```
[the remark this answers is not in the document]
Michael1993: Bit of an overkill, don't you think?

src/main/java/org/junitpioneer/jupiter/PioneerUtils.java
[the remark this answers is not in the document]
Michael1993: I did a bit of reading, and it seems like adding `@throws` for `RuntimeExceptions` is not recommended?

src/test/java/org/junitpioneer/jupiter/CartesianProductTestExtensionTests.java
Michael1993: No clue, this is copy-pasted from JUnit examples.

src/test/java/org/junitpioneer/jupiter/CartesianProductTestExtensionTests.java
Michael1993: Good idea, thank you.

src/test/java/org/junitpioneer/jupiter/CartesianProductTestExtensionTests.java
Michael1993: There should be a discussion about extending Pioneer assertions as need for it arises, (such as in this case).

src/test/java/org/junitpioneer/jupiter/CartesianProductTestExtensionTests.java
Michael1993: What would be a good alternative?

beatngu13:
I just briefly skimmed over the PR, really like that feature. Personally, I already needed the Cartesian product several times in the past. Kudos @Michael1993!

I have only one question: When it comes to supplying test parameters, I just wonder if it would be helpful to stick to the style of `@ParameterizedTest`, where the parameter source is defined via a separate annotation?

src/test/java/org/junitpioneer/jupiter/CartesianProductTestExtensionTests.java
Bukama: As I don't understand it it can't give a good alternative.

src/test/java/org/junitpioneer/jupiter/CartesianProductTestExtensionTests.java
Bukama: Okay.

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,150 @@
...
+
+[#_basic_use]
+== Basic Use
+
+`@CartesianProductTest` is used _instead_ of `@Test` or other such annotations (e.g. `@RepeatedTest`).
+The annotation can have a value (see <<Supplying a value to CartesianProductTest>>), or the test class can have a static factory method (see <<Writing a static factory method for the parameters>>) providing the arguments.
+
+== Supplying a value to CartesianProductTest
+
+If all your test parameters are strings, you can supply all input parameters simultaneously by giving a value to `@CartesianProductTest`.
```
Bukama, line 15: > giving a value 

With a single value there is no Cartesian product (or to be exactly: The CP only has one row). Would change it to plural, e.g. "by giving values to. These values are input for all parameters." or something like that.

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,150 @@
...
+
+```java
+@CartesianProductTest({"0", "1"})
+void threeBits(String a, String b, String c) {
+    int value = Integer.parseUnsignedInt(a + b + c, 2);
+    Assertions.assertThat(value).isBetween(0b000, 0b111);
+}
+```
+
+The test `threeBits` is executed exactly eight times, because all three input parameters can have the value "0" or "1".
```
Bukama, line 26: values

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,150 @@
...
+| 4th test  | 0            | 1            | 1
+| 5th test  | 1            | 0            | 0
+| 6th test  | 1            | 0            | 1
+| 7th test  | 1            | 1            | 0
+| 8th test  | 1            | 1            | 1
+|===
+
+== Writing a static factory method for the parameters
+
+`@CartesianProductTest` can have a static factory method supplying the test parameters.
```
Bukama, line 44: Would add an introduction sentence like "If you don't only supply String values to your test method like in the example above..."

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,150 @@
...
+By default, this method must have the same name as the test method, but you can specify a different name with the `factory()` annotation parameter.
+This method must return `CartesianProductTest.Sets`.
+`CartesianProductTest.Sets` is a helper class, specifically for creating sets for `@CartesianProductTest`.
+The way to use it: instantiate it (`new CartesianProductTest.Sets()`), then use the `add()` method to register values for the parameters (see <<_static-factory-example, the example>>).
```
Bukama, line 48: Capital `I` after `:`

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,150 @@
...
+| 5th test   | "Alpha"           | Predicate.class  | TimeUnit.DAYS
+| 6th test   | "Alpha"           | Predicate.class  | TimeUnit.HOURS
+| 7th test   | "Omega"           | Runnable.class   | TimeUnit.DAYS
+| 8th test   | "Omega"           | Runnable.class   | TimeUnit.HOURS
+| 9th test   | "Omega"           | Cloneable.class  | TimeUnit.DAYS
+| 10th test  | "Omega"           | Cloneable.class  | TimeUnit.HOURS
+| 11th test  | "Omega"           | Predicate.class  | TimeUnit.DAYS
+| 12th test  | "Omega"           | Predicate.class  | TimeUnit.HOURS
+|===
+
```
Bukama, line 98: As you describe the input of `nFold` here this should be moved up to be in place with the example and after that show the possibility to use an individual name for the data provider method (without further explanation)

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,150 @@
...
+=== Conditions for the static factory method
+
+There are a couple conditions the static factory method has to fulfill to qualify:
```
Bukama, line 101: "There are some conditions" or "There are a couple of conditions" or "There are several conditions"

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,150 @@
...
+- must be `static`
+- must have **no** parameters
+- must return `CartesianProductTest.Sets`
+- must register values for every parameter exactly once
+- must register values in order
+
+=== Returning wrong `Sets` in the static factory method
+
+If you register too few, too many, or conflicting parameters, you will get an `org.junit.jupiter.api.extension.ParameterResolutionException`.
+For example, if your test method has a `TestInfo` or `TestReporter` (or any other parameter that is automatically injected by JUnit), and you register too many sets in your factory method:
```
Bukama, line 113: This line reads like there is missing something at the end. Would change it to "The following example shows..." or something like this.

src/main/java/org/junitpioneer/jupiter/CartesianProductTest.java
```diff
@@ -0,0 +1,71 @@
...
+ * because it will throw an exception.
+ * </p>
+ *
+ * <p>Methods annotated with this annotation are different from {@code ParameterizedTest}s because
+ * they can not have {@code ArgumentsSource}s - those are completely disregarded. Instead a
+ * String array must be provided (for methods with only String parameters) or a static factory method
+ * with the same name as the test method must exist.
+ * </p>
+ *
+ * @since ???
```
Bukama, line 38: Remove as you don't know. The changelog will show the version the extension comes available.

src/main/java/org/junitpioneer/jupiter/CartesianProductTest.java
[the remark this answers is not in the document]
Bukama: I see the `Sets` thing as a dataholder class and in itself has nothing to do with the Cartesian production interface. I personally don't like to squeeze a general class/interface into a specific one.

src/main/java/org/junitpioneer/jupiter/CartesianProductTestExtension.java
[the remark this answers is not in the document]
Bukama: Then you argue to throw Pioneer Annotations out, because we can use the JUnit things. The idea was Pioneer Annotation was to encapsulate the JUnit thing with our own, improved methods. But I'll leave it to you / the others which way we use.

Same for TestKit or PioneerAssertions 🤷‍♂️

Bukama (changes requested):
Some smaller things in the docs I obviusly missed on the first turn. Sorry for that. After that it's okay for me.

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,218 @@
...
+> +[...]+
+> One can similarly define the Cartesian product of n sets, also known as an n-fold Cartesian product, which can be represented by an n-dimensional array, where each element is an n-tuple.
+
+What does all this mean?
+
+The cartesian product of sets is all the possible combinations where you take a single element from each set.
+If you have two sets, `{ 1, 2 }` and `{ 3, 4 }`, their cartesian product is `{ { 1, 3 }, { 1, 4 }, { 2, 3 }, { 2, 4 } }`.
+
+You can use `@CartesianProductTest` and define test parameter sets to run the same test on the cartesian product.
+This means that `@CartesianProductTest` takes the possible test parameter values and runs the test for every possible combination of them.
```
Bukama, line 17: For a better transition between the theoretical part and the extension I would change this to something like: 

_Sometimes its useful to test all possible combinations of parameters set. Normally this results in a lot of written test data parameters. For a more comfortable way you may use the `CartesianProductTest` extension. The extension takes the test data parameter values and runs...._

The basic usage of the extension (with the parameter sets) are describe below anyway.

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,218 @@
...
+
+== Basic Use
+
+`@CartesianProductTest` is used _instead_ of `@Test` or other such annotations (e.g. `@RepeatedTest`).
+The annotation can have a value (see <<Supplying a value to CartesianProductTest>>), the test method can be annotated with `@CartesianValueSource` (see <<Annotating your test method with @CartesianValueSource>>), or the test class can have a static factory method (see <<Writing a static factory method for the parameters>>) providing the arguments.
```
Bukama, line 22: I don't know which words I should choose, but this sentence lacks some kind of introduction, like to "There are multiple ways to pass test data to the extension.". As of now you (or at least I) fall "into" this sentences without knowing what to expect.

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,218 @@
...
+
+`@CartesianProductTest` works with sets.
```
Bukama, line 24: The Cartesian product works with sets, so the extension does. (more fluent I think)

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,218 @@
...
+This means that duplicate elements get removed.
+If your input is `{ 1, 1, 3, 4 }` and `{ 2, 2, 3 }` their cartesian product is `{ { 1, 2 }, { 1, 3 }, { 3, 2 }, { 3, 3 }, { 4, 2 }, { 4, 3 } }`.
+This is because otherwise the test would run with the same parameters multiple times.
+If you need that functionality, you might want to look into https://junit.org/junit5/docs/current/user-guide/#writing-tests-repeated-tests[repeated tests].
```
Bukama, line 28: You should add if the duplicates get removed automatically (most probably as `Set` can't work with duplicates). To emphasize that I would remove the final Cartesian product (you described above how it is created) and change the line to the (automatically) "real/reduced" input, e.g.

_If your input is `{ 1, 1, 3, 4 }` and `{ 2, 2, 3 }` the input sets get reduced to  `{ { 1, 3, 4 }, { 2, 3 } }`._

I would change _If you need that functionality_ to _If you need to pass the same parameters multiple times, you might.._ to make clear what you mean with functionality.

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,218 @@
...
+
+== Supplying a value to CartesianProductTest
```
Bukama, line 30: I don't like the wording of "a value". I know it's because `value` is the default name of the annotation variable, but "a value" mislead to "a single value", which is false, because you are passing String arrays.

What about something like "Supplying (simple) sets of Strings to the extension"?

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,218 @@
...
+
+If all your test parameters are strings, you can supply all input parameters simultaneously by giving a string array value to `@CartesianProductTest`.
+This value is the input for all parameters, and the test will try every combination of its elements.
```
Bukama, line 33: No `,` before `and`- this is no enumeration

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,218 @@
...
+
+```java
+@CartesianProductTest({ "0", "1" })
+void threeBits(String a, String b, String c) {
+    // passing test code
+}
+```
+
+The test `threeBits` is executed exactly eight times, because all three input parameters can have the values "0" or "1".
+`@CartesianProductTest` tests for all input combinations; that's `2 × 2 × 2`, so eight tests in total.
```
Bukama, line 43: Change `;` to `,`, because _that's .._ is a subordinate clause.

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,218 @@
...
+| 4th test  | "0"            | "1"            | "1"
+| 5th test  | "1"            | "0"            | "0"
+| 6th test  | "1"            | "0"            | "1"
+| 7th test  | "1"            | "1"            | "0"
+| 8th test  | "1"            | "1"            | "1"
+|===
+
+== Annotating your test method with @CartesianValueSource
+
+If you don't only supply String values to your test method like in the example above, you can annotate your method with `@CartesianValueSource`.
```
Bukama, line 60: _Instead of only suppliying string values to your ... you can annotate ... with `CartesianValueSource` [to do what <- thats missing]

Or written in another way: If I don't supply string values I don't supply them. Period. Starting a sentence like this makes the reader feel that there are (bad) consequences if not doing so.

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,218 @@
...
+This is annotation might look familiar - it mimics JUnits `@ValueSource`.
```
Bukama, line 61:
```suggestion
This is annotation might look familiar (it mimics JUnits `@ValueSource`), but it works differently.....[insert explanation].
```

Or explanation below at end of subsection.

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,218 @@
...
+It is used to define the possible inputs of a single test parameter.
+You define the parameter values in order (left-to-right or top-to-bottom) and the test will try every combination those values can have.
```
Bukama, line 63: `left-to-right or top-to-bottom`: The `or` irritates me. Is there any forced order or not? You example seems like the order of the annotation must match the parameter order? Are the names of "ints" and "strings" fixed or forced too?

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,218 @@
...
+@CartesianValueSource(strings = { "A", "B" })
+void annotated(int number, String character) {
+    // passing test code
+}
+```
+
+The test `annotated` is executed exactly six times.
+The first parameter can have any of the three values `1`, `2` or `4`.
+The second parameter can have any of the two values `A` or `B`.
+`@CartesianProductTest` tests for all input combinations; that's `3 × 2`, so six tests in total.
```
Bukama, line 77: `;`to `,`

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,218 @@
...
+| # of test | value of `number` | value of `character`
+| 1st test  | 1                 | "A"
+| 2nd test  | 1                 | "B"
+| 3rd test  | 2                 | "A"
+| 4th test  | 2                 | "B"
+| 5th test  | 4                 | "A"
+| 6th test  | 4                 | "B"
+|===
+
+Please note that `@CartesianValueSource` is not compatible with `@ParameterizedTest`.
```
Bukama, line 90: Move this up to the explanation (or add a subsubsection / some sentences) here to explain the differences.

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,218 @@
...
+
+== Writing a static factory method for the parameters
+
+Finally, if your tests require special inputs that `@CartesianValueSource` is not able to supply, you can define a static factory method to supply your test parameters.
```
Bukama, line 94: Remove the "finally,"

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,218 @@
...
+By default, this method must have the same name as the test method, but you can specify a different name with the `factory()` annotation parameter.
+This method must return `CartesianProductTest.Sets`.
+`CartesianProductTest.Sets` is a helper class, specifically for creating sets for `@CartesianProductTest`.
+The way to use it: Instantiate it (`new CartesianProductTest.Sets()`), then use the `add()` method to register values for the parameters (see <<_static-factory-example, the example>>).
```
Bukama, line 98: What about a more fluent sentence like: _To create a set with test data, instantiate a new `CartesianProductTest.Sets()`, use the `add()`... and return the created object._

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,218 @@
...
+        .add(Runnable.class, Cloneable.class, Predicate.class)
+        .add(TimeUnit.DAYS, TimeUnit.HOURS);
+}
+```
+
+The test `nFold` is executed exactly twelve times.
+The first parameter can have any of the two values `"Alpha"` or `"Omega"`.
+The second parameter can have any of the three values `Runnable.class`, `Cloneable.class` or `Predicate.class`.
+The third parameter can have any of the two values `TimeUnit.DAYS` or `TimeUnit.HOURS`.
+`@CartesianProductTest` tests for all input combinations; that's `2 × 3 × 2`, so twelve tests in total.
```
Bukama, line 119: again

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,218 @@
...
+| 6th test   | "Alpha"           | Predicate.class  | TimeUnit.HOURS
+| 7th test   | "Omega"           | Runnable.class   | TimeUnit.DAYS
+| 8th test   | "Omega"           | Runnable.class   | TimeUnit.HOURS
+| 9th test   | "Omega"           | Cloneable.class  | TimeUnit.DAYS
+| 10th test  | "Omega"           | Cloneable.class  | TimeUnit.HOURS
+| 11th test  | "Omega"           | Predicate.class  | TimeUnit.DAYS
+| 12th test  | "Omega"           | Predicate.class  | TimeUnit.HOURS
+|===
+
+Remember, you can reuse the same argument provider if you explicitly give its name to `@CartesianProductTest`.
```
Bukama, line 138:
```suggestion
Remember, you can reuse the same argument provider method, by explicitly pass its name to the  `@CartesianProductTest` `factory` attribute.
```

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,218 @@
...
+        .add("Mercury", "Earth", "Venus")
+        .add(1, 12, 144);
+}
+```
+
+=== Conditions for the static factory method
+
+There are multiple conditions the static factory method has to fulfill to qualify:
+
+- must have the same name as the test method or its name must be specified via `factory()`
```
Bukama, line 163:
```suggestion
- must have the same name as the test method (or its name must be specified via `factory()`)
```

You are listing conditions of the method. So the same name is only in default case (as you described above)  and I would put the note about the `factory()` method as a reminder in brackets here.

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,218 @@
...
+}
+
+static CartesianProductTest.Sets resolveParameters() {
+    return new CartesianProductTest.Sets()
+        .add("A", "B", "C")
+        .add(1, 2, 3);
+}
+```
+
+== Combining `@CartesianProductTest` with `@Test`
```
Bukama, line 206:
```suggestion
== Warning: Don't combine `@CartesianProductTest` with `@Test`!
```

src/main/java/org/junitpioneer/jupiter/CartesianValueArgumentsProvider.java
```diff
@@ -0,0 +1,68 @@
...
+                        source.strings(),
+                        source.classes()
+                )
+                .filter(array -> Array.getLength(array) > 0)
+                .collect(toList());
+        // @formatter:on
+
+		Preconditions
+				.condition(arrays.size() == 1, () -> "Exactly one type of input must be provided in the @"
+						+ CartesianValueSource.class.getSimpleName() + " annotation, but there were " + arrays.size());
```
Bukama, line 55: _ Spotless what did you do here?_

Bukama (changes requested):
Again only changes in the docs requested.
On thing in general (not mentioned explicit at any spot): You should decide if you write `String` in upper case or lower case (`string`) when talking about the `String` class.

docs/cartesian-product.adoc
Michael1993: Left-to-right = top-to-bottom. There is a forced order.

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,222 @@
...
+> In terms of set-builder notation, that is `A × B = {(a,b) | a ∈ A and b ∈ B}`
+> +[...]+
+> One can similarly define the Cartesian product of n sets, also known as an n-fold Cartesian product, which can be represented by an n-dimensional array, where each element is an n-tuple.
+
+What does all this mean?
+
+The cartesian product of sets is all the possible combinations where you take a single element from each set.
+If you have two sets, `{ 1, 2 }` and `{ 3, 4 }`, their cartesian product is `{ { 1, 3 }, { 1, 4 }, { 2, 3 }, { 2, 4 } }`.
+
+Sometimes its useful to test all possible combinations of parameters set.
```
Bukama, line 16:
```suggestion
Sometimes its useful to test all possible combinations of parameters sets.
```

Bukama (approved):
Great update! The docs now read very fluent and are super understandable! 🥳 

Nevertheless: One typo I think ;)

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,222 @@
+:page-title: Testing all test parameter combinations (Cartesian product)
+:page-description: Extends JUnit Jupiter with `@CartesianProductTest`, a special `@ParameterizedTest` which tests all combinations of its input
+
+From Wikipedia, the free encyclopedia:
```
nicolaiparlog, line 4: "the free encyclopedia" seems a bit patronizing, I think most people know that small project already. :wink:

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,222 @@
...
+
+> In mathematics, specifically set theory, the Cartesian product of two sets A and B, denoted A × B, is the set of all ordered pairs (a, b) where a is in A and b is in B.
+> In terms of set-builder notation, that is `A × B = {(a,b) | a ∈ A and b ∈ B}`
+> +[...]+
+> One can similarly define the Cartesian product of n sets, also known as an n-fold Cartesian product, which can be represented by an n-dimensional array, where each element is an n-tuple.
+
+What does all this mean?
```
nicolaiparlog, line 11: Love this! :heart: :+1:

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,222 @@
...
+
+The cartesian product of sets is all the possible combinations where you take a single element from each set.
+If you have two sets, `{ 1, 2 }` and `{ 3, 4 }`, their cartesian product is `{ { 1, 3 }, { 1, 4 }, { 2, 3 }, { 2, 4 } }`.
+
+Sometimes it's useful to test all possible combinations of parameter sets.
+Normally, this results in a lot of written test data parameters.
+For a more comfortable way you may use the `@CartesianProductTest` extension.
+The extension takes the test data parameter values and runs the test for every possible combination of them.
+
+== Basic Use
```
nicolaiparlog, line 21: I think the section _Basic Use_ could do with a simple, working code snippet.

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,222 @@
...
+
+Just like the mathematical Cartesian product, `@CartesianProductTest` works with sets.
+Duplicate elements get removed automatically.
+If your input is `{ 1, 1, 3 }` and `{ 2, 2 }` the extension will consider their Cartesian product `{ { 1, 2 }, { 3, 2 } }`.
+Otherwise, the test would run with the same parameters multiple times.
+If you need to pass the same parameters multiple times, you might want to look into https://junit.org/junit5/docs/current/user-guide/#writing-tests-repeated-tests[repeated tests].
+
+== Supplying CartesianProductTest with a single set
+
+If all your test parameters are strings, you can supply all input parameters simultaneously by giving a string array value to `@CartesianProductTest`.
```
nicolaiparlog, line 36: The section header talks about single sets, which made me expect a way to define sets of all types (e.g. a single set of `int`s), but the text only mentions `String` arrays. Does this work for all kinds of arrays? If so/not, this should be stated explicitly.

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,222 @@
...
+```java
+@CartesianProductTest
+@CartesianValueSource(ints = { 1, 2, 4 })
+@CartesianValueSource(strings = { "A", "B" })
+void annotated(int number, String character) {
+    // passing test code
+}
+```
+
+This annotation might look familiar - it mimics JUnits `@ValueSource`, except `@CartesianValueSource` is repeatable.
```
nicolaiparlog, line 78: Good connection to something the reader (may) already know! :+1: Add a link to the user guide or docs to drive it home.

docs/cartesian-product.adoc
```diff
@@ -0,0 +1,222 @@
...
+- must be `static`
+- must have **no** parameters
+- must return `CartesianProductTest.Sets`
+- must register values for every parameter exactly once
+- must register values in order
+
+=== Returning wrong `Sets` in the static factory method
+
+If you register too few, too many, or conflicting parameters, you will get an `org.junit.jupiter.api.extension.ParameterResolutionException`.
+Conflicting parameters meaning your test method has a `TestInfo` or `TestReporter` (or any other parameter that is automatically injected by JUnit), and you register too many sets in your factory method.
```
nicolaiparlog, line 177: This made me think, I could not have `TestInfo` or `TestReporter` params in my test method signature.

nicolaiparlog: If `TestInfo` or `TestReporter` works with correct configurations, that may be worth a mention as well.

nicolaiparlog (commented):
Just had time for the documentation, but I really like it. Good explanation of the feature and I also like the feature it describes. I'll check out the code another time.

docs/cartesian-product.adoc
Michael1993: It works if you don't try to "manually" inject your own `TestReporter`. I will clean up the wording of this sentence.

docs/cartesian-product.adoc
Michael1993: Added a basic snippet without explanation

docs/cartesian-product.adoc
Michael1993: Cleaned up wording to explicitly mean `String[]`

docs/cartesian-product.adoc
Michael1993: Link added.

docs/cartesian-product.adoc
Michael1993: Cleaned up, mention added.

docs/cartesian-product.adoc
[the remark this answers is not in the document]
Michael1993: @Bukama ping. I still would like to hear your thoughts on this, please. 😇

src/test/java/org/junitpioneer/jupiter/CartesianProductTestExtensionTests.java
Michael1993: @Bukama If you think the new introduction properly explains why this example is named `nFold` I will close this discussion.

docs/cartesian-product.adoc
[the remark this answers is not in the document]
Bukama: If not I would have mentioned in the review :)

src/test/java/org/junitpioneer/jupiter/CartesianProductTestExtensionTests.java
Bukama: Didn't thought this "discussion" is still relevant. The explanation of `nFold` is in the into and so I'm totally fine :)

docs/cartesian-product.adoc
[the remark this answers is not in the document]
Michael1993: Thank you!

nicolaiparlog:
This looks really good: great documentation and I love how simple `CartesianProductExtension` is at its heart - really captures the mathematical simplicity of the idea.

Unfortunately, the tests aren't sufficient yet. As demonstrated on stream all but one pass even if we break the extension big time (by repeating the same list of arguments the right number of times). Your idea (from Twitch chat) to pipe the arguments into the test reporter and verify that way sounds really good. Should be comparatively straightforward and thread-safe out of the box. :+1: 

To answer your questions:

> What should this throw, `AssertionError` or `ExtensionConfigurationException`?

For misconfigured extensions, we should never use `AssertionError` - `ExtensionConfigurationException` sounds good to me.

> How do different argument sources interact? Should it be illegal to specify multiple argument sources?

Yeah, I think the extension should verify that there aren't conflicting sources, e.g, `factory` attribute and `@CartesianValueSource`.

> Should `CartesianProductTest` be moved to `org.junitpioneer.jupiter.params` instead of `org.junitpioneer.jupiter`?

No, we use the `params` package for extensions to Jupiter's existing parameterized test feature. While this extension is similar (and uses the same test templating extension point), this is "coincidental" and the extension does interact with parameterized tests.

src/main/java/org/junitpioneer/jupiter/CartesianProductTestExtension.java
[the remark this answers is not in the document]
nicolaiparlog: The advantage of our annotation methods is that they make it easy to find meta annotations that are _indirectly present: (if a supertype of the element is annotated), _meta-present_ (if an annotation that is present on the element is itself annotated), or _enclosing-present (if an enclosing type [think opposite of `@Nested`] is annotated). As I see it, only _meta-present_ is relevant here, but that should indeed be allowed. Not sure if `AnnotationSupport.findAnnotation` does that.

src/main/java/org/junitpioneer/jupiter/CartesianProductTestExtension.java
```diff
@@ -0,0 +1,157 @@
...
+		}
+		if (!CartesianProductTest.Sets.class.isAssignableFrom(factory.getReturnType())) {
+			throw new ExtensionConfigurationException(method + " must return `CartesianProductTest.Sets`");
+		}
+		CartesianProductTest.Sets sets = (CartesianProductTest.Sets) invokeMethod(factory, null);
+		if (sets.getSets().size() > testMethod.getParameterCount()) {
+			throw new ParameterResolutionException(format(
+				"%s must register values for each parameter exactly once. Expected [%d] parameter sets, but got [%d]",
+				method, testMethod.getParameterCount(), sets.getSets().size()));
+		}
```
nicolaiparlog: What if `sets.getSets().size() == testMethod.getParameterCount()`, but one of the parameters is resolved by JUnit itself? Then we should still see the exception.

src/main/java/org/junitpioneer/jupiter/CartesianProductTestNameFormatter.java
```diff
@@ -0,0 +1,76 @@
...
+import static org.junit.jupiter.params.ParameterizedTest.DISPLAY_NAME_PLACEHOLDER;
+import static org.junit.jupiter.params.ParameterizedTest.INDEX_PLACEHOLDER;
+
+import java.text.MessageFormat;
+import java.util.Arrays;
+import java.util.stream.IntStream;
+
+import org.junit.jupiter.api.extension.ExtensionConfigurationException;
+
+class CartesianProductTestNameFormatter {
```
nicolaiparlog, line 24: I really like it that this is its own class. :+1:

nicolaiparlog (commented):
A few details...

src/main/java/org/junitpioneer/jupiter/CartesianProductTestExtension.java
Michael1993: Can't, JUnit throws an exception before we get here - competing ParameterResolvers.

nicolaiparlog: Oh, good to know! Can you add a comment saying as much?

Michael1993:
From the [documentation of `AnnotationSupport`](https://junit.org/junit5/docs/5.0.0/api/org/junit/platform/commons/support/AnnotationSupport.html#findAnnotation-java.lang.reflect.AnnotatedElement-java.lang.Class-):

> Find the first annotation of annotationType that is either directly present, meta-present, or indirectly present on the supplied element.
If the element is a class and the annotation is neither directly present nor meta-present on the class, this method will additionally search on interfaces implemented by the class before finding an annotation that is indirectly present on the class.

It should be enough to use here.

Issue #322, opened 2020-08-16
Provide timeout for Github Actions


Bukama:
Currently sometimes our builds run "forever" and are killed after six hours, e.g. [run 988295906](https://github.com/junit-pioneer/junit-pioneer/pull/321/checks?check_run_id=988295906). 

When I remember correctly there was a timeout flag you can provide to the actions to timeout after a certain of time, maybe 15mins or something. 

Am I right (@aepfli @beatngu13 ) and do we want such a thing?

beatngu13:
Thanks for making me aware of this. The example job you posted is with Java 11 on Windows, and it looks like it timed out during the tests:

```none
2020-08-15T13:32:27.2087572Z > Task :processTestResources
2020-08-15T13:32:27.2087905Z > Task :testClasses
2020-08-15T13:32:57.2085938Z > Task :test
2020-08-15T19:30:51.8898754Z ##[error]The operation was canceled.
2020-08-15T19:30:51.9603377Z Post job cleanup.
2020-08-15T19:30:52.1317368Z Post job cleanup.
2020-08-15T19:30:54.1273868Z [command]"C:\Program Files\Git\bin\git.exe" version
```

Would be interesting to see if this always happens on the same job during the same Gradle task. Do you have more examples @Bukama?

In the meantime, I will open PR do include an appropriate timeout.

Bukama:
> Thanks for making me aware of this. The example job you posted is with Java 11 on Windows, and it looks like it timed out during the tests:
>
> [...]}

> Would be interesting to see if this always happens on the same job during the same Gradle task. Do you have more examples @Bukama?

[961045944 - win / java 11](https://github.com/junit-pioneer/junit-pioneer/pull/261/checks?check_run_id=961045944)

I don't remember where, but there were one or two more runs, but as I don't remember the runs I don't know if they were also win / java 11

aepfli:
https://github.com/junit-pioneer/junit-pioneer/runs/990846534

beatngu13:
So this time Java 14, but still on Windows.

aepfli:
should we increase the logoutput to info for a while, so we have more information?

beatngu13:
Sounds reasonable to me. But before doing so, we should probably create a separate issue (will take care) for the build timeout since this here will be closed once #323 is merged. Then we can collect all necessary information and decisions there.

Bukama:
I merged #323 to have the timeout in the main build.

[...]

PR #323, opened 2020-08-16
Add timeout to main build (#322)

beatngu13:
Proposed commit message:

```none
Add timeout to main build (#322)

According to #322, the main build occasionally times out. Until
further investigation, this PR will add a timeout to the main build.

PR: #323
```

Note that I didn't close the corresponding issue on purpose. Maybe we want to rephrase it because I think we are better off investigating _why_ the build actually runs forever. The added timeout only addresses a sympton, not the disease.

---
**PR checklist**

Documentation (general)

Documentation (new extension)

Code

Contributing

---

### DIFF
--- a/.github/workflows/main-build.yml
+++ b/.github/workflows/main-build.yml
@@ -11,6 +11,7 @@ on:
 jobs:
   build:
     runs-on: ${{ matrix.os }}-latest
+    timeout-minutes: 15
     strategy:
       matrix:
         # test against latest update of each major Java version, as well as specific updates of LTS versions:


Bukama:
> 
> Note that I didn't close the corresponding issue on purpose. Maybe we want to rephrase it because I think we are better off investigating _why_ the build actually runs forever. The added timeout only addresses a sympton, not the disease.
> 

I think it's okay to do this seperated and first kill the sympton to don't block resources too long. I think to define a well chosen timeout even when there are no known issues is fine.

aepfli:
i do understand both sides regarding issue, and i am fine with both - so @Bukama and @beatngu13 as soon as you agreed on that, either merge or let me know and i will merge :)

beatngu13:
I'm all good with @Bukama's suggestions. Well-chosen timeouts are indeed sensible*, I just wanted to emphasize that we have to dig deeper to find out what's actually the issue. I don't care if we create one, two or more issues for that. ✌️

*: Should we add them to the other workflows as well?

PR #318, opened 2020-08-08
Limit docs code snippets line length to max 60 chars (#317 / #318)

Bukama:
I don't see a way to cut the examples in  `TempDir` extension, so I leave this to maybe refactored method names in #277

```
Limit docs code snippets line length to max 60 chars (#317 / #318)

To avoid horizontal scroll bars inside the code examples of the
documentation pages the line length of them must not more
that 60 characters.
```
---
**PR checklist**

The following checklist shall help the PR's author, the reviewers and maintainers to ensure the quality of this project.
It is based on our contributors guidelines, especially the ["writing code" section](https://github.com/junit-pioneer/junit-pioneer/blob/master/CONTRIBUTING.md#writing-code).
It shall help to check for completion of the listed points.
If a point does not apply to the given PR's changes, the corresponding entry can be simply marked as done. 

Documentation (general)
* [X] There is documentation (Javadoc and site documentation; added or updated)
* [X] There is implementation information to describe _why_ a non-obvious source code / solution got implemented
* [X] Site documentation has its own `.adoc` file in the `docs` folder, e.g. `docs/report-entries.adoc`
* [X] Only one sentence per line (especially in `.adoc` files)
* [X] Javadoc uses formal style, while sites documentation may use informal style (see #265)

Documentation (new extension)
* [X] The `docs/docs-nav.yml` navigation has an entry for the new extension
* [X] The `package-info.java` contains information about the new extension

Code
* [X] Code adheres to code style, naming conventions etc.
* [X] Successful tests cover all changes
* [X] There are checks which validate correct / false usage / configuration of a functionality and there are tests to verify those checks (see #164)
* [X] Tests use [AssertJ](https://joel-costigliola.github.io/assertj/) or our own [PioneerAssert](https://github.com/junit-pioneer/junit-pioneer/blob/master/CONTRIBUTING.md#assertions) (which are based on AssertJ)

Contributing
* [X] A prepared commit message exists
* [X] The list of contributions inside `README.md` mentions the new contribution (real name optional) 

---

I hereby agree to the terms of the [JUnit Pioneer Contributor License Agreement](https://github.com/junit-pioneer/junit-pioneer/blob/master/CONTRIBUTING.md#junit-pioneer-contributor-license-agreement).

Issue #317, opened 2020-08-08
Limit line length to max 60 chars in site documentation code snippets


Bukama:
To avoid horizontal scroll bars inside the code examples of the documentation pages the line length of them must not more that 60   characters.

PR #334, opened 2020-09-08
Use AsciiDoc snippets to include code snippets in documentation

### LINKED ISSUE #335: AsciiDoc code snippets in documentation
Demonstrated in #334, we can use a neat built-in feature of AsciiDoc to include live code examples in our `.adoc` documentation.

The proper way to utilize this feature warrants a discussion.

Michael1993:
closes #335


Bukama:
How do you limit to (about) 60 chars per line? (see #317 / #318)

nicolaiparlog:
I really like this approach to showing code snippets because we can always be sure that they actually work. @Bukama's concerns are valid, though. Tests and code snippets have very different concerns and we should hence separate them.

I propose to create a new source set `demo` (as `src/demo/java`) and put all the demos there. They must be written as proper tests, so we can execute them during the build, but they can be written and styled according to the documentation's needs.

If possible, we could even use Spotless to limit line length to 60 chars (if we can have different configurations for different folders - I think we already had that request once).

nicolaiparlog:
Also, @Michael1993 can you please open an issue, so we can discuss there?

Michael1993:
I opened #335 if that's OK?

Bukama:
Any motivation to further work on this @Michael1993?

Michael1993:
> Any motivation to further work on this @Michael1993?

Whoops, I forgot about this. I'll do some work on it in December.

demo/build.gradle.kts
```diff
@@ -0,0 +1,64 @@
+plugins {
+    java
+    id("com.diffplug.spotless")
+}
```
nipafx, line 4: Is it necessary to have a new build file (with mostly redundant entries)? Is there no way to put this under `src/demo/java` and "just" add that folder to be built/executed by the existing build?

In case it's not strictly necessary and only easier, it would be nice to know how much harder the other approach is because to me it seems more obvious (I really like the `demo` folder under `src`) and probably easier to maintain.

demo/src/test/java/org/junitpioneer/jupiter/ReportEntryExtensionTests.java
```diff
@@ -0,0 +1,71 @@
...
+package org.junitpioneer.jupiter;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestReporter;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
+
+public class ReportEntryExtensionTests {
+
+	// tag::no_report_entry[]
```
nipafx, line 20: Are the `[]` needed? So ceremonial... (in a bad way)

nipafx (commented):
Looks good. The extra ceremony of creating, tagging, and referencing the source file is non-trivial, though. I think the benefits are worth it, but we could discuss whether short snippets (like, three lines or so) could be allowed in the Asciidoc or whether this is the new standard for all tests, regardless of complexity.

demo/src/test/java/org/junitpioneer/jupiter/ReportEntryExtensionTests.java
Michael1993: Unfortunately, yes. Non-avoidable. :(

build.gradle.kts
```diff
@@ -195,8 +195,20 @@ tasks {
 			if (modularBuild.toBoolean())
 				java.srcDir("src/main/module")
 		}
+		create("demo") {
+			java {
+				srcDir("src/demo/java")
+			}
+			compileClasspath += sourceSets.main.get().output
+			runtimeClasspath += sourceSets.main.get().output
```
aepfli, line 203:
```suggestion
			runtimeClasspath += sourceSets.main.get().output
			sonarqube {
			    skipProject = true
			}
```

this should theoretically do the trick, and you can remove the build.gradle file

build.gradle.kts
```diff
@@ -195,8 +195,20 @@ tasks {
...
+		}
 	}
```
aepfli, line 209:
```suggestion
	project(":demo") {
		sonarqube {
			isSkipProject = true
		}
	}
```

this should do the trick :)

build.gradle.kts
```diff
@@ -195,8 +195,20 @@ tasks {
 			if (modularBuild.toBoolean())
 				java.srcDir("src/main/module")
 		}
+		create("demo") {
```
aepfli, line 198: can we maybe utilize the test suite feature for this?

aepfli:
```
testing {
  suites {
    val test by getting(JvmTestSuite::class) { useJUnitJupiter() }

    val demoTests by
        registering(JvmTestSuite::class) {
          dependencies { implementation(project) }

          sources { java { srcDir("src/demo/java") } }
          targets { all { testTask.configure { shouldRunAfter(test) } } }
        }
  }
}
```

would do the trick, but we would need to update our gradle wrapper to version 7.4, and apply the `jvm-test-suite` plugin - with that a demoTest task will be automatically added.

The dependson section would look like
```
dependsOn(javadoc, validateYaml, testing.suites.named("demoTests"))
```

Michael1993: What's the advantage of that over the current configuration?

Michael1993:
The branch is ready for review - however if it gets merged, there should be another PR that migrates the rest of the docs to the new 'system'.

build.gradle.kts
aepfli: it is more concise and you do not need to manuially configure the configruations and the classpaths. also the test tasks is automatically created, and the whole configuration for `demotests` is just the block

Michael1993: Okay, I think we should involve the others and discuss on the 2022-02-10 maintainer meeting (because of the expanded scope).

aepfli: i am not sure what you mean with expanded scope - this is just a new feature which was specially introduced by gradle for such a case - before that, you had to do it, like it is now :)

Michael1993: I thought the gradle update would be considered expanded scope... Now I'm not so sure.

aepfli: okay yeah the upgrade from 7.3 to 7.3.3 might be a good topic. and i agree

src/demo/java/org/junitpioneer/jupiter/ReportEntryExtensionTests.java
```diff
@@ -0,0 +1,71 @@
...
+ */
+
+package org.junitpioneer.jupiter;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestReporter;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
+
+public class ReportEntryExtensionTests {
```
nipafx, line 18: Can we name these files `...Demo` instead of `...Tests`?.

build.gradle.kts
nipafx: I don't see a reason to discuss. Updating a build-time dependency to the next minor/patch version is everyday business. 😉

nipafx: Oh, and I think @aepfli's proposal to use test suites is 100% worth the update.

build.gradle.kts
```diff
@@ -215,6 +231,8 @@ tasks {
 		useJUnitPlatform()
 		filter {
 			includeTestsMatching("*Tests")
+			// included for documentation purposes
+			// includeTestsMatching("*Demo")
```
nipafx, line 235: Is this correct?

```suggestion
			includeTestsMatching("*Demo")
```

Other than that, I think we can merge.

Michael1993: Nope. This is just dead code.

Michael1993:
> > however if it gets merged, there should be another PR that migrates the rest of the docs to the new 'system'.
> 
> Can you create an issue for that?

#595

docs/report-entries.adoc
```diff
@@ -1,5 +1,6 @@
 :page-title: Publishing Report Entries
 :page-description: Extends JUnit Jupiter with `@ReportEntry`, which publishes report entries using `TestReporter`.
+:snippetsource: ../src/demo/java/org/junitpioneer/jupiter/ReportEntryExtensionTests.java
```
nipafx, line 3: This attribute points at the wrong file.

That shows that this way of referencing is brittle. Would be nice to verify during a build that the docs actually "work" but then we'd need to include some asciidoc in the build, which.... ugh. Let's ignore that if it doesn't cause too many problems.

nipafx:
Can you extend `CONTRIBUTING.md` with this text (at the end of the section "Feature Documentation"):

	Code blocks in these files should not be pure text.
	Instead, in the `src/demo/java` source tree, create/update a `...Demo` class that is dedicated to this feature and place code snippets in `@Test`-annotated methods in `...Demo`.
	Write each snippet as needed for the documentation and bracket it with tags:

	Code blocks in these files should not be pure text.
	Instead, in the `src/demo/java` source tree, create/update a `...Demo` class that is dedicated to this feature and place code snippets in `@Test`-annotated methods.
	Write each snippet as needed for the documentation and bracket it with tags:

	```java
	// tagging the entire test method:

	// tag::$TAG_NAME[]
	@Test
	@SomePioneerExtension
	void simple() {
		// demonstrate extension
	}
	// end::$TAG_NAME[]


	// tagging a few lines from the test:

	@Test
	void simple() {
		// tag::$TAG_NAME[]
		SomePioneerExtension ex = // ...
		// demonstrate extension
		// end::$TAG_NAME[]
		assertThat(ex). // ...
	}
	```

	Where feasible, include or follow up with assertions that ensure correct behavior.
	Thus `...Demo` classes guarantee that snippets compile and (roughly) behave as explained.

	In the documentation file, include an attribute pointing at the demo source file:

	```adoc
	:snippetsource: ../src/demo/java/org/junitpioneer/jupiter/...Demo.java
	```

	To include these snippets, use a block like the following:

	```adoc
	[source,java,indent=0]
	----
	include::{snippetsource}[tag=$TAG_NAME]
	----
	```

Proposed commit message:

```
Configure build and docs to compile and test code snippets (#335 / #334)

Code snippets in the documentation files in `docs` are just text and
thus run the risk of getting outdated when APIs evolve or
implementations change their behavior. To address that, we now use
Asciidoc's capability to include tagged regions from external
files[1] that are compiled and tested by the build. These files'
primary purpose is to demonstrate features (not to test them) and so
they are placed in a new source set `src/demojava`.

[1]: https://docs.asciidoctor.org/asciidoc/latest/directives/include-tagged-regions/

Closes: #335
PR: #334 
```

docs/report-entries.adoc
Michael1993: What do you mean? It works on _my_ machine! 😱 

Yeah, that's annoying, I'll fix it.

CONTRIBUTING.md
```diff
@@ -205,11 +205,61 @@ Some project-specific requirements apply to all non-`.java` files:
 Each feature is documented on [the project website](https://junit-pioneer.org/docs/), which is pulled from the files in the `docs/` folder, where each feature has:
 
 * an entry in `docs-nav.yml` (lexicographically ordered)
-* it's own `.adoc` file
+* its own `.adoc` file
```
nipafx, line 208: Uff! Thank you for that fix!

tim2346: Rm

Issue #324, opened 2020-08-17
Build jobs sometimes timeout


Bukama:
In the last time several job builds timeout. As far as we can identify they are all Windows jobs at Java 11 or 14:

> Thanks for making me aware of this. The example job you posted is with Java 11 on Windows, and it looks like it timed out during the tests:
> 
> ```
> 2020-08-15T13:32:27.2087572Z > Task :processTestResources
> 2020-08-15T13:32:27.2087905Z > Task :testClasses
> 2020-08-15T13:32:57.2085938Z > Task :test
> 2020-08-15T19:30:51.8898754Z ##[error]The operation was canceled.
> 2020-08-15T19:30:51.9603377Z Post job cleanup.
> 2020-08-15T19:30:52.1317368Z Post job cleanup.
> 2020-08-15T19:30:54.1273868Z [command]"C:\Program Files\Git\bin\git.exe" version
> ```

> 
> [961045944 - win / java 11](https://github.com/junit-pioneer/junit-pioneer/pull/261/checks?check_run_id=961045944)
> 

>  [990846534 - win / java 14](https://github.com/junit-pioneer/junit-pioneer/runs/990846534)

Bukama:
[1105113930 - win / java 14](https://github.com/junit-pioneer/junit-pioneer/pull/326/checks?check_run_id=1105113930)

nicolaiparlog:
Too bad I don't have Windows set up for development to take a look at this. :cry: :grin: Looks like somebody else will have to check this out. :yum: 

And we should actually fix this!

Bukama:
Just as a reminder:
When fixing this `windows` has to be added to both, the main workflow and also the module build workflow.

Michael1993:
It's not just Windows! It seems ubuntu builds have been timing out as well.

nipafx:
I ran about 170k tests and I didn't take notice of any deadlocks (although if just a few threads lock, I wouldn't have noticed). This was on Gentoo with kernel version 5.9.16.

nipafx:
We've investigated this a bit and the only weird behavior we've discovered locally is that lots of classes are loaded and unloading blocks the JVM (0% CPU usage, etc) for a while. We could trace that back to the use of `EqualsVerifier` in `IssueTestCaseTests` and `IssueTestSuiteTests`.

Entire test suite (after ~100k tests):

[...]

PR #342, opened 2020-09-26
Temporarily disable windows builds (#324 / #342)

Bukama:
In #324 it is stated out that quite often
the builds on Windows time out, most probably
because of some kind of deadlock when executing
tests in parallel, not because there are errors.

When a build fails, the release workflow is
blocked, preventing us to release.

We decided to disable the Windows build temporarily
to enable releases.

related to: #324
PR: #342
---


aepfli (approved):

dwalluck:
Perhaps it would be better to disable running tests in parallel on Windows, so at least they could run and catch other potential issues until this is fixed.

PR #362, opened 2020-10-19
CartesianProductTest × RangeSources

### LINKED ISSUE #354: Make `@CartesianProductTest` work with range sources
It seems to me that `@CartesianProductTest` may work pretty well with the range sources - e.,g.:

```java
@CartesianProductTest(name = "word: {0} / number: {1}")
@CartesianValueSource(strings = { "foo", "bar"})
@IntRangeSource(from = 0, to = 10, step = 2)
void combination(String word, int number) { }
```

What do you think, @Michael1993? Let me know of you think this is worth exploring and whether you have time for it.

Michael1993:
Closes #354 

---
**PR checklist**

Documentation (general)
* [x] Javadoc uses formal style, while sites documentation may use informal style (see #265)

Documentation (new extension)

Code
* [x] There are checks which validate correct / false usage / configuration of a functionality and there are tests to verify those checks (see #164)

Contributing

---


Michael1993:
Not having 100% code coverage because of "this should never happen" exceptions. (╯°□°)╯︵ ┻━┻

Michael1993:
TODO:
* CartesianValueSource is now annotated with `@ArgumentsSource`, this might lead to weird behaviour with `@ParameterizedTest`
* Preferably all range sources should have a test
* Documentation should reflect the change

docs/cartesian-product.adoc
```diff
@@ -25,7 +25,7 @@ The extension takes the test data parameter values and runs the test for every p
 You can supply test parameters to `@CartesianProductTest` in multiple ways.
 
  - The annotation can have a `String[]` value (see <<Supplying CartesianProductTest with a `String[]`>>)
- - the test method can be annotated with `@CartesianValueSource` (see <<Annotating your test method with @CartesianValueSource>>)
+ - the test method can be annotated with `@CartesianValueSource` or range source annotations (see <<Annotating your test method>>)
  - the test class can have a static factory method providing the arguments (see <<Writing a static factory method for the parameters>>)
```
Bukama, line 29: First line has an upper `T` the other two a lower one.

docs/cartesian-product.adoc
```diff
@@ -78,7 +78,7 @@ To demonstrate with a table:
 | 8th test  | "1"            | "1"            | "1"
 |===
 
-== Annotating your test method with @CartesianValueSource
+== Annotating your test method
 
 If you don't only supply string values to your test method like in the example above, you can annotate your method with `@CartesianValueSource`.
 `@CartesianValueSource` is used to define the possible inputs of a single test parameter - as annotations are listed top-to-bottom, they provide parameter values left-to-right.
```
Bukama, line 84: A bit below there is link to ValueSources of JUnit (can't comment that unchanged line), but the link only opens the user guide but not the anker of that sub topic

docs/cartesian-product.adoc
```diff
@@ -113,6 +113,38 @@ To demonstrate with a table:
 | 6th test  | 4                 | "B"
 |===
 
+Alternatively, you can annotate your test method with link:range-sources.adoc[range source annotations].
```
Bukama, line 116: Note: We have to check this link (and the one below) after site deployment, as of now in the PR it points to `https://github.com/junit-pioneer/junit-pioneer/blob/d0ef6d0e6d307246a54747055866f12973d9971c/docs/range-sources.adoc` but I know that it some kind of relative link.

Bukama (changes requested):
As already mentioned in stream chat yesterday this looks quite good to me. Some minor things about the docs.

One maybe bigger thing we should investigate before merging this: The `@RangeSources` are now repeatable.  But I don't see any changes - not even tests - how the RangeSource itself handles them if there are multiple annotations (which is technically possible now). You wrote in the docs to not do it, but we should have investigated this a bit more to know whats happen if anyone does that.

docs/cartesian-product.adoc
Michael1993: https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests-sources-ValueSource This is the link. It absolutely _should_ open the user guide at the proper anchor.

Bukama:
HAHA I'm so stupid. I wanted to write a commit message for the PioneerAssertion, but wrote one for this 🤯 

As I already wrote it leave it here for maybe further usage:

```
Make CartesianProductTest work with RangeSources (#354 / #362)

A CartesianProduct is the set of all ordered pairs of elements from at least two sets.
The CartesianProductExtension is already able to create this set from multiple different sources, e.g. CartesianValueSource.

The RangeSourceExtension creates multiple parameters by taking an interval and a step size, e.g. create every from from 1 to 10 with step size 1.
It's available for multiple variable types, including byte, short and more.

Combining these extensions would be an easy way to provide each combination of parameters from multiple intervals.
Up to this PR the combination was not possible.
Main cause for this was that the annotations of the RangeSourceExtension were not repeatable, so it was not possible two ranges to a test method and then let the CartesianProductExtension create the CartesianProduct to use all combination of inputs from both ranges.

This PRs adds repeatable range source annotations and extends the CartesianProductExtension to build the product of the given intervals.

closes #354
PR: #362
```

nipafx:
Couple of things that need to be done and might as well be part of this PR:

* even though [you're now a maintainer](https://github.com/junit-pioneer/junit-pioneer/commit/a07a28b024c7d24ff5aaae70ffdcb1f5fbb2ab4c), you still get to credit yourself for this extension in the README
* add `@since` annotations to the existing Cartesian product API
* add Javadoc to `CartesianProductTest.Sets`
* add a test to `CartesianProductTest.Sets#addAll(Iterable)`

docs/cartesian-product.adoc
```diff
@@ -113,6 +113,38 @@ To demonstrate with a table:
...
+For _this purpose only_, range sources are repeatable.
+When you annotate your `@CartesianProductTest` with a range source, you specify the inputs to a single test parameter.
+This is not like `@ParameterizedTest` where you have to define all inputs in a single `@ArgumentsSource`.
```
nipafx, line 119:
```suggestion
This is _not_ like `@ParameterizedTest`, where you have to define all inputs in a single `@ArgumentsSource`.
```

😁

docs/cartesian-product.adoc
```diff
@@ -113,6 +113,38 @@ To demonstrate with a table:
...
+
+```java
+@CartesianProductTest
+@ShortRangeSource(from = 1, to = 3, step = 1)
+@LongRangeSource(from = 0L, to = 2L, step = 1, closed = true)
+void testShortAndLong(short s, long l) {
+    // passing test code
+}
+```
```
nipafx, line 128: Asciidoc source blocks and tab indentation, please. :)

src/main/java/org/junitpioneer/jupiter/params/ByteRangeSource.java
```diff
@@ -36,6 +37,7 @@
 @Documented
 @ArgumentsSource(RangeSourceArgumentsProvider.class)
 @RangeClass(ByteRange.class)
+@Repeatable(ByteRangeSource.ByteRangeSources.class)
```
nipafx, line 40: As these annotations are now repeatable and can be used with Cartesian products, I think they need some more changes:

* explain their dual use in their Javadoc (including a link to Cartesian product docu)
* explain where they can be repeated and where they can't

src/main/java/org/junitpioneer/jupiter/params/RangeSourceArgumentsProvider.java
```diff
@@ -63,17 +78,16 @@
 			throw new IllegalArgumentException(message);
```
nipafx, line 78: This message may be more helpful if we explain why, even though the annotation is repeatable, it can only occur once on `@ParameterizedTests`.

src/test/java/org/junitpioneer/jupiter/CartesianProductTestExtensionTests.java
```diff
@@ -590,4 +671,41 @@ void distinctInputsFactory(TimeUnit unit, String string) {
...
 	}
 
+	static class ArgumentsSourceTestCases {
+
+		@CartesianProductTest
+		@IntRangeSource(from = 1, to = 4, closed = true)
+		@IntRangeSource(from = 2, to = 4, step = 2, closed = true)
+		@ReportEntry("{0},{1}")
+		void basicIntRangeSource(int i, int j) {
+		}
```
nipafx, line 681: This looks great!

src/test/java/org/junitpioneer/jupiter/CartesianProductTestExtensionTests.java
```diff
@@ -590,4 +671,41 @@ void distinctInputsFactory(TimeUnit unit, String string) {
...
+
+		@CartesianProductTest
+		@IntRangeSource(from = 0, to = 4)
+		@ValueSource(ints = { 2, 4 })
```
nipafx, line 685: If this works with `@CartesianValueSource`, there should be a test for that.

src/main/java/org/junitpioneer/jupiter/PioneerAnnotationUtils.java
```diff
@@ -117,6 +121,39 @@ public static boolean isAnyRepeatableAnnotationPresent(ExtensionContext context,
...
+	 */
+	public static <A extends Annotation> List<Annotation> findAnnotatedAnnotations(AnnotatedElement element,
+			Class<A> annotation) {
+		return Arrays
+				.stream(element.getDeclaredAnnotations())
+				// flatten @Repeatable aggregator annotations
+				.flatMap(PioneerAnnotationUtils::flatten)
+				.filter(a -> !(findOnElement(a.annotationType(), annotation, true).isEmpty()))
+				.collect(Collectors.toList());
+	}
```
nipafx, line 137: Without looking into this, I'm a but surprised that there's a combination of annotation search parameters that we don't already cover. Can you help me understand how this is different from the other methods?

src/main/java/org/junitpioneer/jupiter/CartesianValueArgumentsProvider.java
```diff
@@ -13,28 +13,29 @@
...
 
 	private Object[] arguments;
 
 	@Override
 	public void accept(CartesianValueSource source) {
 		// @formatter:off
 		List<Object> arrays =
-				// Declaration of <Object> is necessary due to a bug in Eclipse Photon.
-				Stream.<Object> of(
+				Stream.of(
```
nipafx, line 38: Did you check newer Eclipse versions work with this?

src/main/java/org/junitpioneer/jupiter/CartesianProductTestExtension.java
```diff
@@ -94,17 +102,50 @@ private static void ensureNoInputConflicts(CartesianProductTest annotation,
...
 	}
 
+	private List<Object> getSetFromAnnotation(ExtensionContext context, Annotation source) {
+		try {
+			ArgumentsProvider provider = initializeArgumentsProvider(source);
+			return provideArguments(context, source, provider);
+		}
+		catch (Exception e) {
+			throw new ExtensionConfigurationException("Could not provide arguments because of exception.", e);
+		}
```
nipafx, line 121:
```suggestion
		catch (Exception ex) {
			throw new ExtensionConfigurationException("Could not provide arguments because of exception.", ex);
		}
```

src/main/java/org/junitpioneer/jupiter/CartesianProductTestExtension.java
```diff
@@ -94,17 +102,50 @@ private static void ensureNoInputConflicts(CartesianProductTest annotation,
...
-			provider.accept(source);
-			List<Object> collect = provider.provideArguments().distinct().collect(toList());
-			sets.add(collect);
+		for (Annotation source : argumentsSources) {
+			sets.add(getSetFromAnnotation(context, source));
 		}
 		return sets;
 	}
 
+	private List<Object> getSetFromAnnotation(ExtensionContext context, Annotation source) {
```
nipafx, line 114: Is `Annotation source` a `...Rangesource`? If so, can you rename the variable(s) accordingly?

src/main/java/org/junitpioneer/jupiter/CartesianProductTestExtension.java
```diff
@@ -94,17 +102,50 @@ private static void ensureNoInputConflicts(CartesianProductTest annotation,
...
+				.findAnnotation(source.annotationType(), ArgumentsSource.class)
+				// never happens, we already know these annotations are annotated with @ArgumentsSource
+				.orElseThrow(() -> new PreconditionViolationException(format(
+					"%s was not annotated with @ArgumentsSource but should have been.", source.annotationType())));
+		return ReflectionSupport.newInstance(providerAnnotation.value());
+	}
+
+	private List<Object> provideArguments(ExtensionContext context, Annotation source, ArgumentsProvider provider)
+			throws Exception {
+		if (Arrays.asList(provider.getClass().getInterfaces()).contains(Consumer.class)) {
```
nipafx, line 135: This should be...

```suggestion
		if (provider instanceof Consumer) {
```

... but, it's not? (One test fails after the change.)

src/main/java/org/junitpioneer/jupiter/CartesianProductTestExtension.java
```diff
@@ -94,17 +102,50 @@ private static void ensureNoInputConflicts(CartesianProductTest annotation,
...
+					.provideArguments(context)
+					.map(Arguments::get)
+					.flatMap(Arrays::stream)
+					.distinct()
+					.collect(toList());
+		} else {
+			throw new PreconditionViolationException(
+				format("%s does not implement the Consumer<T> interface.", provider.getClass()));
+		}
+	}
```
nipafx, line 147: Note to self: Take a close look at this when you're more awake. :)

nipafx (commented):
Looks great, I really like how this turned out. Two things:

1. Please remove curly braces around one-liners. My eyes... the googles, they do nothing.

2. We need a test that makes sure `@ParameterizedTest` fails if a (now repeatable) value source is applied twice, e.g.:

```java
@LongRangeSource(from = 1L, to = 2L)
@LongRangeSource(from = 1L, to = 5L)
@ParameterizedTest
void twoAnnotations() { }
```

nipafx:
I've just heard, there may be a reason to introduce `CartesianAnnotationConsumer` - if there is, let's give it a shot. I like expressing ideas in types.

src/main/java/org/junitpioneer/jupiter/CartesianValueArgumentsProvider.java
Michael1993: Hm, I can't remember. I think I did, but I forgot to write it down. 🤦

src/main/java/org/junitpioneer/jupiter/CartesianProductTestExtension.java
Michael1993: Not a `RangeSource` - it is an `Annotation` that is itself annotated with `@ArgumentsSource`.

It works like this (basically copying `ParameterizedTest`, btw):
 - Your test is a `CartesianProductTest`
 - This test has annotations on it that are annotated with `@ArgumentsSource`
 - The `@ArgumentsSource` defines and `ArgumentsProvider` (class) that must implement `AnnotationConsumer<A extends Annotation>` (in our case this should be `CartesianAnnotationConsumer`)
 - We instantiate the `ArgumentsProvider` (must have no-args constructor)
 - We call the `accept(Annotation)` method with the _original_ annotation
 - We call `provideArguments(ExtensionContext context)`

Easy-peasy, right?

src/main/java/org/junitpioneer/jupiter/CartesianProductTestExtension.java
Michael1993: `ArgumentsProvider`s used by annotations belonging to JUnit's `@ParameterizedTest` (e.g.: `@ValueSource`) implement `Consumer` through `AnnotationConsumer`. We will fix this with defining `CartesianAnnotationConsumer`.

src/main/java/org/junitpioneer/jupiter/CartesianValueArgumentsProvider.java
Michael1993: I will revert it just to be safe - but we should check it sometime.

src/main/java/org/junitpioneer/jupiter/PioneerAnnotationUtils.java
Michael1993: No. 👿

Michael1993: JUnits `AnnotationSupport` and our own `PioneerAnnotationUtils` has a lot of methods for getting specific annotations _directly_. We can not get annotations based on a criteria. I wanted to get all annotations on a test method that are themselves annotated with `@ArgumentsSource`. So, that's what this does - returns all annotations on an `AnnotatedElement` that are *annotated with a given annotation*.

Bukama:
Can you give me a quick update on this. I can't see if Nicolai's requests are already in or not. Thank you.

Michael1993:
> Can you give me a quick update on this. I can't see if Nicolai's requests are already in or not. Thank you.

Nicolai needs to review it again. I implemented the things he mentioned but it was a late-night review, so he might have missed some things earlier.

scordio (commented):
Thank you @Michael1993 for your effort! I only noticed two minor points in the docs.

Looking forward to having this feature available.

docs/cartesian-product.adoc
```diff
@@ -119,6 +119,53 @@ To demonstrate with a table:
...
+----
+@CartesianProductTest
+@ShortRangeSource(from = 1, to = 3, step = 1)
+@LongRangeSource(from = 0L, to = 2L, step = 1, closed = true)
+void testShortAndLong(short s, long l) {
+	// passing test code
+}
+----
+
+Just like with `@CartesianValueSource` annotations get assigned to parameters in order.
```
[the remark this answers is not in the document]
Michael1993: Fixed. 👍

docs/cartesian-product.adoc
```diff
@@ -138,8 +185,7 @@ static CartesianProductTest.Sets nFold() {
 	return new CartesianProductTest.Sets()
 		// e.g. with a `List<String> letters`,
 		// you can use `addAll`:
-		// .addAll(letters)
-		.add("Alpha", "Omega")
+		// .addAll(letters).add("Alpha", "Omega")
```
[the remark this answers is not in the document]
Michael1993: Good catch! 👍 
Fixed.

src/main/java/org/junitpioneer/jupiter/CartesianProductTestExtension.java
Michael1993: In case this is not clear, here is an example:
```java
@ArgumentsSource(ExampleArgumentsProvider.class)
@interface ExampleSourceAnnotation {
   int intValue();
}

class ExampleArgumentsProvider implements ArgumentsProvider, CartesianAnnotationConsumer<ExampleSourceAnnotation> {
  int i;

  void accept(ExampleSourceAnnotation annotation) {
    /* do something with intValue() */
    i = annotation.intValue();
  }

  Stream<? extends Arguments> provideArguments(ExtensionContext context) {
    /* `accept` was already called so `i` should have a value */
    return Stream.of(Arguments.of(i));
  }
}
```

So, if we have a `CartesianProductTest`, annotated with `ExampleSourceAnnotation`, we store the annotation, instantiate `ExampleArgumentsProvider` (with no-arg constructor), 'feed it' the `ExampleSourceAnnotation` and call `provideArguments` to get the set for our test.

src/main/java/org/junitpioneer/jupiter/params/RangeSourceArgumentsProvider.java
```diff
@@ -41,20 +44,34 @@
...
+	private void initArgumentsSource(ExtensionContext context) {
+		// since it's a method annotation, the element will always be present
+		AnnotatedElement element = context.getElement().orElseThrow(IllegalStateException::new);
+
+		verifyNoContainerAnnotationIsPresent(element);
+		List<Annotation> argumentsSources = Stream
+				.of(element.getAnnotations())
+				.filter(annotations -> Arrays
+						.stream(annotations.annotationType().getAnnotationsByType(ArgumentsSource.class))
+						.anyMatch(annotation -> getClass().equals(annotation.value())))
```
nipafx, line 73: Will this find meta-annotations? I doubt it would. Does this make sense?

```java
@IntRangeSource(from = 0, to = 10)
public @interface FromZeroToTen { }

@CartesianProductTest
@FromZeroToTen
void test(int i) {
}
```

If it does, we need to make sure that we find such annotations. Please use/extend `org.junitpioneer.jupiter.params.PioneerAnnotationUtils` if it's suitable - as painful as that may be.

src/main/java/org/junitpioneer/jupiter/params/RangeSourceArgumentsProvider.java
```diff
@@ -63,17 +80,36 @@
...
+				.anyMatch(method -> method.getName().equals("value") && method.getReturnType().isArray()
+						&& method.getReturnType().getComponentType().isAnnotation()
+						&& declaresContainer(method.getReturnType().getComponentType(), annotation));
 	}
 
-	private Stream<?> asStream(Range r) {
+	private static boolean declaresContainer(Class<?> componentType, Annotation annotation) {
+		Repeatable repeatable = componentType.getAnnotation(Repeatable.class);
+		return repeatable != null && repeatable.value().equals(annotation.annotationType());
+	}
```
nipafx, line 104: I just saw this code in `PioneerAnnotationUtils`, no need to repeat it here. Just expose there and add (reflectively) to the `params` variant.

docs/cartesian-product.adoc
```diff
@@ -119,6 +119,53 @@ To demonstrate with a table:
...
+[source, java]
+----
+@CartesianProductTest
+@ShortRangeSource(from = 1, to = 3)
+@CartesianValueSource(strings = { "text one", "text two" })
+void mixAndMatchTest(short number, String text) {
+	// passing test code
+}
+----
+
```
nipafx, line 168: Please document how to create `ArgumentsProvider` (as promised in `CartesianAnnotationConsumer`).

nipafx (commented):
Looking good, we're getting there.

Bukama (approved):
I like the meaningfull names etc (and I don't like if-clauses without brackets, but I had this conversion with Nicolai some time ago, too...)

I keep my state that this looks good for me. I leave final thoughts to Nicolai

src/main/java/org/junitpioneer/jupiter/params/RangeSourceArgumentsProvider.java
Michael1993: All range sources have `@Target(ElementType.METHOD)` on them. No meta-annotations can exist.

nipafx: True. We should probably discuss on an issue first whether this makes sense.

nipafx:
Looks good to me, merge ahead! We still need a commit message, though, and I'll cowardly leave this to somebody else. 😊

nipafx:
Never mind, found @Bukama's. :)

PR #365, opened 2020-10-27
Add JUnit 5.6.3 to JUnit workflow (#364 / #365)

### LINKED ISSUE #364: Add JUnit 5.6.3 to JUnit workflow
JUnit 5.6.3 was releases yesterday. As it is a more recent Jupiter version as the one, we use in Pioneer it has to be added to the JUnit workflow.

Bukama:
```
JUnit 5.6.3 was releases yesterday. As it is a more recent,
but still supported, Jupiter version as the one we use in
Pioneer it has to be added to the JUnit workflow.

closes #364
PR: #365
```

### DIFF
--- a/.github/workflows/junit5-build.yml
+++ b/.github/workflows/junit5-build.yml
@@ -15,7 +15,7 @@ jobs:
     continue-on-error: true
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


beatngu13 (approved):
Thx for the addition!

Side note: Is there a rule we follow that determines when we _remove_ a certain JUnit version from our build matrix?

PR #372, opened 2020-11-01
Allow adding an existing collection to CartesianProduct$Sets

### LINKED ISSUE #358: Extend CartesianProductTest$Sets#add
Consider the case where one wants reuse a method inside several Cartesian product factory methods:

```java
static CartesianProductTest.Sets someFactoryMethod() {
    return new CartesianProductTest.Sets()
        .add(reusedMethod())
        .add(false, true);
}

static CartesianProductTest.Sets anotherFactoryMethod() {
    return new CartesianProductTest.Sets()
        .add(reusedMethod())
        .add("foo", "bar");
}

static Object[] reusedMethod() {
    return new Object[]{new Object(), null};
}
```

Since `CartesianProductTest$Sets#add` currently only accepts `Object...`, `reusedMethod` must return an array.

I think it would be helpful to have more ways to `add` to a Cartesian set, e.g.:

```java
public Sets add(Object... entries) {
    sets.add(toDistinctList(Arrays.stream(entries)));
    return this;
}
 
public Sets add(Collection<Object> entries) {
    sets.add(toDistinctList(entries.stream()));
    return this;
}

private static List<Object> toDistinctList(Stream<Object> stream) {
    return stream.distinct().collect(toList());
}
```

I have already created a branch in my fork for this, but beforehand I would like hear what's your say on this? And/or do you have other `add` signatures in mind?

beatngu13:
Closes #358.

Proposed commit message:

```
Allow adding an existing collection to CartesianProduct$Sets (#358 / #372)

`CartesianProductTest$Sets#add` currently only accepts `Object`
varargs, which is why methods used within Cartesian product factory
methods must return an array. This PR introduces
`CartesianProductTest$Sets#addAll` to allow adding an existing
collection or stream.

Closes: #358 
PR: #372
```

---
**PR checklist**

Documentation (general)

Documentation (new extension)

Code

Contributing

---


beatngu13:
Wonder why Sonar complains that [these lines](https://sonarcloud.io/code?id=junit-pioneer_junit-pioneer&pullRequest=372&selected=junit-pioneer_junit-pioneer%3Asrc%2Fmain%2Fjava%2Forg%2Fjunitpioneer%2Fjupiter%2FCartesianProductTest.java&line=82) are not covered. 🤔

src/test/java/org/junitpioneer/jupiter/CartesianProductTestSetsTest.java
```diff
@@ -0,0 +1,42 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.util.Lists.list;
+
+import java.util.List;
+
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+@DisplayName("The sets for CartesianProductTest")
+class CartesianProductTestSetsTest {
```
Michael1993, line 22: Personally, I'd put these in a `@Nested` class in `CartesianProductTestExtensionTests`.

Michael1993: Also, this test never runs because we only run test classes that match the pattern: `.*Tests`.

src/main/java/org/junitpioneer/jupiter/CartesianProductTest.java
```diff
@@ -72,10 +74,19 @@
 		private final List<List<?>> sets = new ArrayList<>(); //NOSONAR
 
 		public Sets add(Object... entries) {
-			sets.add(new ArrayList<>(Arrays.stream(entries).distinct().collect(toList())));
+			sets.add(toDistinctList(Arrays.stream(entries)));
 			return this;
 		}
 
+		public Sets addAll(Collection<?> entries) {
```
Michael1993, line 81: Didn't you say we should be able to add a `Stream` too...?

Michael1993 (changes requested):
Promising, but at least add the missing `s`?

src/test/java/org/junitpioneer/jupiter/CartesianProductTestSetsTest.java
beatngu13: Ah, thank you! Never realized this. This is also the reason why Sonar isn't happy.

Regarding `@Nested`: `CartesianProductTestExtensionTests` is already quite big, so I thought having a separate class helps. But I can change that.

src/main/java/org/junitpioneer/jupiter/CartesianProductTest.java
beatngu13: Uh, no? 🤷‍♂️ You mean `addAll(Stream<?>)` in addition?

Michael1993: Yes, the option is already there, just make it `public` and rename it to `add`/`addAll`. 😄

src/test/java/org/junitpioneer/jupiter/CartesianProductTestSetsTest.java
beatngu13: @Michael1993 converted the test to a nested one. We "now" have some redundant test cases since e.g. `RedundantInputSetTestCases` does essentially the same, just via a different test interface.

Michael1993 (approved):

src/test/java/org/junitpioneer/jupiter/CartesianProductTestSetsTest.java
Bukama: > Regarding `@Nested`: `CartesianProductTestExtensionTests` is already quite big, so I thought having a separate class helps. But I can change that.

I'm totally fine with splitting large test classes into multiple files and don't overuse `@Nested`.

Bukama: > @Michael1993 converted the test to a nested one. We "now" have some redundant test cases since e.g. `RedundantInputSetTestCases` does essentially the same, just via a different test interface.

When they use different interfaces they go different ways, don't they?

Bukama (approved):
LGTM 👍

src/main/java/org/junitpioneer/jupiter/CartesianProductTest.java
Bukama: Should we add the possibility to documentation, with a small code example?

nipafx:
[Once again](https://github.com/junit-pioneer/junit-pioneer/pull/361#issuecomment-733206479), screw Sonar. 😁

PR #388, opened 2020-12-24
An alternative solution to PioneerAnnotationUtils (#389 / #388)

### LINKED ISSUE #389: Handle duplication of PioneerAnnotationUtils
_by @Michael1993 _

PioneerAnnotationUtils is currently not public, but is used by
multiple classes in different packages. Previously, we created
a copy of the class with reflection magic, which is a fragile
approach. This PR introduces a test, that checks (and copies) the
PioneerAnnotationUtils in different packages. The test fails if
the classes don't match.

Michael1993:
closes #389

Proposed commit message:

```
Create test-time copy of PioneerAnnotationUtils (#389 / #388)

PioneerAnnotationUtils is currently not public, but is used by
multiple classes in different packages. Previously, we created
a copy of the class with reflection magic, which is a fragile
approach. This PR introduces a test, that checks (and copies) the
PioneerAnnotationUtils in different packages. The test fails if
the classes don't match.

closes: #389
PR: #388
```

---
**PR checklist**

Code
* [X] There are checks which validate correct / false usage / configuration of a functionality and there are tests to verify those checks
* [X] Tests use [AssertJ](https://joel-costigliola.github.io/assertj/) or our own [PioneerAssert](https://github.com/junit-pioneer/junit-pioneer/blob/master/CONTRIBUTING.md#assertions) (which are based on AssertJ)

Contributing

---


Michael1993:
Duplication is expected, code coverage should not be considered for duplicate class - or we can duplicate the tests too, I guess. 🤔

Bukama:
I‘ll review, when I‘m back home (mobile is too small and the app sucks). But one thing I can tell you right now that Nicolai will bring up: We need an issue for that topic :)

Bukama (commented):
I don't like the that we copy our own classes only because they are not public.

I vote to make them public an use them in the way classes are used to. 

Yes without module system this opens the class for others, but why not? If anyone wants to use it to find annotations on their own extension, why not?

nipafx:
I'm sorry-not-sorry, but I'll close this one, ok? :)

PR #403, opened 2021-01-12
Move Pioneer utils to new package `internal`

nipafx:
Proposed commit message:

```
Move internal utility classes to new package `internal` (#389 / #403)

Requirements for `PioneerUtils` and `PioneerAnnotationUtils`:

* reachable within Pioneer from different packages
* as unreachable as possible for Pioneer users

The former approach was to have one original version of the class and
one duplicate with the same method signatures, but without the bodies.
Instead reflection was used to execute the method from the original.
Felt like an eldritch horror to some.

The most unsurprising and less weird approach is to put the classes
into their own (non-exported) package and make them public. This still
allows access (unless the module system is used), but that could be
warned about.

This change introduces the new package `internal` for these two classes
and adds Javadoc warning against their use.

Closes: #389
PR: #403
```

---


Michael1993 (approved):
Could you improve the test coverage as well? 👼

Bukama (changes requested):
The `module-info.java` needs to be updated for Java 9+ users :)

Michael1993:
> The `module-info.java` needs to be updated for Java 9+ users :)

Wait, why? The whole reason for moving to `org.junitpioneer.internal` was to _not_ export it.

Michael1993 (changes requested):
Wait, shouldn't this PR include the deletion of the reflection-based solution? 🤔

Michael1993 (approved):
I blame my blindness on my glasses. Disregard my previous ignorance. LGTM. 🤓

nipafx:
> The `module-info.java` needs to be updated for Java 9+ users :)

I don't think it should, the package shouldn't be accessible, so it's not eyported.

Bukama:
> > The `module-info.java` needs to be updated for Java 9+ users :)
> 
> Wait, why? The whole reason for moving to `org.junitpioneer.internal` was to _not_ export it.

Aehm of course... #brainlag

Bukama (approved):

PR #404, opened 2021-01-12
Improve documentation/logging of reflective warning for environment variable extensions

nipafx:
Proposed commit message:

```
More clearly document/log env. var. reflective warning (#387 / #404)

The environment variable extension uses reflection to change the
otherwise immutable variables. On Java 9+, this leads to warnings or
errors and this change makes it easier for users to understand the
reason behind that:

* Before this change, we printed a related warning via report entries,
  but those are often lost in the wild. We now log an additional
  warning to `System.out`. While we initially considered `System.err`
  it is rather invasive, particularly since, with good configuration,
  the module system won't emit a warning and hence it's only Pioneer
  polluting `System.err` - not good.
* This change also expands the documentation by including the exact
  warning/error messages, how to prevent them, and links to build tool
  solutions.

Closes: #387
PR: #404
```

---

### DIFF
--- a/src/main/java/org/junitpioneer/jupiter/SetEnvironmentVariable.java
+++ b/src/main/java/org/junitpioneer/jupiter/SetEnvironmentVariable.java
@@ -35,9 +35,9 @@ import org.junit.jupiter.api.extension.ExtendWith;
  * uses reflection to change them. This requires that the {@link SecurityManager}
  * allows modifications and can potentially break on different operating systems and
  * Java versions. Be aware that this is a fragile solution and consider finding a
- * better one for your specific situation. If you're running on Java 9 or later, you
- * may have to add {@code --add-opens=java.base/java.util=ALL-UNNAMED} to your test
- * execution to prevent warnings or even errors.</p>
+ * better one for your specific situation. If you're running on Java 9 or later and
+ * are encountering warnings or errors, check
+ * <a href="https://junit-pioneer.org/docs/environment-variables/">the documentation</a>.</p>
  *
  * <p>During
  * <a href="https://junit.org/junit5/docs/current/user-guide/#writing-tests-parallel-execution" target="_top">parallel test execution</a>,
--- a/src/main/java/org/junitpioneer/jupiter/SetEnvironmentVariable.java
+++ b/src/main/java/org/junitpioneer/jupiter/SetEnvironmentVariable.java
@@ -37,7 +37,7 @@ import org.junit.jupiter.api.extension.ExtendWith;
  * Java versions. Be aware that this is a fragile solution and consider finding a
  * better one for your specific situation. If you're running on Java 9 or later and
  * are encountering warnings or errors, check
- * <a href="https://junit-pioneer.org/docs/environment-variables/">the documentation</a>.</p>
+ * <a href="https://junit-pioneer.org/docs/environment-variables/#warnings-for-reflective-access">the documentation</a>.</p>
  *
  * <p>During
  * <a href="https://junit.org/junit5/docs/current/user-guide/#writing-tests-parallel-execution" target="_top">parallel test execution</a>,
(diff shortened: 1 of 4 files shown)


src/test/java/org/junitpioneer/jupiter/EnvironmentVariableExtensionTests.java
```diff
@@ -255,17 +255,21 @@ void resetWarning() {
 		}
 
 		@Test
-		void shouldNotReportWarningIfExtensionNotUsed() {
+		@StdIo
+		void shouldNotReportWarningIfExtensionNotUsed(StdOut out) {
 			ExecutionResults results = executeTestMethod(ReportWarningTestCases.class, "testWithoutExtension");
 
 			assertThat(results).hasNoReportEntries();
+			assertThat((out.capturedLines())).containsExactly("");
```
Michael1993, line 263: Why the extra parentheses?

src/test/java/org/junitpioneer/jupiter/EnvironmentVariableExtensionTests.java
```diff
@@ -255,17 +255,21 @@ void resetWarning() {
...
 		}
 
 		@Test
-		void shouldReportWarningIfExtensionUsed() {
+		@StdIo
+		void shouldReportWarningIfExtensionUsed(StdOut out) {
 			ExecutionResults results = executeTestMethod(ReportWarningTestCases.class, "testWithExtension");
 
 			assertThat(results).hasSingleReportEntry().withKeyAndValue(WARNING_KEY, WARNING_VALUE);
+			assertThat((out.capturedLines())).containsExactly(WARNING_KEY + ": " + WARNING_VALUE);
```
Michael1993, line 272: Why the extra parentheses?

src/main/java/org/junitpioneer/jupiter/SetEnvironmentVariable.java
```diff
@@ -35,9 +35,9 @@
  * uses reflection to change them. This requires that the {@link SecurityManager}
  * allows modifications and can potentially break on different operating systems and
  * Java versions. Be aware that this is a fragile solution and consider finding a
- * better one for your specific situation. If you're running on Java 9 or later, you
- * may have to add {@code --add-opens=java.base/java.util=ALL-UNNAMED} to your test
- * execution to prevent warnings or even errors.</p>
+ * better one for your specific situation. If you're running on Java 9 or later and
+ * are encountering warnings or errors, check
+ * <a href="https://junit-pioneer.org/docs/environment-variables/">the documentation</a>.</p>
```
Michael1993, line 40: Could we link to the specific section? Would that be better?

Michael1993 (changes requested):
Looks good, apart from a few minor things.

Bukama (changes requested):
Same as @Michael1993 says :)

src/test/java/org/junitpioneer/jupiter/EnvironmentVariableExtensionTests.java
nipafx: Er... _checks list of excuses_ ... readability?

src/test/java/org/junitpioneer/jupiter/EnvironmentVariableExtensionTests.java
nipafx: This one is easy: copy-paste!

src/main/java/org/junitpioneer/jupiter/SetEnvironmentVariable.java
nipafx: Will do.

Michael1993 (approved):

Bukama (approved):
LGTM, thanks for the update!

Issue #379, opened 2020-12-03
Add @CartesianEnumSource


scordio:
I am currently looking for a way to provide Enum values as input of a cartesian product test.

I understand it can be achieved with a [static factory method](https://junit-pioneer.org/docs/cartesian-product/#writing-a-static-factory-method-for-the-parameters), but having something like a `@CartesianEnumSource` which mimics the JUnit `@EnumSource` including `names` and `mode` parameters would be really nice.

If this sounds reasonable, I am happy to give it a shot (probably taking inspiration from #321).

----

Bonus topic: it would be also nice to have a `@CartesianArgumentSource` which would allow plugging in a custom `ArgumentProvider` implementation. Maybe this deserves a separate issue, but I'm not sure yet about its feasibility.

Michael1993:
How `CartesianProductTest` interacts with arguments sources [will change soon](https://github.com/junit-pioneer/junit-pioneer/pull/362). Once that is done, I will update the documentation to reflect these changes - i.e.: How you can write your own `ArgumentsProvider` for `CartesianProductTest` using API provided by Pioneer (it's almost like with JUnit just slightly different interfaces). This should also make implementing `@CartesianEnumSource` fairly trivial.

The issue as it is, is kind of blocked by the coming change, unfortunately. If you don't need this feature urgently, this will probably be implemented in about a month or so. Does that work?

scordio:
Thanks for the quick feedback, good to know that something in this direction was already on the table 🙂

I would definitely wait for the change, but please do let me know if there is anything that can be picked up to speed it up.

Michael1993:
Feel free to [review the PR](https://github.com/junit-pioneer/junit-pioneer/pull/362) implementing the change. All feedback is appreciated.

scordio:
@Michael1993 now that #362 is merged, I would be happy to put something together to support `@CartesianEnumSource`, just let me know if I can go ahead with it.

Michael1993:
Sure, feel free to put something together!

PR #409, opened 2021-01-15
Add `@CartesianEnumSource`

scordio:
Closes: #379

---
Proposed commit message:

```
Add `@CartesianEnumSource` (#379 / #409)

The `@CartesianEnumSource` annotation adds support for Enum constants 
to the Cartesian product test extension, mimicking the behavior of
JUnit's `@EnumSource`.

Closes: #379
PR: #409
```

---
**PR checklist**

Documentation (general)

Documentation (new extension)

Code

Contributing

---


Michael1993 (commented):
Looking good so far. Just a single suggestion from my side.

src/main/java/org/junitpioneer/jupiter/CartesianEnumSource.java
```diff
@@ -0,0 +1,183 @@
...
+		private final BiConsumer<CartesianEnumSource, Set<String>> validator;
+		private final BiPredicate<String, Set<String>> selector;
+
+		Mode(BiConsumer<CartesianEnumSource, Set<String>> validator, BiPredicate<String, Set<String>> selector) {
+			this.validator = validator;
+			this.selector = selector;
+		}
+
+		void validate(CartesianEnumSource enumSource, Set<String> names) {
+			notNull(enumSource, "EnumSource must not be null");
```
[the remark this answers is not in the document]
Michael1993: Probably should have tests for `null` values.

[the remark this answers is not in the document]
scordio: > Feel free to use `Objects.requireNonNull(Object, String)` or `Preconditions.notNull(Object, String)`.

Originally I used `org.junit.platform.commons.util.Preconditions.notNull` as it's done in the original `@EnumSource`, but the module-build failed as that package is not exported so I created the custom `notNull(Object, String)`.

I also thought about `Objects.requireNonNull` but I was not sure if the `NullPointerException` would have been a good replacement for `PreconditionViolationException`. It might be the latter is picked up in JUnit internals during the execution, but to be honest I didn't look deeply into that and I tried to maintain the style of sibling features.

However, the `notNull` check is invoked only by `CartesianEnumSource.Mode` methods which have package-private visibility and the input values are generated in the `CartesianEnumArgumentsProvider`, so with the current implementation there is no real chance that they can be `null` and the exception is thrown.

What do you think, shall I use `Objects.requireNonNull` instead?

> Probably should have tests for `null` values.

That's tricky because of the point above, the input values are generated by `CartesianEnumArgumentsProvider#accept` and, as far as I can see, there is no way they can be `null` with any behavioral test case.

However, I can create separate unit tests to cover them. Would you suggest going this way?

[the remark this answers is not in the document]
Michael1993: If it can't be `null` then it does not have to be checked for `null`. Or, alternatively, it does not matter what you use to check for `null`.

You could do:
```java
validator.accept(requireNonNull(enumSource), requireNonNull(names));
```

So compact. 😄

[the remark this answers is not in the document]
scordio: Theoretically, it makes sense to do null check because it's a package-private api, potentially callable by any other class in the same package. However, I see it as very unlikely.

I guess it's fine if we drop the `PreconditionViolationException` in favor of `NullPointerException`.

> You could do:
> 
> ```java
> validator.accept(requireNonNull(enumSource), requireNonNull(names));
> ```
> 
> So compact. 😄

And this should also make Sonar happy 😄

Michael1993:
Do not forget to update the documentation and the README.md.

Michael1993:
Please fix the code smell. Either by accepting the suggestion from Sonar or by adding `//NOSONAR` and an explanation. E.g.:
```java
//NOSONAR using wildcards is not a problem here because [...]
```

Note that it **must** be `//NOSONAR` and **not** `// NOSONAR` (no space).

docs/cartesian-product.adoc
```diff
@@ -119,6 +121,52 @@ To demonstrate with a table:
...
+}
+----
+
+Like `@CartesianValueSource`, `@CartesianEnumSource` is an annotation that might look familiar - it mimics JUnit's `@EnumSource`, except `@CartesianEnumSource` is repeatable.
+It also does NOT work with `@ParameterizedTest`.
+https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests-sources-EnumSource::[You can read about `@EnumSource` here.]
+
+The test `testEnumValues` is executed exactly six times.
+The first parameter can have any of the three constants `ONE`, `TWO` or `THREE`.
+The second parameter can have any of the two constants `BETA` or `GAMMA` (note the `EXCLUDE` mode applied to the other two constants).
```
Michael1993, line 153: This is unfortunately not enough. You have to talk about the configuration options in a bit more detail.

Michael1993 (changes requested):
Looks very good! Thank you!

Just expand the documentation a bit more, please.

docs/cartesian-product.adoc
scordio: Fair point, will do.

scordio:
> Please fix the code smell. Either by accepting the suggestion from Sonar or by adding `//NOSONAR` and an explanation. 

The suggestion from Sonar would trigger a raw type warning which can be suppressed. I would say not the first choice here.

I'm trying to figure out how to declare that Generic in a way to make Sonar happy, but I cannot see a proper way. Maybe Sonar is too aggressive when the subject is an Enum.

Any ideas?

Michael1993:
> > Please fix the code smell. Either by accepting the suggestion from Sonar or by adding `//NOSONAR` and an explanation.
> 
> The suggestion from Sonar would trigger a raw type warning, which can be suppressed but I would not consider it as the first choice.
> 
> I'm trying to figure out how to declare that Generic in a way to make Sonar happy, but I cannot see a proper way. Maybe Sonar is too aggressive when the subject is an Enum.
> 
> Any ideas?

Sonar won't be happy either way. Either you are using a raw generic without type parameters (major code smell!! oh no!!) or you are using generic wildcard in a return value (oh no!! major code smell!!). 🙄 

Feel free to use `Class<Enum>`,  `Class<Enum<?>>`, `Class<? extends Enum>`,  `Class<? extends Enum<?>>`, or  `Class<Enum<?extends Enum<?>>>` or whatever strikes your fancy and commenting `//NOSONAR` mentioning this is unavoidable.

Then again, I'm not an expert on generics so I'm just going to ping some other people.

@nipafx, @Bukama, @aepfli, @beatngu13 do you have any ideas?

scordio:
I'd not go far from how JUnit's `@EnumSource` is declared, so I vote for `//NOSONAR`

scordio:
> I wouldn't go far from the way JUnit's `@EnumSource` is declared, so I vote for `//NOSONAR`

On second thought, instead of polluting the annotation code, what about closing the issue on Sonar as [false positive](https://docs.sonarqube.org/latest/faq/)?

docs/cartesian-product.adoc
scordio: Done! It is mostly a shameless copy of the JUnit docs.

docs/cartesian-product.adoc
```diff
@@ -119,6 +121,110 @@ To demonstrate with a table:
...
+    assertNotNull(unit);
+}
+----
+
+Like `@CartesianValueSource`, `@CartesianEnumSource` is an annotation that might look familiar - it mimics JUnit's `@EnumSource`, except `@CartesianEnumSource` is repeatable.
+It also does NOT work with `@ParameterizedTest`.
+https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests-sources-EnumSource::[You can read about `@EnumSource` here.]
+
+The annotation’s `value` attribute is optional.
+When omitted, the declared type of the first method parameter is used.
```
Michael1993, line 142: Interesting. Consider the following test:
```java
@CartesianProductTest
@CartesianValueSource(ints = { 1, 2, 3 })
@CartesianEnumSource
void moreParams(int i, ChronoUnit cUnit) {
  // some test code
}
```

Would that work? Would you expect it to work?

docs/cartesian-product.adoc
```diff
@@ -119,6 +121,110 @@ To demonstrate with a table:
...
+=== `@CartesianEnumSource`
+
+`@CartesianEnumSource` provides a convenient way to use `Enum` constants.
+
+[source,java]
+----
+@CartesianProductTest
+@CartesianEnumSource(ChronoUnit.class)
+void testWithEnumSource(TemporalUnit unit) {
+    assertNotNull(unit);
```
Michael1993, line 133: Prefer empty test block with comment if possible (so it's consistent with the existing documentation). Otherwise prefer AssertJ (so it's consistent with other Pioneer documentation).

docs/cartesian-product.adoc
```diff
@@ -119,6 +121,110 @@ To demonstrate with a table:
...
+When omitted, the declared type of the first method parameter is used.
+The test will fail if it does not reference an `Enum` type.
+Thus, the `value` attribute is required in the above example because the method parameter is declared as `TemporalUnit`, i.e. the interface implemented by `ChronoUnit`, which isn’t an enum type.
+Changing the method parameter type to `ChronoUnit` allows you to omit the explicit enum type from the annotation as follows.
+
+[source,java]
+----
+@CartesianProductTest
+@CartesianEnumSource
+void testWithCartesianEnumSourceWithAutoDetection(ChronoUnit unit) {
```
Michael1993, line 151: Could you add a unit test for this?

Michael1993 (changes requested):
Looks very good. I still have some questions.

docs/cartesian-product.adoc
scordio: Uhm, good point. I try to see what happens with a new test case.

docs/cartesian-product.adoc
scordio: I thought I did it together with the last doc changes, but I clearly forgot one point 🤦 

https://github.com/junit-pioneer/junit-pioneer/blob/273035edb6c9fe7cb63a69932b5eaf58545639cc/src/test/java/org/junitpioneer/jupiter/CartesianProductTestExtensionTests.java#L934-L938

docs/cartesian-product.adoc
scordio: Empty block test should be enough

docs/cartesian-product.adoc
scordio: Most importantly, the docs were inspired by the latest version of JUnit (5.7), while the implementation comes from the version included by junit-pioneer (5.5)... still some work is needed :thinking:

docs/cartesian-product.adoc
scordio: This also requires aligning the implementation with the latest JUnit 5.7

scordio: I think I need to be a bit creative to keep this feature as JUnit specifically refers to the first parameter of the test method (which makes sense in the context of `@ParameterizedTest`). How I would like to see it is that the type is inferred based on the position of the annotation. E.g., having:

```java
@CartesianProductTest
@CartesianValueSource(ints = { 1, 2, 3 })
@CartesianEnumSource
@CartesianValueSource(ints = { 4, 5, 6 })
@CartesianEnumSource
void moreParams(int i, ChronoUnit cUnit, int j, MyEnum myEnum) {
  // some test code
}
```

the first `@CartesianEnumSource` should provide all the values of `ChronoUnit`, while the second should provide all the values of `MyEnum`.

Let's see if it's feasible.

scordio: @Michael1993 I spent some time on this topic and the change is not really straightforward. I propose to finalize this PR without the auto-detection feature and then having a follow-up issue to do this investigation. What do you think?

docs/cartesian-product.adoc
scordio: As mentioned in the other comment, I removed any reference to the auto-detection feature

docs/cartesian-product.adoc
scordio: Changed to AssertJ

src/main/java/org/junitpioneer/jupiter/CartesianEnumSource.java
```diff
@@ -0,0 +1,172 @@
...
+@ArgumentsSource(CartesianEnumArgumentsProvider.class)
+public @interface CartesianEnumSource {
+
+	/**
+	 * The enum type that serves as the source of the enum constants.
+	 *
+	 * @see #names
+	 * @see #mode
+	 */
+	Class<? extends Enum<?>> value();
```
scordio, line 63: Opening a thread not to leave this behind - this line causes a Sonar [code smell](https://sonarcloud.io/project/issues?id=junit-pioneer_junit-pioneer&issues=AXcIQv9wtsx-iI78REli&open=AXcIQv9wtsx-iI78REli&pullRequest=409). 

Instead of polluting the annotation with `//NOSONAR`, I suggest closing the issue on Sonar as [false positive](https://docs.sonarqube.org/latest/faq/).

Michael1993 (approved):
LGTM. I'll open an issue about the proposed improvements.

docs/cartesian-product.adoc
Michael1993: Closing this thread for now. Will open an issue about your proposed improvements.

Bukama (approved):

src/main/java/org/junitpioneer/jupiter/CartesianEnumSource.java
nipafx: I've just marked it as _won't fix_. I'm tempted to deactivate the rule as a whole - maybe if it pops up more often.

src/main/java/org/junitpioneer/jupiter/CartesianEnumSource.java
```diff
@@ -0,0 +1,172 @@
...
+	 *
+	 * <p>Defaults to {@link CartesianEnumSource.Mode#INCLUDE INCLUDE}.
+	 *
+	 * @see CartesianEnumSource.Mode#INCLUDE
+	 * @see CartesianEnumSource.Mode#EXCLUDE
+	 * @see CartesianEnumSource.Mode#MATCH_ALL
+	 * @see CartesianEnumSource.Mode#MATCH_ANY
+	 * @see #names
+	 */
+	CartesianEnumSource.Mode mode() default CartesianEnumSource.Mode.INCLUDE;
```
nipafx, line 91: Just for the record, I don't really like this way to structure the API - having one "general purpose" container (`names`) with an only half-way matching name (how do `names` relate to `MATCH_ALL`?) and then `mode` that defines how to process it, feels off to me. I'd prefer an API with straightforward attributes `String[] include()`, etc. (although that opens the can of what-if-multiple-attributes-are-set can of worms).

Anyway, since we want this annotation to mimic Jupiter's and it does the same thing, this is perfectly fine.

nipafx (approved):
Wow, I'm very impressed with how polished this PR is! I kinda wish mine look like that. 😆 Really good work, @scordio!

scordio:
Thanks for the nice words @nipafx! 🙂

Issue #408, opened 2021-01-13
Improve `@RetryingTest`: Offer minSuccess or similar for requiring multiple successful test runs


Michael1993:
Related to #355 

User should be able to specify an attribute that requires that `@RetryingTest` succeeds at least _N_ times or that at least _P_ percentage of all tests should succeed.
Note that the latter is not trivial, because it runs perpendicular to how `@RetryingTest` currently behaves.

Example:
```java
int count = 0;

@RetryingTest(value = 5, minSuccess = 2)
void minTwoSuccess() {
  count++;
  if (count != 1)
    Assertions.fail(); // will fail 4 times, making the test as a whole fail because `minSuccess` is 2
}
```

PR #430, opened 2021-02-18
Add `minSuccess` to RetryingTest

dump247:
The `minSuccess` field specifies the minimum number of times the test must
complete without an error.

Added a `maxAttempts` field as an alias for `value` so that the intent
is clearer. Only one of `maxAttempts` or `value` can be set. This does
change the behavior slightly in that the max attempt count is no longer
required. The change should be backward compatible
with existing usages though. If `maxAttempts` or `value` is not set,
the max attempts will default to `minSuccess`.

Closes: #408 

Proposed commit message:

```

${references}: ${issues}
PR: ${pull-request}
```

---
**PR checklist**

Documentation (general)

Documentation (new extension)

Code

Contributing

---

### DIFF
--- a/docs/retrying-test.adoc
+++ b/docs/retrying-test.adoc
@@ -57,6 +57,27 @@ The test `aborted` is aborted before (or during) its first execution.
 The first execution is marked as aborted/skipped, containing the underlying cause.
 The test suite as a whole is also marked as aborted/skipped.
 
+[source,java]
+----
+@RetryingTest(maxAttempts = 4, minSuccess = 2)
+void requiresTwoSuccesses() {
+	// test code that must complete successfully twice
+}
+----
+
+The test `requiresTwoSuccesses` must run at least two times without raising
+an exception. However, it will not run more than four times.
+
+[source,java]
+----
+@RetryingTest(minSuccess = 2)
+void runsTwice() {
+	// test code that must complete successfully twice
+}
+----
+
+The test `runsTwice` must run at least two times without raising an exception.
+
 == Combining `@RetryingTest` with `@Test` et al
 
 If `@RetryingTest` is combined with `@Test` or `TestTemplate`-based mechanisms (like `@RepeatedTest` or `@ParameterizedTest`), the test engine will execute it according to each annotation (i.e. more than once).
--- a/docs/retrying-test.adoc
+++ b/docs/retrying-test.adoc
@@ -68,16 +68,6 @@ void requiresTwoSuccesses() {
 The test `requiresTwoSuccesses` must run at least two times without raising an exception.
 However, it will not run more than four times.
 
-[source,java]
-----
-@RetryingTest(minSuccess = 2)
-void runsTwice() {
-	// test code that must complete successfully twice
-}
-----
-
-The test `runsTwice` must run at least two times without raising an exception.
-
 == Combining `@RetryingTest` with `@Test` et al
 
 If `@RetryingTest` is combined with `@Test` or `TestTemplate`-based mechanisms (like `@RepeatedTest` or `@ParameterizedTest`), the test engine will execute it according to each annotation (i.e. more than once).
--- a/docs/retrying-test.adoc
+++ b/docs/retrying-test.adoc
@@ -5,6 +5,27 @@ Some tests, e.g. those depending on external systems, may fail through no fault 
 Such tests make a suite fragile and it makes sense to try and avoid them, but if that's infeasible, it may help to retry a number of times before eventually assuming actual failure.
 `RetryingTest` provides that functionality.
 
+== Attributes
+
+=== maxAttempts (alias: value) [required]
+
+The `maxAttempts` attribute specifies the maximum number of times to execute the test function.
+If the test function throws an exception, the test will be executed again, up to this limit.
+
+The `value` attribute is an alias for `maxAttempts`. The attribute can be set either
+with `@RetryingTest(N)` or `@RetryingTest(maxAttempts=N)`.
+
+This attribute is required and must be a value greater than `minSuccess`. Only one of
+`maxAttempts` or `value` can be set, but not both.
+
+=== minSuccess [optional]
+
+The `minSuccess` attribute specifies the number of times the test function must complete
+successfully (i.e. without throwing an exception). The test function will be executed,
+up to `maxAttempts`, until it completes successfully this number of times.
+
+The default for `minSuccess` is 1. The value must be greater than or equal to 1.
+
 == Basic Use
 
 `@RetryingTest(n)` is used _instead_ of `@Test` or other such annotations (e.g. `@RepeatedTest`).
--- a/docs/retrying-test.adoc
+++ b/docs/retrying-test.adoc
@@ -12,17 +12,16 @@ Such tests make a suite fragile and it makes sense to try and avoid them, but if
 The `maxAttempts` attribute specifies the maximum number of times to execute the test function.
 If the test function throws an exception, the test will be executed again, up to this limit.
 
-The `value` attribute is an alias for `maxAttempts`. The attribute can be set either
-with `@RetryingTest(N)` or `@RetryingTest(maxAttempts=N)`.
+The `value` attribute is an alias for `maxAttempts`.
+The attribute can be set either with `@RetryingTest(N)` or `@RetryingTest(maxAttempts=N)`.
 
-This attribute is required and must be a value greater than `minSuccess`. Only one of
-`maxAttempts` or `value` can be set, but not both.
+This attribute is required and must be a value greater than `minSuccess`.
+Only one of `maxAttempts` or `value` can be set, but not both.
 
 === minSuccess [optional]
 
-The `minSuccess` attribute specifies the number of times the test function must complete
-successfully (i.e. without throwing an exception). The test function will be executed,
-up to `maxAttempts`, until it completes successfully this number of times.
+The `minSuccess` attribute specifies the number of times the test function must complete successfully (i.e. without throwing an exception).
+The test function will be executed, up to `maxAttempts`, until it completes successfully this number of times.
 
 The default for `minSuccess` is 1. The value must be greater than or equal to 1.
 
--- a/docs/retrying-test.adoc
+++ b/docs/retrying-test.adoc
@@ -65,8 +65,8 @@ void requiresTwoSuccesses() {
 }
 ----
 
-The test `requiresTwoSuccesses` must run at least two times without raising
-an exception. However, it will not run more than four times.
+The test `requiresTwoSuccesses` must run at least two times without raising an exception.
+However, it will not run more than four times.
 
 [source,java]
 ----
--- a/src/main/java/org/junitpioneer/jupiter/RetryingTestExtension.java
+++ b/src/main/java/org/junitpioneer/jupiter/RetryingTestExtension.java
@@ -68,12 +68,14 @@ public class RetryingTestExtension implements TestTemplateInvocationContextProvi
 	private static class FailedTestRetrier implements Iterator<RetryingTestInvocationContext> {
 
 		private final int maxRetries;
+		private final int minSuccess;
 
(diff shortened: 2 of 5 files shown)


src/main/java/org/junitpioneer/jupiter/RetryingTestExtension.java
```diff
@@ -82,7 +84,34 @@ static FailedTestRetrier createFor(Method test) {
...
+					throw new IllegalStateException("@RetryTest requires that one of `value` or `maxAttempts` be set, but not both.");
+				}
+
+				maxAttempts = retryingTest.value();
+				maxAttemptsField = "value";
+			}
+
+			if (maxAttempts == 0) {
+				// maxAttempts defaults to minSuccess
+				maxAttempts = minSuccess;
```
Michael1993, line 103: At which point this becomes a `@RepeatedTest`. Why is this not an exception?

Michael1993 (changes requested):
Thank you for your PR! Please explain your thought process why you made an alias for `value()` and why you made it optional.

src/main/java/org/junitpioneer/jupiter/RetryingTestExtension.java
dump247: I am not sure if I understand the question, but I will try to address it. With this condition, `@RepeatedTest(minSuccess = 4)` is equivalent to `@RepeatedTest(maxAttempts = 4, minSuccess = 4)`. So "the test must run 4 times successfully". Does that make sense?

dump247:
> Thank you for your PR! Please explain your thought process why you made an alias for `value()` and why you made it optional.

I added the alternative field name because I think that `@RepeatedTest(maxAttempts = 4, minSuccess = 2)` is easier to understand than `@RepeatedTest(value = 4, minSuccess = 2)`. The `value` field can not be removed because the existing usage has to be supported.

There are several reasons it is optional. First is that it has to be to have the two fields `value` and `maxAttempts`. It can not be that one is required. Second is that I think it is useful to support `@RepeatedTest(minSuccess = 4)` meaning "the test must run 4 times successfully" (equivalent to `@RepeatedTest(maxAttempts = 4, minSuccess = 4)`).

src/main/java/org/junitpioneer/jupiter/RetryingTestExtension.java
dump247: I completely misread your comment. You are correct. It becomes equivalent to a RepeatedTest. It can be an exception of you want it to. Someone could could still set maxAttempts == minSucess, but they would then have to do it explicitly. Do you want me to change it to an exception?

dump247: Well, it's not exactly like RepeatedTest. `@RepeatedTest(4)` will always run the test 4 times, even if it fails. `RetryingTest(minSuccess = 4)` will run the test 4 times, but stop on the first failure. I do not know if that is useful or not. I am just pointing out the difference.

Michael1993: The test suite will fail either way. Unless the test is *very* expensive, there is no real advantage to use `@RetryingTest` over `@RepeatedTest` in such a scenario. Am I missing something?

dump247: Correct. My apologies. I was just pointing out the slight difference. I'm not arguing to keep it the way I have it.

just let me know what you want me to do. Do you want me to change to an exception?

Michael1993: Sorry, I can't give you a definitive answer yet, because I am not sure. I'll let the maintainer team know about this, have a team discussion and update you with our decision.

Michael1993:
Could you fix the builds? It seems it's mainly just spotless.

src/main/java/org/junitpioneer/jupiter/RetryingTestExtension.java
Bukama: I'm more with @Michael1993, I think the max tries must always be higher than the min tries. Otherwise you don't need the annotation anyway as you can scale it down to one.

nipafx:
As I see it, there are three separate topics being discussed:

* `value` vs `maxAttempts`
* which are optional?
* is `minSuccess == maxAttempts` allowed?

## `value` vs `maxAttempts`

* `value`: We can make backwards incompatible changes, so deprecating and eventually removing `value` is not out of the question. Yet, I think it is a good idea to keep the extension's common use case simple, i.e. allowing `@RetryingTest(3)` instead of forcing `@RetryingTest(maxAttempts = 3)`.
* `maxAttempts`: It is always unfortunate to introduce an attribute with the same functionality as an existing one, but not only is there precedence, I think it makes sense here. `@RetryingTest(maxAttempts = 3, minSuccess = 2)` reads much better than `@RetryingTest(value = 3, minSuccess = 2)`

So I'm in favor of having both (for now and in the long run).

## Which are optional?

This can be answered syntactically and semantically:

* Syntactically, obviously both need to be optional or the user would always have to specify both.
* Semantically, I think at least one of them should be set. Otherwise it is not clear what exactly will happen - one could even interpret `@RetryingTest(minSuccess = 2)` to mean "retry until you get two successes", which would be a fun implementation, but probably undesired. 😂 

That means the extension should check whether exactly one of the two parameters was set and throw an error if zero or two were used.

## Is `minSuccess == maxAttempts` allowed?

I agree with Michael and Bukama that `minSuccess == maxAttempts` makes little sense as one could just use `@RepeatedTest`. Then again, it feel almost vindictive to prevent the user from using the extension for something that is clearly and easily capable of "just because".

(Aside: I don't think it's a good argument that `@RetryingTest` is lazy, but `@RepeatedTest` isn't. As far as I know neither of these behaviors are documented and if laziness is the better solution (comes at a cost of parallelism, though), `@RepeatedTest` could be changed accordingly.)

Since I'm undecided but two other maintainers aren't, I think we can conclude that we forbid that case. The error message should recommend using `@RepeatedTest`.

nipafx:
By the way, thanks for the contribution @dump247! Don't forget to add yourself to the README.md.

(I took the liberty to run `gradle spotlessApply` to fix the formatting errors.)

dump247:
I believe I have addressed the comments, but let me know if there is anything else. I am happy to make any requested changes.

src/main/java/org/junitpioneer/jupiter/RetryingTestExtension.java
dump247: It now raises an exception if maxAttempts is not set or if maxAttempts <= minSuccess.

src/main/java/org/junitpioneer/jupiter/RetryingTestExtension.java
```diff
@@ -82,7 +84,37 @@ static FailedTestRetrier createFor(Method test) {
 			RetryingTest retryingTest = AnnotationSupport
 					.findAnnotation(test, RetryingTest.class)
 					.orElseThrow(() -> new IllegalStateException("@RetryingTest is missing."));
-			return new FailedTestRetrier(retryingTest.value());
+
+			int maxAttempts = retryingTest.maxAttempts();
+			String maxAttemptsField = "maxAttempts";
```
Michael1993, line 89: I don't like this. It's `maxAttempts` that you can set through `value` for convenience. Other than that, it's good.

dump247: Ok, I will change the error to always say "maxAttempts". I was just trying to reference the attribute the user actually used in the annotation.

src/main/java/org/junitpioneer/jupiter/RetryingTestExtension.java
```diff
@@ -82,7 +84,35 @@ static FailedTestRetrier createFor(Method test) {
...
+				if (retryingTest.maxAttempts() != 0) {
+					throw new IllegalStateException(
+						"@RetryTest requires that one of `value` or `maxAttempts` be set, but not both.");
+				}
+
+				maxAttempts = retryingTest.value();
+			}
+
+			if (maxAttempts == 0) {
+				throw new IllegalStateException("@RetryTest requires that one of `value` or `maxAttempts` be set.");
```
Michael1993, line 101:
```suggestion
				throw new IllegalStateException("@RetryingTest requires that only one of `value` or `maxAttempts` be set.");
```

src/main/java/org/junitpioneer/jupiter/RetryingTestExtension.java
```diff
@@ -82,7 +84,35 @@ static FailedTestRetrier createFor(Method test) {
...
 					.orElseThrow(() -> new IllegalStateException("@RetryingTest is missing."));
-			return new FailedTestRetrier(retryingTest.value());
+
+			int maxAttempts = retryingTest.maxAttempts();
+			int minSuccess = retryingTest.minSuccess();
+
+			if (retryingTest.value() != 0) {
+				if (retryingTest.maxAttempts() != 0) {
+					throw new IllegalStateException(
+						"@RetryTest requires that one of `value` or `maxAttempts` be set, but not both.");
```
Michael1993, line 94:
```suggestion
						"@RetryingTest requires that one of `value` or `maxAttempts` be set, but not both.");
```

docs/retrying-test.adoc
```diff
@@ -57,6 +57,17 @@ The test `aborted` is aborted before (or during) its first execution.
 The first execution is marked as aborted/skipped, containing the underlying cause.
 The test suite as a whole is also marked as aborted/skipped.
 
+[source,java]
```
Michael1993, line 60: Could you add a few more sentences about `minSuccess`? I think this is a bit too short.

Michael1993 (changes requested):
Still finding small things, I'd like to see improved, but we are getting there.

src/main/java/org/junitpioneer/jupiter/RetryingTestExtension.java
```diff
@@ -92,16 +122,14 @@ void failed(Throwable exception) {
...
-						exceptionsSoFar, maxRetries),
-					exception);
-			else
+			if (hasNext())
 				throw new TestAbortedException(
-					format("Test execution #%d (of up to %d) failed ~> will retry...", exceptionsSoFar, maxRetries),
+					format("Test execution #%d (of up to %d) failed ~> will retry...", retriesSoFar, maxRetries),
 					exception);
+			else
+				throw new AssertionError(format(
```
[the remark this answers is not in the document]
Michael1993: This is purely a "nice-to-have" thing, I'm just mentioning it just in case you happen to be in the mood to do it.

src/test/java/org/junitpioneer/jupiter/RetryingTestExtensionTests.java
```diff
@@ -126,6 +203,51 @@ void skipByAssumption() {
...
+		@RetryingTest(maxAttempts = 1, minSuccess = 1)
+		void maxAttemptsEqualsMinSuccess() {
+			// Do nothing
+		}
+
+		@RetryingTest(maxAttempts = 2, minSuccess = 0)
+		void minSuccessLessThanOne() {
+			// Do nothing
+		}
+
```
Michael1993, line 250: More tests:
* `maxAttempts` <= 0
* `maxAttempts` < `minSuccess`

dump247: The second suggested test should be covered by `maxAttemptsEqualsMinSuccess`. The condition is that `maxAttempts` <= `minSuccess`. I added the requested test though.

docs/retrying-test.adoc
dump247: I added sections describing the attributes to the documentation. Let me know how you feel about that.

nipafx:
Thank you for your contribution @dump247. 😃👍🏾

PR #433, opened 2021-02-24
Fix SetEnvironmentVariable and SetSystemProperty interference

### LINKED ISSUE #432: SetEnvironmentVariable and SetSystemProperty interfere
When `@SetEnvironmentVariable` and `@SetSystemProperty` are used on the same test, only one set of properties is restored at the end of the test.

It looks like they both use `AbstractEntryBasedExtension` and that will overwrite the `EntriesBackup` for the first annotation when the second annotation is processed since it's stored in the same place in the context.

I think either the namespace or the key used to store the backup in the context needs to be derived from the implementing class.

beatngu13:
Closes #432.

Proposed commit message:

```
Create namespace per entry-based extension (#432 / #433)

Instead of creating a single namespace based on
`AbstractEntryBasedExtension.class` for all entry-based extensions,
using `getClass()` ensures per-extension namespaces via the runtime
class (i.e. the actual extension).

Closes: #432
PR: #433
```

---
**PR checklist**

Documentation (general)

Documentation (new extension)

Code

Contributing

---

### DIFF
--- a/src/main/java/org/junitpioneer/jupiter/AbstractEntryBasedExtension.java
+++ b/src/main/java/org/junitpioneer/jupiter/AbstractEntryBasedExtension.java
@@ -24,6 +24,7 @@ import org.junit.jupiter.api.extension.BeforeEachCallback;
 import org.junit.jupiter.api.extension.ExtensionConfigurationException;
 import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.jupiter.api.extension.ExtensionContext.Namespace;
+import org.junit.jupiter.api.extension.ExtensionContext.Store;
 
 /**
  * An abstract base class for entry-based extensions, where entries (key-value
@@ -35,7 +36,6 @@ import org.junit.jupiter.api.extension.ExtensionContext.Namespace;
 abstract class AbstractEntryBasedExtension<K, V>
 		implements BeforeAllCallback, BeforeEachCallback, AfterAllCallback, AfterEachCallback {
 
-	private static final Namespace NAMESPACE = Namespace.create(AbstractEntryBasedExtension.class);
 	private static final String BACKUP = "Backup";
 
 	/**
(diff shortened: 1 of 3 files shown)


src/main/java/org/junitpioneer/jupiter/AbstractEntryBasedExtension.java
```diff
@@ -147,7 +147,11 @@ public void afterAll(ExtensionContext context) throws Exception {
 	}
 
 	private void restoreOriginalEntries(ExtensionContext context) {
-		context.getStore(NAMESPACE).get(BACKUP, EntriesBackup.class).restoreBackup();
+		getStore(context).get(BACKUP, EntriesBackup.class).restoreBackup();
+	}
+
+	private Store getStore(ExtensionContext context) {
+		return context.getStore(Namespace.create(getClass()));
```
Michael1993, line 154: Always creates a new `Namespace`?

Michael1993: Consider the following solution:
```java
protected abstract Namespace namespace();
```
which in turn gets turned into
```java
private static final Namespace NAMESPACE = Namespace.create(...);

@Override
protected Namespace namespace() {
  return NAMESPACE;
}
```
Is that too complex?

Michael1993 (commented):
It looks good to me, but I do have a question.

src/main/java/org/junitpioneer/jupiter/AbstractEntryBasedExtension.java
beatngu13: I borrowed this from JUnit's `TimingExtension` [example](https://junit.org/junit5/docs/current/user-guide/#extensions-lifecycle-callbacks-timing-extension). It may not be ideal, but I'm not sure if it's worth to further blow up the base class -- we already have 6 `abstract` methods.

Alternatively, we could also set it in the abstract ctor. Something like:

```java
abstract class AbstractEntryBasedExtension {

    private final Namespace namespace;

    protected AbstractEntryBasedExtension() {
        this.namespace = Namespace.create(getClass());
    }

}
```

What do you prefer? Keep it as it is, additional `abstract` method, abstract ctor, ...?

Michael1993: It doesn't really matter, it's a very minor thing.

Michael1993: Out of all the possible solution, yours is probably the easiest to understand (it's all in one place) and it shouldn't impact performance. Let's leave it as it is.

Michael1993 (approved):
LGTM

Azquelt (approved):
This change looks good to me and I have confirmed it fixes the issue I was seeing. Thanks :smile:

Bukama (approved):
LGTM :)

Issue #320, opened 2020-08-12
Thinking about default JUnit5 Version


aepfli:
As of Junit 5.6-M1 one of the changes is
```
 In the `EngineTestKit` API, the `all()`, `containers()`, and `tests()` methods in
  `EngineExecutionResults` have been deprecated in favor of the new `allEvents()`,
  `containerEvents()`, and `testEvents()` methods, respectively. The deprecated methods
  will be removed in JUnit Platform 1.7.0.
```

We do rely on those methods in their deprecated way. As we are close to release version `1.0.0` i think we might want to reconsider our targeted junit5 version and increase from `5.5.2` to `5.6.0` to be save for the future.

Thoughts @junit-pioneer/junit-pioneer-team ?

beatngu13:
Just a side note: This issue nicely shows that the experimental JUnit 5 builds are really neat. Kudos @aepfli!

Bukama:
As they will be removed in near future I agree that we need to migrate. `5.6.0` would fit our rule of "the lowest possible" version.

aepfli:
adapted title and changed description

aepfli:
based on the maintainer meeting, we decided to postpone this update till we really needs this feature, as it could block us.

Normally there should be no influence to users.

Bukama:
If update to 5.7.0 is needed the closed draft #345 shows which updates are needed (as of today)

Bukama:
From #431 

> Regarding the JUnit 5 version we depend on:
> 
> We usually want to depend on the lowest minor version (newest patch of that) that supports our feature set. Generally speaking, that's still the case, but the upcoming hack.commit.push event required an exception. Updating from Jupiter 5.6 to 5.7 requires some changes to our test code, which means it's a bit disruptive. In case some contributors need to update during the event, they would also need to make these changes in our test code, which would be an unnecessary burden. To preempt that, I decided to update to 5.7 now.

This update was done in #445.

So this issue can be closed.

PR #445, opened 2021-03-02
Test more junit versions

nipafx:
Added 5.7 and 5.8 (milestone) to the test matrix.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,7 +1,7 @@
 # JUnit5 jupiter uses version 5.5.2 and JUnit Platform uses version 1.5.2
 # and the minor version is always in sync. Due to that fact, we just set
 # the minor verison.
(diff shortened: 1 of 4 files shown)


Bukama (approved):
LGTM, but why the thoughtchange to add them now and change the code?

See also: #320

gradle.properties
```diff
@@ -1,7 +1,7 @@
 # JUnit5 jupiter uses version 5.5.2 and JUnit Platform uses version 1.5.2
```
Michael1993, line 1: What about the comment here?

nipafx: Good that you spotted that. This was left over from before using the JUnit 5 BOM and should've been removed when we used that instead of creating Jupiter and Platform version by string concatenation.

Issue #460, opened 2021-04-11
Investigate usage of findClosestEnclosingRepeatableAnnotations in EnvironmentVariableExtension and SystemPropertyExtension


beatngu13:
During a discussion on `findClosestEnclosingRepeatableAnnotations` vs. `findAllEnclosingRepeatableAnnotations` (`PioneerAnnotationUtils`) in Discord, the question came up why `EnvironmentVariableExtension` and `SystemPropertyExtension` are using `findClosestEnclosingRepeatableAnnotations` instead of `findAllEnclosingRepeatableAnnotations`.

In contrast to e.g. `DefaultLocalExtension` or `DefaultTimezoneExtension` (where #307 removed `BeforeAllCallback` and `AfterAllCallback`, which is still implemented by the extensions in question), `EnvironmentVariableExtension` and `SystemPropertyExtension` must not stop on the closest annotation:

```java
@ClearSystemProperty(key = "foo")
class MyTest {

  @ClearSystemProperty(key = "bar")
  void test() {
    assertThat(System.getProperty("foo")).isNull();
    assertThat(System.getProperty("bar")).isNull();
  }

}
```

Otherwise, the assert on "foo" would fail (suppose the original value is not `null`).

Therefore, we should investigate (i) why the extensions are using `findClosestEnclosingRepeatableAnnotations` instead of `findAllEnclosingRepeatableAnnotations`, (ii) and why the given tests – which are covering the above example – are actually passing.

beatngu13:
Consider the following test:

```java
@ClearSystemProperty(key = "foo")
@SetSystemProperty(key = "bar", value = "spam")
class MyTest {

  void testA() {
    assertThat(System.getProperty("foo")).isNull();
    assertThat(System.getProperty("bar")).isEqualTo("spam");
  }

  @SetSystemProperty(key = "foo", value = "eggs")
  @ClearSystemProperty(key = "bar")
  void testB() {
    assertThat(System.getProperty("foo")).isEqualTo("eggs");
    assertThat(System.getProperty("bar")).isNull();
  }

}
```

A setup like this perfectly reasonable, i.e. to have class-level defaults but method-level configurations that overwrite these defaults.

In order to address this with `findAllEnclosingRepeatableAnnotations` (and `BeforeEachCallback`/`AfterEachCallback` exclusively), we would have to preserve order during annotation discovery and when creating the collections of entries to clear and set:

- `testA()`: Clear "foo" and set "bar" to "spam".
[...]

PR #481, opened 2021-05-12
Fix backup mechanism in AbstractEntryBasedExtension

### LINKED ISSUE #480: SetSystemProperty / SetEnvironmentVariable broken for nested classes
Looks like 3ecf1338f552d99a5192999ee142d09760413d78 had the side effect of breaking SetSystemProperty and SetEnvironmentVariable when used in nested classes. Example for SetSystemProperty:

```java
@SetSystemProperty(key = "PROP", value = "PROP-VALUE")
class NestedSetSystemPropertyTest {
    @Nested
    @TestMethodOrder(MethodOrderer.OrderAnnotation.class)
    class nested {
        @Test
        @Order(1)
        void worksAsExpected() {
            assertEquals("PROP-VALUE", System.getProperty("PROP"));
        }

        @Test
        @Order(2)
        void breaksUnexpectedly() {
            assertEquals("PROP-VALUE", System.getProperty("PROP"));
        }
    }
}
```

This is due to `SystemPropertyExtension.entriesToSet()` now only returning the property "PROP" for the outermost `ExtensionContext` in `AbstractEntryBasedExtension.beforeEach()`. While this is fine during the setup phase (beforeEach/beforeAll) during tear-down `AbstractEntryBasedExtension.restoreOriginalEntries()` uses `ExtensionContext.Store.get()` to return the matching `EntriesBackup`: https://github.com/junit-pioneer/junit-pioneer/blob/3ecf1338f552d99a5192999ee142d09760413d78/src/main/java/org/junitpioneer/jupiter/AbstractEntryBasedExtension.java#L149-L151 

Note that `ExtensionContext.Store.get()` has the following [documented behaviour](https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/extension/ExtensionContext.Store.html#get(java.lang.Object,java.lang.Class)):
> If no value is stored in the current ExtensionContext for the supplied key, ancestors of the context will be queried for a value with the same key in the Namespace used to create this store.

So although the inner class and inner methods have no `EntriesBackup`, `AbstractEntryBasedExtension.restoreOriginalEntries()` grabs the outer context's instance during tear-down of `worksAsExpected()`, which accidentally clears the properties for the `breaksUnexpectedly()` test.

It would be great if `ExtensionContext.Store` were to provide a way to only query the store of the provided context and not the ancestor stores too. Haven't come up with an elegant solution for this yet. A workaround could be to store something that identifies the originating context within EntriesBackup on creation and to only restore an EntriesBackup-instance if the current context matches the originating context.

(Our workaround is to pin the junit-pioneer version to 1.4.0 for now.)

beatngu13:
Closes #480.

Proposed commit message:

```
Fix AbstractEntryBasedExtension backups in nested tests (#480 / #481)

`ExtensionContext$Store#get(Object)` queries ancestors of the current
context if no value is stored under the supplied key. This can mess up
backups of system properties and environment variables in nested
classes if, for instance, the current test method has no corresponding
annotation but the surrounding container. Then, the first test
execution accidentally restores the backup, which leads to incorrect
system property / environment variables settings.

This PR sidesteps this issue by ensuring that the store / backup key
is unique per test or container, respectively, by using the given
display name.

Closes: #480
PR: #481
```

---
**PR checklist**

Documentation (general)

Documentation (new extension)

Code

Contributing

---

### DIFF
--- a/src/main/java/org/junitpioneer/jupiter/AbstractEntryBasedExtension.java
+++ b/src/main/java/org/junitpioneer/jupiter/AbstractEntryBasedExtension.java
@@ -146,7 +146,7 @@ abstract class AbstractEntryBasedExtension<K, V>
 	}
 
 	private Object getStoreKey(ExtensionContext context) {
-		return context.getDisplayName();
+		return context.getUniqueId();
 	}
 
 	private class EntriesBackup {
--- a/src/main/java/org/junitpioneer/jupiter/AbstractEntryBasedExtension.java
+++ b/src/main/java/org/junitpioneer/jupiter/AbstractEntryBasedExtension.java
@@ -38,12 +38,6 @@ abstract class AbstractEntryBasedExtension<K, V>
 
 	/**
 	 * @param context The current extension context.
-	 * @return <code>true</code> if one or more of the extension's annotations are present.
-	 */
-	protected abstract boolean isAnnotationPresent(ExtensionContext context);
-
-	/**
-	 * @param context The current extension context.
 	 * @return The entry keys to be cleared.
 	 */
 	protected abstract Set<K> entriesToClear(ExtensionContext context);
@@ -135,8 +129,7 @@ abstract class AbstractEntryBasedExtension<K, V>
 
 	@Override
 	public void afterEach(ExtensionContext context) throws Exception {
-		if (isAnnotationPresent(context))
-			restoreOriginalEntries(context);
+		restoreOriginalEntries(context);
 	}
 
 	@Override
--- a/src/main/java/org/junitpioneer/jupiter/AbstractEntryBasedExtension.java
+++ b/src/main/java/org/junitpioneer/jupiter/AbstractEntryBasedExtension.java
@@ -36,8 +36,6 @@ import org.junit.jupiter.api.extension.ExtensionContext.Store;
 abstract class AbstractEntryBasedExtension<K, V>
 		implements BeforeAllCallback, BeforeEachCallback, AfterAllCallback, AfterEachCallback {
 
-	private static final String BACKUP = "Backup";
-
 	/**
 	 * @param context The current extension context.
 	 * @return <code>true</code> if one or more of the extension's annotations are present.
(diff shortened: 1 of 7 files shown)


Michael1993:
Wouldn't it be simpler to only use `BeforeEachCallback` and `AfterEachCallback`?

beatngu13:
@Michael1993 as described in #460, this would require some bigger changes in the extension(s) such as preserving the order of entries to clear and set. I already have a stash locally for that, but I would like to separate refactoring from fixing this regression, if that is OK for you.

src/main/java/org/junitpioneer/jupiter/AbstractEntryBasedExtension.java
```diff
@@ -147,18 +138,26 @@ public void afterAll(ExtensionContext context) throws Exception {
...
-		getStore(context).get(BACKUP, EntriesBackup.class).restoreBackup();
+		getStore(context).getOrDefault(getStoreKey(context), EntriesBackup.class, new EntriesBackup()).restoreBackup();
 	}
 
 	private Store getStore(ExtensionContext context) {
 		return context.getStore(Namespace.create(getClass()));
 	}
 
+	private Object getStoreKey(ExtensionContext context) {
+		return context.getDisplayName();
```
Bukama, line 149: The `Displayname` is *NOT* unique! this can lead to further collisions!
But each test (and container) has an uniqueid. I think this is the more safer way.

src/test/java/org/junitpioneer/jupiter/AbstractEntryBasedExtensionTest.java
```diff
@@ -38,6 +38,7 @@ void setUp() {
 	}
 
 	@Test
+	@Issue("432")
```
Bukama, line 41: <3

Bukama (changes requested):
I set this to "request changes" because I see further conflict potential we should look into and see if there's s a "safer" way. Meaning if we can access the unique Id of a test at the point where you get the displayname at the moment.

src/main/java/org/junitpioneer/jupiter/AbstractEntryBasedExtension.java
beatngu13: Good point, switched to [`ExtensionContext#getUniqueId()`](https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/extension/ExtensionContext.html#getUniqueId()).

src/test/java/org/junitpioneer/jupiter/AbstractEntryBasedExtensionTest.java
beatngu13: Wonder why we didn't use it before, so helpful! 👏

Michael1993 (commented):
I don't know enough about this extension to approve this now.

Bukama (approved):
Thanks for changing. We should be safe with the uniqueID. 👍

PR #483, opened 2021-05-12
Sign artifacts conditionally

### LINKED ISSUE #482: Sign artifacts conditionally
As pointed out by @fknittel in [this comment](https://github.com/junit-pioneer/junit-pioneer/issues/480#issuecomment-839670979), JUnit Pioneer currently can't be used via JitPack. This is because `publishToMavenLocal` fails as it tries to sign the artifacts ([JitPack log](https://jitpack.io/com/github/junit-pioneer/junit-pioneer/v1.4.1/build.log) on v1.4.1):

```text
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':signMavenPublication'.
> Cannot perform signing task ':signMavenPublication' because it has no configured signatory
```

According to the [Signing Plugin docs](https://docs.gradle.org/current/userguide/signing_plugin.html#sec:conditional_signing), signin artifacts conditionally is a common usage pattern. We should do this too, not just to enable JitPack usage, but also to e.g. allow users to conveniently publish to their local Maven repo.

beatngu13:
Closes #482. (Disclaimer: Gradle noob!)

Proposed commit message:

```
Sign artifacts conditionally (#482 / #483)

To e.g. enable JitPack usage or make `publishToMavenLocal` pass
_without_ configured signatory, signing should happen conditionally.
The given PR modifies the build to only signs artifacts iff there's a
`project.version` and `publishToSonatype` is included in the Gradle
task graph.

This approach is compatible with JUnit Pioneer's release build since
`publishToSonatype` is executed and `project.version` is set via the
environment variable `ORG_GRADLE_PROJECT_version`, which in turn is
injected as a workflow parameter.

Closes: #482
PR: #483
```

---
**PR checklist**

Documentation (general)

Documentation (new extension)

Code

Contributing

---

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -174,7 +174,7 @@ publishing {
 
 signing {
 	setRequired({
-		project.version != "unspecified"
+		project.version != "unspecified" && gradle.taskGraph.hasTask("publishToSonatype")
 	})
 	val signingKey: String? by project
 	val signingPassword: String? by project
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -173,6 +173,9 @@ publishing {
 }
 
 signing {
+	setRequired({
+		project.version != "unspecified"
+	})
 	val signingKey: String? by project
 	val signingPassword: String? by project
 	useInMemoryPgpKeys(signingKey, signingPassword)
--- a/README.md
+++ b/README.md
@@ -111,7 +111,7 @@ The least we can do is to thank them and list some of their accomplishments here
 #### 2021
 
 * [Cory Thomas](https://github.com/dump247) contributed the `minSuccess` attribute in [retrying tests](https://junit-pioneer.org/docs/retrying-test/) (#408 / #430)
-* [Daniel Kraus](https://github.com/beatngu13) fixed bugs in the environment variable and system property extensions (#432 / #433, #448 / #449, and #473 / #476)
+* [Daniel Kraus](https://github.com/beatngu13) fixed bugs in the environment variable and system property extensions (#432 / #433, #448 / #449, and #473 / #476) and improved the build process (#482 / #483)
 * [Slawomir Jaranowski](https://github.com/slawekjaranowski) Migrate to new Shipkit plugins (#410 / #419)
 * [Stefano Cordio](https://github.com/scordio) contributed [the Cartesian Enum source](https://junit-pioneer.org/docs/cartesian-product/#cartesianenumsource) (#379 / #409 and #414 / #453)
 


aepfli (approved):
lgtm :)

beatngu13:
OK, I just realized this doesn't solve the JitPack issue (see [log](https://jitpack.io/com/github/beatngu13/junit-pioneer/issue~482-conditional-signing-v1.3.0-g0133c60-32/build.log)):

```text
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':signMavenPublication'.
> Cannot perform signing task ':signMavenPublication' because it has no configured signatory
```

This is because JitPack attempts to set the `project.version` via the command line flag `-Pversion`…

aepfli:
you could also make it conditionally based on the environment variables we are using

build.gradle.kts
```diff
@@ -174,7 +174,7 @@ publishing {
 
 signing {
 	setRequired({
-		project.version != "unspecified"
+		project.version != "unspecified" && gradle.taskGraph.hasTask("publishToSonatype")
```
aepfli, line 177: i am not sure if the `project.version` check is needed. I would rather check for the other two project variables we are using `signingKey` and `signingPassword` - this should cover all the cases sufficiently.

Not sure about the taskgraph, i think this is maybe too limiting, to a specific target

beatngu13: I followed the suggestion in the [Signing Plugin docs](https://docs.gradle.org/current/userguide/signing_plugin.html#sec:conditional_signing) to include the Gradle task graph. My assumption is that we only want to sign iff there's a project version _and_ we publish to Sonatype. This also seems to fix the JitPack build:

https://jitpack.io/com/github/beatngu13/junit-pioneer/e0bcc6c426/build.log

Regarding `signingKey` and `signingPassword`: consider the case where these variables are absent in the release build (for whatever reason). Since signing is skipped, the build would continue and attempt to upload to Sonatype, it wouldn't fail early.

Not sure about this, but we could let the build fail in the `signin` task by throwing an appropriate exception, if that makes sense?

aepfli: on the other hand, what if i want to sign my packages to test everything is running as expected without uploading to SonaType. In this case i would rather not have a task dependency, nor a project property - tough question.

I agree with your fail fast approach on releases, but i also tend to have builds work as expected locally. Tricky situation :D

beatngu13: Let's add another condition:

```kotlin
project.hasProperty("enforceSigning")
		|| project.version != "unspecified" && gradle.taskGraph.hasTask("publishToSonatype")
```

Then do:

```text
$ ./gradlew sign -PenforceSigning
```

🙈

But yeah, it's indeed tricky. At the moment, I don't think I have a clear preference … you?

aepfli: me neither, as said i agree on both sides :D you got my approve and i am fine - important is that jitpack works again :) for the rest we do have time

Bukama (commented):
I trust @aepfli knowledge :)

Issue #101, opened 2018-08-01
Support JSON as source for parameterized tests arguments


nipafx:
Implement an `ArgumentConverter` that parses JSON and an `ArgumentSource` that parses JSON files and passes them as arguments to a parameterized test.

This is pretty straightforward for a single file and parameter:

```java
@ParameterizedTest
@JsonSource("customers.json")
void testCustomer(Customer customer) { /*...*/ }
```

It's a little unclear how to parse files for two parameters, though. There could be one file per parameter...

```java
@ParameterizedTest
@JsonSource({ "customers.json", "contracts.json" })
void testCustomer(Customer customer, Contract contract) { /*...*/ }
```

... but then customer/contract pairs would be far apart and hard to identify by looking at the files.

Alternatively, there could be a top-level container...

```js
{
	customer: { /*...*/ }
	contract: { /*...*/ }
}
```

... and the parser takes them apart when resolving parameter:

```java
@ParameterizedTest
@JsonSource({ "customers-contracrs.json" })
void testCustomer(Customer customer, Contract contract) { /*...*/ }
```

NPException:
Since this issue is more than a year old now, which of the suggested approaches would be preferred?  
I personally would prefer the "single-file-with-top-level-container" variant. I feel one file per parameter could lead people to an unnecessarily large number of small JSON files.

aepfli:
i like the idea of a top level container, as it is quiet intuitive, and the testdata always belongs together. furthermore with the toplevel container in an array, we could also define a default object outside of the container, with default values. Sometimes you need all to have some data, and it needs to be the same. So we could utilize this, without duplicating data. just an idea.

```javascript
{
	defaults: {
		customer: {},
		contracts: {}
	}
	data: [{
		customer: {},
		contracts: {}
	}]
}
```

But i think multiple files are impractical, as you do not see data belonging together, and you always should ensure that they do contain the same amount of datasets anyways.

nipafx:
JSON-parsing should not be reimplemented for this issue, but also [JUnit Pioneer doesn't want to take on additional run-time dependencies](https://github.com/junit-pioneer/junit-pioneer/blob/main/CONTRIBUTING.md#dependencies), which means this issue can not be resolved within this project. If a PR is offered for it, we will _not merge_ it but instead take the new sources, documentation, etc. and spin it out into its own project. (That might of course take some time, so don't pick this up if you're looking for a quick win. 😉 )

filiphr:
[...]

Issue #415, opened 2021-01-25
Investigate parameter targeting of CartesianProductTest arguments source annotations


Michael1993:
A very strong argument can be made that it's much more convenient to define a CartesianProductTest like the following:
```java
@CartesianProductTest
void test(
    @CartesianValueSource(strings = { "A", "B", "C" }) String str,
    @CartesianEnumSource(ChronoUnit.class) ChronoUnit unit
) {
    // some passing test code here
}
```

Investigate if this is possible - it is likely it is not possible, because of how the annotations invoke their corresponding `ArgumentsProvider` implementations, but it might work.

scordio:
My two cents: I personally consider mimicking JUnit's `@Parameterized` tests a strong characteristic of `@CartesianProductTest`, which makes its adoption very quick. Targeting the parameters would deviate from the known JUnit patterns. On the other side, sticking to those patterns might limit the growth of the `@CartesianProductTest` feature.

It could still be fine if the example above is an alternative way of declaring the parameters, but then I am wondering if it's worth offering multiple ways to do the same thing (and maybe supporting combinations of both ways?)

Michael1993:
The main reason I opened this issue is because this is a frequently asked question. I would like to have a discussion/discovery about this, so I can link to it as an explanation, i.e.: "This is not possible, we investigated this in #415" or "While this is possible, the team came to a decision to not support it - see discussion on #415" or we might just decide to support it but discourage users from using it, etc.

Michael1993:
With the discovery of the repeatable annotations bug (#447), this investigation should have a bit more priority.

nipafx:
The similarity to `@ParameterizedTests` is pretty nice, but there's one important difference between the two extensions: For parameterized tests, annotating individual arguments makes no sense. For this extension it does (and a lot, at that). I'm really interested in this idea, particularly as it seems to offer a great solution to #447.

Let's discuss first whether we think this is a good idea - maybe you can support that conversation by digging into the implementation or even working on a proof of concept.

Once we know whether we prefer this solution and by how much and how complex it is to implement, we can discuss whether we offer one (which one?) or both approaches and how to possibly migrate if that becomes necessary.

nipafx:
While developing the proof of concept for #49 yesterday, I created annotations `@RandomString` and `@RandomInt` to be used as follows:

```java
@ParameterizedTest
@RandomArguments(count = 4)
void test(
[...]

PR #492, opened 2021-05-30
Support JSON as source for parametrized tests arguments

filiphr:
This is an initial implementation fro #101. It currently only supports [Jackson 2](https://github.com/FasterXML/jackson-databind) for Json Parsing. However, it should be fairly simple to add support for different parsers.

I know the policy about no runtime dependencies. Therefore, I tried using [Optional Dependencies](https://blog.gradle.org/optional-dependencies) from Gradle and its [feature variants](https://docs.gradle.org/6.0.1/userguide/feature_variants.html). If this is something interesting we can look into doing this with different source sets in order for complete isolation (currently the Checkstyle import-control is handling this).

There is still documentation missing. However, if you like the approach I will work on the documentation.

Proposed commit message:

```
Support for JSON as source for parametrized tests

Closes: #101
PR: #492
```

---
**PR checklist**

Documentation (general)

Documentation (new extension)

Code

Contributing

---


Bukama:
I would like to pass this to the other maintainers cause of the dependency thing and changes in the build script

src/main/java/org/junitpioneer/jupiter/json/JsonSource.java
```diff
@@ -0,0 +1,77 @@
...
+ * @see org.junit.jupiter.params.ParameterizedTest
+ * @see org.junitpioneer.jupiter.CartesianProductTest
+ * @since TBD
+ */
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@ArgumentsSource(JsonSourceArgumentsProvider.class)
+@Repeatable(JsonSource.JsonSources.class)
+public @interface JsonSource {
```
Michael1993, line 48: This is confusing. JUnit 5 has `@CsvSource` which can take CSV as a `String` and `CsvFileSource` which takes CSV file names. I think following that, this should be `JsonFileSource`.

filiphr: You are right about this. It would make more sense this to be named `@JsonFileSource`. I didn't really think about the naming and took what was there in the issue

Michael1993 (changes requested):
Pioneer actually has a `module-info.java` and a modular build, so you need to update that if you want the build to succeed. You can find it at `src/main/java/module/module-info.java`.

> (package com.fasterxml.jackson.databind is declared in module com.fasterxml.jackson.databind, but module org.junitpioneer does not read it)

nipafx:
From just looking at the tests, this seems to be exactly what we had in mind. Good job! I think it makes sense to pursue this further. 👍🏾

src/main/java/org/junitpioneer/jupiter/json/JsonSource.java
filiphr: I renamed this into `@JsonFileSource`.

filiphr:
>Pioneer actually has a `module-info.java` and a modular build, so you need to update that if you want the build to succeed. You can find it at `src/main/java/module/module-info.java`.

Thanks @Michael1993, I've updated the file. The modular build is now working correctly.

I've taken this a bit further, added some documentation, and did some improvements;

* Support for using `@Param` for single method arguments, by extracting only one specific element from it
* Support for deconstructing into complex types

nipafx:
Thank you for this PR, @filiphr. You've seen my streams, you know how much time management challenges me. I will take a closer look at this, but it will take some more time before I do.

Also, we need to resolve #502 before we can merge/move this.

docs/json.adoc
```diff
@@ -0,0 +1,156 @@
...
+:page-description: Extends JUnit Jupiter with `@JsonFileSource`, a parametrized test that creates test based on a JSON Source
+
+The `@JsonFileSource` annotation lets you use JSON files from the classpath or local file system to create parametrized tests.
+
+By default the root of the source will be treated as candidate for the test arguments.
+If the root is an object then the entire object will be one argument, if the root is an array then every element of the array will be one argument.
+
+There is also `JsonFileSource#data` that can be used to tell the extraction mechanism to use the element with that name to look for the source of the data.
+
+Depending on how the test method parameters the extraction of the values might differ.
```
nipafx, line 11:
```suggestion
Depending on the test method parameters, the extraction of the values might differ.
```

docs/json.adoc
```diff
@@ -0,0 +1,156 @@
...
+
+== Single Argument Methods
+
+When the method has a single argument then the json object argument will be converted to that type.
```
nipafx, line 15:
```suggestion
If the method has a single argument, the json object argument will be converted to that type.
```

docs/json.adoc
```diff
@@ -0,0 +1,156 @@
...
+[source,java]
+----
+@ParametrizedTest
+@JsonFileSource(resources = "/customers.json")
+void singleCustomer(Customer customer) {
+    assertThat(Collections.singleton(tuple(customer.getName(), customer.getHeight())))
+            .containsAnyOf(
+                tuple("Luke", 172),
+                tuple("Yoda", 66)
+            );
```
nipafx, line 55: I appreciate the attention to detail, but I don't think we need to flesh out the test. 😉

src/main/java/org/junitpioneer/jupiter/json/Param.java
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * An annotation indicating the name of the parameter that should be extracted into the method.
+ * This is needed when the test code is not compiled with Java 8 {@code -parameters} flag.
+ *
+ * @since TBD
+ */
+@Target(ElementType.PARAMETER)
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+public @interface Param {
```
nipafx, line 28: This name should probably be a bit longer - see [related discussion](https://github.com/junit-pioneer/junit-pioneer/issues/415#issuecomment-861770281) on Cartesian tests.

docs/json.adoc
```diff
@@ -0,0 +1,156 @@
...
+----
+
+This parametrized test will generate the following tests:
+
+* [1] Luke
+* [2] Yoda
+
+=== Multiple Argument Methods
+
+When the method has multiple arguments then each json object argument will be deconstructed to each of the method arguments.
```
nipafx, line 83:
```suggestion
If the method has multiple arguments, each json object argument will be deconstructed to each of the method arguments.
```

docs/json.adoc
```diff
@@ -0,0 +1,156 @@
...
+[source,java]
+----
+@ParametrizedTest
+@JsonFileSource(resources = "/customers.json")
+void deconstructFromArray(@Param("name") String customerName, @Param("height") int height) {
+    assertThat(Collections.singleton(tuple(customerName, height)))
+					.containsAnyOf(
+							tuple("Luke", 172),
+							tuple("Yoda", 66)
+					);
```
nipafx, line 104: As above.

docs/json.adoc
```diff
@@ -0,0 +1,156 @@
...
+[source,java]
+----
+@ParameterizedTest
+@JsonFileSource(resources = "customer-luke.json", data = "vehicles")
+void lukeVehicles(@Param("name") String name, @Param("length") double length) {
+    assertThat(Collections.singleton(tuple(name, length)))
+            .containsAnyOf(
+                tuple("Snowspeeder", 4.5),
+                tuple("Imperial Speeder Bike", 3d)
+            );
```
nipafx, line 149: As above.

docs/json.adoc
```diff
@@ -0,0 +1,156 @@
...
+}
+----
+
+This parametrized test will generate the following tests:
+
+* [1] Snowspeeder, 4.5
+* [2] Imperial Speeder Bike, 3
```
nipafx, line 156: Can you add a section on thread safety like the other extensions have? Will probably look [like this](https://junit-pioneer.org/docs/range-sources/#thread-safety).

src/main/java/org/junitpioneer/jupiter/json/JacksonJsonConverter.java
```diff
@@ -0,0 +1,48 @@
...
+import java.io.InputStream;
+import java.io.UncheckedIOException;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+/**
+ * A {@link JsonConverter} using Jackson 2 {@link ObjectMapper} to perform the conversion
+ */
+class JacksonJsonConverter implements JsonConverter {
```
nipafx, line 23: Can you explain why we need this class and `JacksonNode`?

nipafx: I also don't know why we need `JsonConverter` and `JsonConverterProvider`. 🙈 Halp?

src/main/java/org/junitpioneer/jupiter/json/JsonFileArgumentsProvider.java
```diff
@@ -0,0 +1,166 @@
...
+import org.junit.jupiter.params.support.AnnotationConsumer;
+import org.junit.platform.commons.PreconditionViolationException;
+import org.junitpioneer.internal.PioneerPreconditions;
+import org.junitpioneer.jupiter.CartesianAnnotationConsumer;
+
+/**
+ * The reading of the resources / files is heavily inspired by {@link org.junit.jupiter.params.provider.CsvFileArgumentsProvider}.
+ */
+class JsonFileArgumentsProvider
+		implements ArgumentsProvider, AnnotationConsumer<JsonFileSource>, CartesianAnnotationConsumer<JsonFileSource> {
```
nipafx, line 38: Since we're rebuilding Cartesian product tests, I'm not sure whether it makes sense to implement `CartesianAnnotationConsumer<JsonFileSource>`. @Michael1993, what do you think?

src/test/java/org/junitpioneer/jupiter/json/JsonFileSourceArgumentsProviderTests.java
```diff
@@ -0,0 +1,252 @@
...
+		assertThat(displayNames.get("deconstructCustomerMultipleFilesComplexType"))
+				.containsExactly("[1] Yoda, Location{name='unknown'}", "[2] Luke, Location{name='Tatooine'}");
+
+		assertThat(displayNames.get("singleCustomer"))
+				.containsExactly("[1] Customer{name='Luke', height=172}", "[2] Customer{name='Yoda', height=66}");
+
+		assertThat(displayNames.get("singleCustomerName")).containsExactly("[1] Luke", "[2] Yoda");
+
+		assertThat(displayNames.get("customDataLocation"))
+				.containsExactly("[1] Snowspeeder, 4.5", "[2] Imperial Speeder Bike, 3");
```
nipafx, line 67: Good job with these tests! Since they already check all the correct values, don't you think we can remove the assertions in the actual tests below? Otherwise, all changes need to be made in two places.

docs/json.adoc
```diff
@@ -0,0 +1,156 @@
...
+					);
+}
+----
+
+This parametrized test will generate the following tests:
+
+* [1] Luke, 172
+* [2] Yoda, 66
+
+=== Extracting nested array
```
nipafx, line 113:
```suggestion
== Extracting nested array
```

Shouldn't this be h2 - it applies to single-argument methods as well, right?

docs/json.adoc
```diff
@@ -0,0 +1,156 @@
+:page-title: JSON
+:page-description: Extends JUnit Jupiter with `@JsonFileSource`, a parametrized test that creates test based on a JSON Source
+
+The `@JsonFileSource` annotation lets you use JSON files from the classpath or local file system to create parametrized tests.
```
nipafx, line 4: I propose to slightly change the doc's structure. Add another sentence or two here what to look out for (single vs multiple arguments; nested arrays) and then add `== Method Arguments` and make the next two headers on single and multiple argument methods h3.

nipafx (changes requested):
This looks really good, I think we can merge it "soon" (modulo my very long latency). A few comments below.

Also, if I'm not missing something, this "only" addresses part of #101. The `ArgumentConverter` and a source that parses inline JSON (as opposed to from a file) are also mentioned there. If we close #101, we need to open new issues for those. (If we still want to have them.)

src/main/java/org/junitpioneer/jupiter/json/JacksonJsonConverter.java
filiphr: The idea of the `JsonConverter` and `Node` is to provide an extra layer of abstraction over the Json Parsers. This would allow for using different parser implementations, such as [GSON](https://github.com/google/gson), [Johnzon](https://johnzon.apache.org/), [Moshi](https://github.com/square/moshi), without any changes in the main logic. 

This was inspired by the [`NodeFactory`](https://github.com/lukas-krecan/JsonUnit/blob/master/json-unit-core/src/main/java/net/javacrumbs/jsonunit/core/internal/NodeFactory.java) from the JsonUnit project.

See also `JsonConverterProvider` in this PR.

I will add some Javadoc to `JsonConverter` and `Node` to make it clearer.

src/main/java/org/junitpioneer/jupiter/json/JsonFileArgumentsProvider.java
filiphr: I've added this because I saw it in the range functionality. I can remove it if it is not needed.

src/test/java/org/junitpioneer/jupiter/json/JsonFileSourceArgumentsProviderTests.java
filiphr: We can remove them yes, but then if you want to debug you'll need to write something to test things out.

If new things are added I expect new json files to be used.

docs/json.adoc
filiphr: Will doo that

docs/json.adoc
filiphr: Makes sense. I'll do that

src/main/java/org/junitpioneer/jupiter/json/Param.java
filiphr: I will change it to `Parameter`. If you have better idea let me know

docs/json.adoc
filiphr: 😄  I'll remove them from the docs

filiphr:
>This looks really good, I think we can merge it "soon" (module my very long latency).

Thanks for the review. I've applied the feedback and have answered some of your comments. I am not in a rush (as you can see I have a long latency as well). 

>Also, if I'm not missing something, this "only" addresses part of #101. The ArgumentConverter and a source that parses inline JSON (as opposed to from a file) are also mentioned there. If we close #101, we need to open new issues for those. (If we still want to have them.)

Indeed this solves only a part of #101, the one where data is read from a classpath resource or a file. Reading from an inline JSON would be quite trivial as well. I'll try to add it as part of this PR as well.

I am not sure about the `ArgumentConverter` though, what would you expect from it? Currently we extract the properties in the types that they are in the JSON (string, number, boolean, etc.)

filiphr:
By the way, I forgot to mention, I watched the stream afterwards to see your thought process when reviewing this 😄 . Couldn't be there when you were doing it live

src/main/java/org/junitpioneer/jupiter/json/JsonFileArgumentsProvider.java
nipafx: Ad hoc decision: For new extensions that can interact with Cartesian tests, let's only implement the new approach. Can you change this? :)

src/test/java/org/junitpioneer/jupiter/json/JsonFileSourceArgumentsProviderTests.java
nipafx: Got it. It's unfortunate, but makes sense.

src/main/java/org/junitpioneer/jupiter/json/JacksonJsonConverter.java
nipafx: Got it!

nipafx (changes requested):
Looking good, but we found some things still left to do:

* whatever we decide [on the issue](https://github.com/junit-pioneer/junit-pioneer/issues/101#issuecomment-968281501)
* support new Cartesian tests instead of the old approach
* I still need to wrap my head around #502 (this PR currently uses [Gradle's feature variants](https://docs.gradle.org/current/userguide/feature_variants.html)).

src/main/java/org/junitpioneer/jupiter/json/JsonConverterProvider.java
```diff
@@ -0,0 +1,38 @@
...
+			return false;
+		}
+	}
+
+	static JsonConverter getJsonConverter(ExtensionContext context) {
+		if (jacksonPresent) {
+			return JacksonJsonConverter.getConverter();
+		}
+
+		throw new PreconditionViolationException("There is no available Json parsing library");
```
nipafx, line 35: This error case comes up when users haven't configured a (supported) JSON parse, so we need to make sure, it's visible. Both Maven and Gradle will cut the error message very short, so it's unclear what the problem is.

Some options:

* throw `NoJsonParserConfiguredException` (extends `PreconditionViolationException`)
* print to `System.err`
* log a report entry with JUnit Jupiter

We should also mention which parsers we support.

filiphr: Which Maven did you use? Using 3.8.4 I get

>[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.031 s <<< FAILURE! - in org.junitpioneer.demo.JsonSourceTests
[ERROR] org.junitpioneer.demo.JsonSourceTests.jsonTest(String, int)  Time elapsed: 0.001 s  <<< ERROR!
org.junit.platform.commons.PreconditionViolationException: There is no available Json parsing library

Nothing is cut short. I will add more information about which json parsers are currently supported. 

For gradle when I do `./gradlew test` I only get that test failed. In the test report you can see the full error with its stacktrace.

When I add

```
tasks.named<Test>("test") {
	useJUnitPlatform()

	testLogging {
		setExceptionFormat("full")
	}
}
```

then I get

>JsonSourceTests > jsonTest(String, int) > org.junitpioneer.demo.JsonSourceTests.initializationError FAILED
    org.junit.platform.commons.PreconditionViolationException: There is no available Json parsing library
        at app//org.junitpioneer.jupiter.json.JsonConverterProvider.getJsonConverter(JsonConverterProvider.java:35)
        .... There is actually more stacktrace


Regarding the options you mentioned:

* What is the benefit of throwing `NoJsonParserConfiguredException`?
* I am not a fan of printing to `System.err`. So will leave it to you to add it in case you really really want it
* I can log a report entry with JUnit Jupiter, need to see how to do that first.

nipafx: I don't recall the details any more (stupid me, I should've added what I saw as text or screenshot), so I'll rely on your analysis. Regarding the options:

* The exception name showed up prominently - if that's `NoJsonParserConfiguredException`, it might not even be necessary to read the text.
* Me neither, let's not do it then.
* you know what, ignore that. I think it makes sense to limit report entries to (mostly) successful operations, not for situations that abort the test run.

src/test/java/org/junitpioneer/jupiter/ReportEntryExtensionTests.java
```diff
@@ -338,7 +338,7 @@ void conditional_doesNotLogOnDisabled() {
 
 	@Nested
 	@DisplayName("with parameterized tests")
-	class ParameterEntriesTests {
+	class PropertyEntriesTests {
```
nipafx, line 341: Why did you rename this? It looks like it's indeed related to parameterized tests.

docs/json-argument-source.adoc
```diff
@@ -0,0 +1,192 @@
+:page-title: JSON Argument Source
```
nipafx, line 1: I renamed the file (and thus the URL) and changed the title to limit it to JSON argument sources - who knows what other fancy things we can do with JSON in the future that would otherwise name-clash. 😉

src/test/java/org/junitpioneer/jupiter/json/JsonClasspathSourceArgumentsProviderTests.java
```diff
@@ -0,0 +1,327 @@
...
+import org.junit.platform.engine.support.descriptor.MethodSource;
+import org.junit.platform.testkit.engine.Event;
+import org.junitpioneer.jupiter.cartesian.CartesianTest;
+import org.junitpioneer.testkit.ExecutionResults;
+import org.junitpioneer.testkit.PioneerTestKit;
+
+/**
+ * Tests for {@link JsonClasspathSourceArgumentsProvider}
+ */
+class JsonClasspathSourceArgumentsProviderTests {
```
nipafx, line 37: No tests for files? Too brittle?

nipafx (changes requested):
Looks really good. The documentation doesn't reflect the three annotations yet (I only updated the intro up until the bullet point list), but otherwise this looks good to merge. 👍🏾

nipafx:
I just noticed that the module declaration (in `src/modules`) still needs to be updated.

src/test/java/org/junitpioneer/jupiter/ReportEntryExtensionTests.java
filiphr: IntelliJ hickup during some other renaming. I'll revert it

filiphr:
I closed this by accident. I've added some replies to your comments. And I fixed some of the problems that you found. Let me know what you think.

I've also added myself to the contributors list to this list, I hope I didn't mess this up.

nipafx:
> What is missing in the documentation?

It still uses the old-style `@JsonFileSource(resources = "/jedis.json")` syntax. While udpating that, you could also create the snippets as a proper demo in `src/demo/java`, which gets compiled, so we see when things are outdated. See `report-entries.adoc` for how to reference them (note the attribute `:xp-demo-dir:` at the top).

> I've also added myself to the contributors list to this list, I hope I didn't mess this up.

Nope, looks good. And you definitely belong there for this 12-month-juggernaut. :D

> What needs to be updated there?

Nothing, looks good. 😬

src/test/java/org/junitpioneer/jupiter/json/JsonClasspathSourceArgumentsProviderTests.java
nipafx: This should work, right?

```java
@ParameterizedTest
@JsonFileSource("build/resources/test/org/junitpioneer/jupiter/json/jedis.json")
void deconstructCustomerMultipleLinesComplexType(@Property("name") String name, @Property("height") int height) {
	assertThat(Collections.singleton(tuple(name, height)))
		.containsAnyOf(tuple("Luke", 172), tuple("Yoda", 66));
}
```

filiphr:
I've updated the documentation. It was a good thing to mention the use of proper compilable demo, because even the test annotation had a typo. I think everything is covered now, if there is something more missing, please do let me know

nipafx:
Commit message:

```
Support JSON sources for parametrized tests 

This extension needs a JSON parser. In order not do pull in a
specific run-time dependency, it makes use of Gradle's feature
variants as described in CONTRIBUTING.md (including Checkstyle
rules). For now, only Jackson is supported as JSON parser. Other
parsers can be added in future versions if users need them.

Closes: #101
PR: #492
```

nipafx (approved):

nipafx:
Great job, @filiphr, and big thanks for your patience!

filiphr:
Thanks for merging this @nipafx. It was fun working on this, it doesn't matter how long it took :)

PR #493, opened 2021-05-30
Add TestKitTest for easier running of PioneerTestKit in tests

filiphr:
This is a potential idea for making it simpler (in my eyes) the execution of `PioneerTestKit` tests.

before:

```java
@Test
@WritesDefaultLocale
@DisplayName("should execute tests with configured Locale")
void shouldExecuteTestsWithConfiguredLocale() {
    ExecutionResults results = executeTestClass(ClassLevelTestCase.class);

    assertThat(results).hasNumberOfSucceededTests(2);
}
```

after:

```java
@TestKitTest(testClass = ClassLevelTestCase.class)
@WritesDefaultLocale
@DisplayName("should execute tests with configured Locale")
void shouldExecuteTestsWithConfiguredLocale(ExecutionResults results) {
    assertThat(results).hasNumberOfSucceededTests(2);
}
```

It also allows running a specific method and passing additional parameters. If you think that this is something interesting for the Pioneer test suite, I can go ahead and add tests for the `@TestKitTest` annotation and also adapt the rest of the suite.

Proposed commit message:

```

${references}: ${issues}
PR: ${pull-request}
```

---
**PR checklist**

Documentation (general)

Documentation (new extension)

Code

Contributing

---

### DIFF
--- a/src/test/java/org/junitpioneer/jupiter/DefaultLocaleTests.java
+++ b/src/test/java/org/junitpioneer/jupiter/DefaultLocaleTests.java
@@ -11,8 +11,6 @@
 package org.junitpioneer.jupiter;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.junitpioneer.testkit.PioneerTestKit.executeTestClass;
-import static org.junitpioneer.testkit.PioneerTestKit.executeTestMethod;
 import static org.junitpioneer.testkit.assertion.PioneerAssert.assertThat;
 
 import java.util.Locale;
@@ -26,6 +24,7 @@ import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtensionConfigurationException;
 import org.junitpioneer.testkit.ExecutionResults;
+import org.junitpioneer.testkit.TestKitTest;
 
 @DisplayName("DefaultLocale extension")
 class DefaultLocaleTests {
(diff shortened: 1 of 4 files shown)


Bukama:
Hey,
thank you for opening this PR and sorry for not getting some response yet.

I'm personally not sure yet, if I like or like not this PR :D
I see the advantages of lesser writing and clearly identifying those somehow special tests.

On the other hand this somehow hides the `AAA (arrange - act - assert)` of testing as in the method body's you only have the `assert` step left, while the `act` is hidden in the annotation. On the other hand one can argue that this a special case anyway, because the `arrange` part (here the method to be executed) is somewhere else anyway.

And because of that I now decide that I like the request :D

Bukama (approved):

Michael1993:
This is an interesting idea but my personal opinion is that we should not merge this.

My opinion:
Tests verify behaviour. This is usually done in the three steps Matthias listed above (AAA or Given-When-Then). Tests are also a form of documentation and therefore be as simple as possible to understand but not simpler.

To help with this, we often move configuration (the "arrange" part) *out of the tests themselves*, e.g.: into `@BeforeEach` or `@BeforeAll` methods. Sometimes we can simplify so much that the tests themselves only have the `act` and the `assert` parts. But it *should stop there* - because the `act` is the behaviour and the `assert` is the verification. You can't extract them from the tests because then it would be only behaviour or verification.

In this case, I think executing the method or class is not a configuration or input of the test - it's the behaviour itself that we want to verify.

If you think we should discuss further, please open an issue about this.

nipafx:
When I first saw the proposed changes I went "uh, ahh" (no really, check the stream recording!) because this looks sooo smooth. But I quickly had the same reaction as @Bukama and @Michael1993. It feels like we're gutting the test and are moving its soul out of the method body and into an annotation. Also, you end up with the same amount of code (variable ~> parameter, class under test gets moved, method call ~> annotation).

So as much as I like the look of this proposal, I will close the PR.

filiphr:
Thanks for the feedback. I saw just watched the stream as well 😄 . The amount of saved characters is indeed more or less the same, only the imports are less now.

And I don't mind this being closed. I had an idea to try something out and wanted to see your opinion.

Issue #474, opened 2021-05-03
Restructure build matrix


nipafx:
Our current build configuration operates on the following sets:

* operating systems `OS` = { Ubuntu, Windows, macOs }
* Java versions `J` = { 8, 11, 16 }
* modularity `M` = { `true`, `false` }
* JUnit versions `U` = { 5.7.0, 5.7.1, 5.8.0-M1 }

But we don't build all possible combinations `OS x J x M x U`, which would be 54 builds and slightly above the [20 builds GitHub runs in parallel for free](https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration#usage-limits). Even if we upgrade to team, which raises the limit to 60 (and costs 4$/member/month), we'd be just one additional Java or JUnit version away from going overboard.

Instead we build:

* `OS x J x { false } x { 5.7.1 }` in `main-build.yml` (9 builds)
* `OS x { 11, 16 } x { true } x { 5.7.1 }` in `module-build.yml` (6 builds)
* `{ Ubuntu } x { 8 } x { false } x U` in `junit5-build.yml` (3 builds)

I find that combination of builds somewhat arbitrary. Also, just one more Java or JUnit version and we blow the limit of 20 builds and need to wait twice as long (worst case) for the entire set to finish.

Can we find a better way to create builds?

nipafx:
An obvious solution would be to define a baseline combination and then explore each set from there, keeping the others fixed, i.e.:

* `{ Ubuntu } x { 8 } x { false } x { 5.7.1 }` as baseline (1 build)
* `OS x { 8 } x { false } x { 5.7.1 }` (`|OS|` builds)
* `{ Ubuntu } x J x { false } x { 5.7.1 }` (`|J|` builds)
* `{ Ubuntu } x { 8 } x M x { 5.7.1 }` (2 builds)
* `{ Ubuntu } x { 8 } x { false } x U` (`|U|` builds)

This would result in `3 + |OS| + |J| + |U|` builds, which means it's not gonna tip over 20 any time soon. It's also structurally pleasing (very important!).

But does it miss important combinations? One could argue that testing on all `OS x J` is kinda important...

aepfli:
I would also evaluate what do we really need on a merge request and what can be handled as an action on a merge into develop.

Generally speaking it is nice, if everything runs during a merge request - but is it really needed or is it needed all the time?

[...]

PR #495, opened 2021-06-01
Restructure build (#474)

aepfli:
Unifying builds into one pipeline with configurable options.
Removing all other now unnecessary builds.
Adding experimental builds.

Closes: #474

Proposed commit message:

```
Restructure build into a mono build (#474 / #495)

Instead of multiple smaller builds, we introduces one mono build. This
allows us to provide a short and useful feedback cycle for
contributors, but ensures safety regarding merging for maintainers.
As for maintainers it does not matter how long our build runs before
a merge, and in the end it does not matter if it takes hours.

Furthermore our release build is now part of the pipeline and
we do not need to wait for another build for releasing.

Closes: #474 
PR: #495
```

---
**PR checklist**

Documentation (general)

Documentation (new extension)

Code

Contributing

---


.github/workflows/build.yml
```diff
@@ -0,0 +1,265 @@
...
+  basic:
+    runs-on: ${{ matrix.os }}-latest
+    needs: validation
+    timeout-minutes: 15
+    strategy:
+      matrix:
+        # test against latest update of each major Java version, as well as specific updates of LTS versions:
+        java: [ 8 ]
+        os: [ubuntu, macos, windows]
+    name: with Java ${{ matrix.java }} on ${{ matrix.os }}
```
aepfli, line 60: maybe we should just call that `Basic on ${{ matrix.os }}`

beatngu13:
So far, I haven't had time to take a closer look – sorry! ✌️

What do you think of a _short_ description of that build/PR process in `CONTRIBUTING.md`? (New) contributors might be a bit confused when they see a failing check, even if the job/step names are already quite descriptive.

.github/workflows/build.yml
Bukama: But then you don't see that Java version "Basic" is

aepfli:
it might be even worth to not break the build but set up a dependent build step and add a stage `ready to merge` at the end, like we had for before for `buildOk`. we can than setup the project to rely on that build before merging. which might be even better than a failing build. This seems to be less intrusive and easier to understand.

aepfli:
Okay i tried with "needed workflows" but as soon as one is skipped on purpose, it will still not block a merge.

I will adapt the CONTRIBUTORS.MD as i see no better way arround this for for now

nipafx:
I really like the entire setup - it looks amazing in GitHub's workflow visualization. Great job, @aepfli!

Only thing I want to find a better solution for is how to prevent premature merges (i.e. before running the full build). At the moment, the full build is triggered with by adding a label _merge-ready_ and because that can be somewhat easily forgotten, the step _Maintainer Action is Missing before Merge_ checks that and fails if it's absent. It works, but I don't like the non-technical implications. Contributors will get mails for failing builds and the PR overview will be mostly ❌. That's a lot of negative feedback even though nobody made a mistake.

Are there other options we can explore? What about defaulting to leaving the PR in draft mode (which runs the small build suite) until all is said and done and then switch to "proper" PR (which runs all the builds)? If we find no other option, we could also simply remove that step and hope that we don't forget the label too often (adding a checkbox to the default PR message would help).

aepfli:
based on @filiphr input (https://github.com/marketplace/actions/github-checks) i am experimenting with this action.

We do have multiple options here, like:
- set it to in_progress:
    Will never finish might be also bad feedback
- set to action_needed:
    Will cause an email still the question if we can work with that see https://github.com/aepfli/junit-pioneer/pull/11
- set it to neutral:
    will not prevent a merge, it will be just grey :(

aepfli:
okay for `action_required` i did not receive an email. which is good - and at least resolves the feedback for the contributors. but it will still  show an red `x` beside the PR

Bukama:
TBH I don't see in what state this is and what's your final thoughts after your last concerns :/

aepfli:
My last concerns summarized:

- Action needed still shows a red X on the merge request in the overview - but it does not send an email
- Label based approach works, but as soon as the label was added once, our condition is always evaluating to true. I would leave this for now, and define this as a future improvement.
- The build works now as intended, and i would suggest to give this a go. As always testing CI stuff is hard, so i expect issues, but i did not find any in my fork for now.

.github/workflows/build.yml
```diff
@@ -0,0 +1,271 @@
...
+          feature: ${{ env.EXPERIMENTAL_JAVA }}
+      - uses: actions/setup-java@v2
+        with:
+          java-version: ${{ env.JDK_VERSION }}
+          distribution: jdkfile
+          jdkFile: ${{ env.JDK_FILE }}
+      - name: Prepare JDK_EXPERIMENTAL env var
+        shell: bash
+        run: echo "JDK_EXPERIMENTAL=$JAVA_HOME" >> $GITHUB_ENV
+      - uses: actions/setup-java@v2
```
beatngu13, line 204: Hmmm, why this additional Java setup with v16 and AdoptOpenJDK?

.github/workflows/build.yml
```diff
@@ -0,0 +1,271 @@
...
+      - name: Gradle toolchains
+        uses: eskatos/gradle-command-action@v1
+        with:
+          gradle-version: rc
+          arguments: -Porg.gradle.java.installations.auto-download=false javaToolchains 
+      - name: Gradle build
+        uses: eskatos/gradle-command-action@v1
+        with:
+          gradle-version: rc
+          arguments: --refresh-dependencies -PexperimentalJavaVersion=17 -PmodularBuild=${{ matrix.modular }} -PjunitVersion=${{ matrix.junit-version }} -Porg.gradle.java.installations.auto-download=false --stacktrace --scan clean build
```
beatngu13, line 223:
```suggestion
          arguments: --refresh-dependencies -PexperimentalJavaVersion=${{ env.EXPERIMENTAL_JAVA }} -PmodularBuild=${{ matrix.modular }} -PjunitVersion=${{ matrix.junit-version }} -Porg.gradle.java.installations.auto-download=false --stacktrace --scan clean build
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,271 @@
...
+    if: github.event.inputs.version != '' && github.event.inputs.releasing == 'true'
+    runs-on: ubuntu-latest
+    needs: full-featured
+    timeout-minutes: 15
+    name: Release Pioneer into the wild
+    steps:
+      - name: Check out repo
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: '0'
```
beatngu13, line 237: A comment would be helpful – just like in the case of Sonar – explaining why unshallow is needed (I also can't remember right now). Also, above we use an integer, here string.

.github/workflows/build.yml
```diff
@@ -0,0 +1,271 @@
...
+          # used by `io.github.gradle-nexus.publish-plugin` to publish to Maven Central
+          ORG_GRADLE_PROJECT_sonatypeUsername: ${{ secrets.NEXUS_TOKEN_USERNAME }}
+          ORG_GRADLE_PROJECT_sonatypePassword: ${{ secrets.NEXUS_TOKEN_PASSWORD }}
+          # defines released version according to GitHub Action input
+          ORG_GRADLE_PROJECT_version: ${{ github.event.inputs.version }}
+        with:
+          arguments: publishToSonatype closeAndReleaseSonatypeStagingRepository githubRelease
+
+  # After our release, we also need to trigger an update for our website build.
+  updateWebsite:
```
beatngu13, line 259: This job key uses camel case, but there's also e.g. `full-featured`.

.github/workflows/build.yml
```diff
@@ -0,0 +1,271 @@
...
+    runs-on: ubuntu-latest
+    needs: release
+    timeout-minutes: 15
```
beatngu13, line 262: All timeouts seem to be 15 min, maybe define a envvar `DEFAULT_JOB_TIMEOUT`?

build.gradle.kts
```diff
@@ -23,13 +23,21 @@ plugins.withType<JavaPlugin>().configureEach {
 group = "org.junit-pioneer"
 description = "JUnit 5 Extension Pack"
 
-val modularBuild = findProperty("modularBuild") != null;
+val modularBuild : String by project;
+val experimentalJavaVersion : String? by project;
+val experimentalBuild: Boolean = experimentalJavaVersion?.isNotEmpty()?: false
```
beatngu13, line 28:
```suggestion
val experimentalBuild: Boolean = experimentalJavaVersion?.isNotEmpty() ?: false
```

build.gradle.kts
```diff
@@ -23,13 +23,21 @@ plugins.withType<JavaPlugin>().configureEach {
 group = "org.junit-pioneer"
 description = "JUnit 5 Extension Pack"
 
-val modularBuild = findProperty("modularBuild") != null;
+val modularBuild : String by project;
+val experimentalJavaVersion : String? by project;
```
beatngu13, line 27: Elsewhere, we don't use semicolons.

build.gradle.kts
```diff
@@ -23,13 +23,21 @@ plugins.withType<JavaPlugin>().configureEach {
...
+val experimentalBuild: Boolean = experimentalJavaVersion?.isNotEmpty()?: false
 
 java {
-	if(modularBuild) {
-		sourceCompatibility = JavaVersion.VERSION_11
+	if(experimentalBuild) {
```
beatngu13, line 31:
```suggestion
	if (experimentalBuild) {
```

build.gradle.kts
```diff
@@ -23,13 +23,21 @@ plugins.withType<JavaPlugin>().configureEach {
...
+		toolchain {
+			languageVersion.set(JavaLanguageVersion.of(experimentalJavaVersion))
+		}
 	} else {
-		sourceCompatibility = JavaVersion.VERSION_1_8
+		if(modularBuild.toBoolean()) {
```
beatngu13, line 36:
```suggestion
		if (modularBuild.toBoolean()) {
```

(We need a Spotless config for Kotlin, just like we used to have with Groovy: #484)

build.gradle.kts
```diff
@@ -195,6 +203,10 @@ tasks {
 	}
 
 	test {
+		
+    	configure<JacocoTaskExtension> {
```
beatngu13, line 207:
```suggestion
    	    	configure<JacocoTaskExtension> {
```

beatngu13 (changes requested):

.github/workflows/build.yml
```diff
@@ -0,0 +1,271 @@
+name: Main build
+
+env:
+  EXPERIMENTAL_JAVA: 17
+
+on:
+  # We want to trigger our builds all the time for the default branch and
+  # for tags starting with `v`.
```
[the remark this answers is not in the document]
aepfli: true, i should have looked better

.github/workflows/build.yml
```diff
@@ -0,0 +1,271 @@
...
+    branches: '*'
+    types: [labeled, opened, synchronize, reopened]
+  # We also utilize this pipeline for releasing. By providing a `version`
+  # and setting releasing to `true`, we can trigger a release.
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version string, e.g. 1.4.0'
+        required: true
+      releasing:
```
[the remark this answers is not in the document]
aepfli: would be, but this is some kind of safety feature, you have to double set something, so this happening by accident is reduced

.github/workflows/build.yml
```diff
@@ -0,0 +1,271 @@
...
+  # only build this on our default java version and for all operating systems
+  # we are supporting. This allows for a fast execution and fast feedback, the
+  # extended tests are run later on some conditions.
+  basic:
+    runs-on: ${{ matrix.os }}-latest
+    needs: validation
+    timeout-minutes: 15
+    strategy:
+      matrix:
+        # test against latest update of each major Java version, as well as specific updates of LTS versions:
```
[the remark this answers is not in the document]
aepfli:
```suggestion
```

.github/workflows/build.yml
aepfli: gradle is not working with ea builds - it will fail, so we start gradle with a suported java version, but execute the build on the experimental version

.github/workflows/build.yml
aepfli:
```suggestion
          # Sonar needs the whole git history for issue assignment
          fetch-depth: '0'
```

.github/workflows/build.yml
aepfli:
```suggestion
  update-website:
```

.github/workflows/build.yml
aepfli: that is a good idea - i will add this as soon as i have time

.github/workflows/build.yml
```diff
@@ -0,0 +1,271 @@
...
+    runs-on: ubuntu-latest
+    name: "Maintainer Action is Missing before Merge"
+    needs: basic
+    steps:
+      - uses: LouisBrunner/checks-action@v1.1.1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          name: Maintainer Action required, but your PR looks great
+          conclusion: action_required
+          action_url: "http://google.com" #replace this with a link to junit-pioneer contributing.md with a text about this
```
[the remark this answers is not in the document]
aepfli:
```suggestion
          action_url: "https://github.com/junit-pioneer/junit-pioneer/blob/main/CONTRIBUTING.md"
```

.github/workflows/build.yml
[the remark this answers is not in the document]
beatngu13: Let's go one step further and implement a [two-man rule](https://en.wikipedia.org/wiki/Two-man_rule). 😉

Personally, I wouldn't need this, but it's fine.

.github/workflows/build.yml
beatngu13: Would a comment make sense then?

.github/workflows/build.yml
beatngu13: I understand why we need it for the `sonarcloud` job, but why for the `release` job?

aepfli: for changelog generation

Bukama (approved):
Everything green :D
So technical it works, but @beatngu13  and @aepfli I ask you to finish the open points.

Michael1993 (approved):
Everything is green. 😄

build.gradle.kts
aepfli:
```suggestion
val modularBuild : String by project
val experimentalJavaVersion : String? by project
```

nipafx:
I really liked the idea to decide between basic and full build based on the PR's draft state, but that requires all PRs to be draft by default and most of their life time. This is counter to how they are used elsewhere on GitHub (and even in this project so far) and will probably lead to most new contributors opening "regular" PRs, which means the get the full build. I'm not sure that's such a big issue: most contributors aren't new and we can easily turn the PR into a draft and let them know about out practice; then again, anything non-standard will always lead to friction. For now we'll not use this approach.

.github/workflows/build.yml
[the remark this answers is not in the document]
nipafx: Shipkit doesn't push commits, just tags. If that triggers a _push_ event, we need to make sure we don't build on new _v*_ tags. Or really any tags. Then again, if a new commit gets pushed, no matter what tags it comes with, the build needs to be executed.

.github/workflows/build.yml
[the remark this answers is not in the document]
nipafx: The reason for having that boolean flag is that I found it better to make explicit whether one wants to release or not.

aepfli:
Here is a little overview of different actions for different builds on GitHub pull request overview - https://github.com/aepfli/junit-pioneer/pulls

Here are my thoughts on those:

- action_required - here is the overview of the action requiered https://github.com/aepfli/junit-pioneer/pull/12/checks?check_run_id=3001483599 - it is a red cross but no email

- with queued, it will always stay yellow, if that is a better solution i am not sure :)

.github/workflows/build.yml
```diff
@@ -0,0 +1,271 @@
...
+  maintainer-action-missing:
+    if: "!contains(github.event.pull_request.labels.*.name, 'merge-ready') && github.event.pull_request"
+    runs-on: ubuntu-latest
+    name: "Maintainer Action is Missing before Merge"
+    needs: basic
+    steps:
+      - uses: LouisBrunner/checks-action@v1.1.1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          name: Maintainer Action required, but your PR looks great
```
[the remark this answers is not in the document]
nipafx: What @beatngu13 said!

build.gradle.kts
nipafx: Fixed.

nipafx:
It's only us maintainers who can merge a PR and maybe we can just remember to apply the label before actually merging. So let's go with the simplest solution for now: No mechanism preventing accidental merges that didn't get a full build. If it turns out that we make too many mistakes, we can still add something later.

nipafx:
As far as I can tall, only [this comment](https://github.com/junit-pioneer/junit-pioneer/pull/495#discussion_r664754199) needs to be resolved before this can be merged. Can you take a look, @aepfli? 😁

.github/workflows/build.yml
nipafx: Agreed, lets leave the Java version in there.

.github/workflows/build.yml
```diff
@@ -0,0 +1,271 @@
...
+        run: ./gradlew jacocoTestReport sonarqube --stacktrace -i
+
+  # An Action that should signalize that we still need some mainter action, before
+  # we can merge an pull request. The Action needed is defined in the full-featured
+  # steps and this is just a reminder for us maintainer to prevent unintentional
+  # merges.
+  maintainer-action-missing:
+    if: "!contains(github.event.pull_request.labels.*.name, 'merge-ready') && github.event.pull_request"
+    runs-on: ubuntu-latest
+    name: "Maintainer Action is Missing before Merge"
```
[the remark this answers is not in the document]
nipafx: That section got removed.

.github/workflows/build.yml
nipafx: I added a comment.

.github/workflows/build.yml
nipafx: I tried, it lead to the build not starting...

nipafx:
@beatngu13 Can you take a look whether the changes you requested are done?

.github/workflows/build.yml
beatngu13: Thx for trying it out, didn't know/expect that.

As pointed out by @aepfli during the stream, I believe we can use [`defaults`](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#defaults), "[a] map of default settings that will apply to all jobs in the workflow".

```yaml
defaults:
  timeout-in-minutes: 15
```

This way, we should be able to remove the timeouts from all jobs.

beatngu13 (approved):

beatngu13:
One optional improvement we could try out:

https://github.com/junit-pioneer/junit-pioneer/pull/495#discussion_r667385417

But anyway, great job – approved! 🚀

.github/workflows/build.yml
nipafx: See 7304aae.

nipafx (approved):
LGTM

PR #496, opened 2021-06-01
Align use of extension point

### LINKED ISSUE #479: Enclosing classes vs Before/AfterAll
In #473/#476 we removed the search for annotations on enclosing classes because it's handled by the Before/AfterAll callbacks. That made me wonder whether there are other extensions that misuse `PioneerAnnotationUtils::findClosestEnclosingAnnotation`. I looked at two extensions:

* Issue extension uses it correctly because it needs to file a report entry for each test based on the annotation on some enclosing class.
* Default time zone (and I assume locale as well) use it correctly because they don't implement the Before/AfterAll extension points. That means the classes are not set up by Jupiter and so the method needs to go hunting for enclosing annotations. This can be refactored quite easily to the all callbacks without looking for enclosing (although one test fails, then).

Questions:

* do we prefer one approach over the other?
* do we want to refactor towards that?

nipafx:
Closes #479.

Proposed commit message:

```
Align use of extension point (#479 / #496)

If an extension needs to search for enclosing elements, there are
basically two ways to do so: 

1. Combine `BeforeAllCallback`/`AfterAllCallback` with
`BeforeEachCallback`/`AfterEachCallback`.
2. Use`BeforeEachCallback`/`AfterEachCallback` and look up enclosing
elements (e.g. via
`PioneerAnnotationUtils#findClosestEnclosingAnnotation(...)`).

We agreed on approach no. 1 since it enables the use of extensions in
setup/teardown methods. This PR aligns the use of extension points to
guarantee consistent behavior across different extensions.

Closes: #479
PR: #496
```

---
**PR checklist**

Documentation (general)

Documentation (new extension)

Code

Contributing

---


nipafx:
What's missing:

* documenting the extension scope in their respective documentations
* documenting the general approach in `CONTRIBUTING.md`
* checking all extensions - checked:
	* [ ] system properties and environment variables
	* [ ] default locale and time zone
* checking means:
	* verify behavior re before/after all
	* potentially add tests for resetting

beatngu13:
@nipafx did a rebase, addressed the remaining tasks, and wrote a commit message. ✌️

nipafx:
Great work, thank you @beatngu13! I changed a few things - see commit messages.

Bukama (approved):
I like it, especially the addition in the docs 👍

nipafx:
I think we wanted to model the tests of DefaulTimeZone after DefaultLocale before merging.

src/test/java/org/junitpioneer/jupiter/DefaultLocaleTests.java
```diff
@@ -173,37 +157,83 @@ void shouldSetLocaleFromMethodOfNestedClass() {
...
+			// still, better to assert the value was actually set
+			assertThat(Locale.getDefault().getLanguage()).isEqualTo("en");
+		}
+
+	}
+
+	@Nested
+	@DefaultLocale(language = "en")
+	@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+	class ResettingDefaultNestedLocaleTests {
```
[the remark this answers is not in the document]
nipafx: I think these aren't needed. `ResettingDefaultLocaleTests` already checks whether executing another test class (namely `ClassLevelResetCase`) resets the old state.

src/test/java/org/junitpioneer/jupiter/DefaultLocaleTests.java
```diff
@@ -173,37 +157,83 @@ void shouldSetLocaleFromMethodOfNestedClass() {
...
+		@BeforeAll
+		void setsBeforeBeforeEachMethod() {
+			assertThat(Locale.getDefault().getLanguage()).isEqualTo("en");
+		}
+
+		@Test
+		@DefaultLocale(language = "de")
+		void setForTestMethod() {
+			// only here to set the locale, so another test can verify whether it was reset;
+			// still, better to assert the value was actually set
```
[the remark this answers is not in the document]
nipafx: No, that's fine. The assertion is redundant with other tests (which already check whether the annotation correctly sets the value), but it shouldn't be removed "just to be safe".

docs/default-locale-timezone.adoc
```diff
@@ -136,8 +136,3 @@ Tests that cover code that reads or writes the default locale or time zone need
 * `@WritesDefaultTimeZone`
 
 Tests annotated in this way will never execute in parallel with tests annotated with `@DefaultLocale` or `@DefaultTimeZone`.
-
-== Extension Scope
```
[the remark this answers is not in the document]
nipafx: Interesting observation. [I replied on the issue](https://github.com/junit-pioneer/junit-pioneer/issues/479#issuecomment-968255705).

nipafx:
At least the tests need to be aligned before this is merge-ready. Also, [this conversation](https://github.com/junit-pioneer/junit-pioneer/issues/479#issuecomment-968255705) needs to be resolved.

Bukama (changes requested):
Merge conflict needs to be resolved

nipafx:
@beatngu13, you still working on this? 😬

nipafx:
No worries, I'll try to find some time (during JavaLand?) to go over this and merge it.

Please don't delete the branch.

beatngu13:
I think I did it … proposed commit message:

```text
Align use of extension point (#479 / #496)

If an extension needs to search for enclosing elements, there are
basically two ways to do so: 

1. Combine `BeforeAllCallback`/`AfterAllCallback` with
`BeforeEachCallback`/`AfterEachCallback`.
2. Use`BeforeEachCallback`/`AfterEachCallback` and look up enclosing
elements (e.g. via
`PioneerAnnotationUtils#findClosestEnclosingAnnotation(...)`).

We agreed on approach no. 2 since it appears to be less error-prone
and cover more common use cases.

This PR aligns the use of extension points to guarantee consistent
behavior across different extensions.

Closes: #479
PR: #496
```

src/main/java/org/junitpioneer/internal/PioneerUtils.java
```diff
@@ -82,6 +84,22 @@ private PioneerUtils() {
...
+	 * Find all (parent) {@code ExtensionContext}s via {@link ExtensionContext#getParent()}.
+	 *
+	 * @param context the context for which to find all (parent) contexts; never {@code null}
+	 * @return a list of all contexts, including the given context; never {@code null} or empty
+	 */
+	public static List<ExtensionContext> findAllContexts(ExtensionContext context) {
+		List<ExtensionContext> contexts = new ArrayList<>();
+		Optional<ExtensionContext> currentContext = Optional.of(context);
+		do {
+			currentContext.ifPresent(contexts::add);
```
Michael1993, line 97: But `currentContext` is always `present`...

src/main/java/org/junitpioneer/internal/PioneerUtils.java
```diff
@@ -82,6 +84,22 @@ private PioneerUtils() {
...
+	/**
+	 * Find all (parent) {@code ExtensionContext}s via {@link ExtensionContext#getParent()}.
+	 *
+	 * @param context the context for which to find all (parent) contexts; never {@code null}
+	 * @return a list of all contexts, including the given context; never {@code null} or empty
+	 */
+	public static List<ExtensionContext> findAllContexts(ExtensionContext context) {
+		List<ExtensionContext> contexts = new ArrayList<>();
+		Optional<ExtensionContext> currentContext = Optional.of(context);
+		do {
```
Michael1993, line 96: I don't like `do...while` loops, but it looks super appropriate here. Dammit. ☹️

src/main/java/org/junitpioneer/jupiter/AbstractEntryBasedExtension.java
```diff
@@ -139,12 +132,8 @@ private void setEntries(Map<K, V> entriesToSet) {
...
 	@Override
 	public void afterEach(ExtensionContext context) throws Exception {
-		restoreOriginalEntries(context);
-	}
-
-	@Override
-	public void afterAll(ExtensionContext context) throws Exception {
-		restoreOriginalEntries(context);
+		List<ExtensionContext> contexts = PioneerUtils.findAllContexts(context);
+		contexts.forEach(this::restoreOriginalEntries);
```
Michael1993, line 136: I think these two lines could be merged.
```java
PioneerUtils.findAllContexts(context).forEach(this::restoreOriginalEntries)
```

src/main/java/org/junitpioneer/jupiter/AbstractEntryBasedExtension.java
```diff
@@ -43,60 +45,52 @@
...
-	@Override
-	public void beforeAll(ExtensionContext context) {
-		clearAndSetEntries(context);
-	}
+		implements BeforeEachCallback, AfterEachCallback {
 
 	@Override
 	public void beforeEach(ExtensionContext context) {
-		clearAndSetEntries(context);
+		List<ExtensionContext> contexts = PioneerUtils.findAllContexts(context);
```
Michael1993, line 52: Honestly, just return with a `Stream` from findAllContexts, so you can write:
```java
PioneerUtils.findAllContexts(context)
    .sorted(Collections.reverseOrder()) // Could be static imported, so it's just reverseOrder()
    .forEach(this::clearAndSetEntries);
```

Michael1993 (approved):
LGTM - I do have a couple nitpicks about it.

Michael1993 (changes requested):
Actually, the build is platform-dependent now? How did that happen?

src/main/java/org/junitpioneer/jupiter/AbstractEntryBasedExtension.java
beatngu13: [Javadoc](https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#reverseOrder--):

>Returns a comparator that imposes the reverse of the _natural ordering_ on a collection of objects that implement the `Comparable` interface.

We don't want the list to be sorted (and even if, `ExtensionContext` isn't comparable). Instead, we want to reverse the entire list. This isn't straightforward for streams (see ["Java 8 stream reverse order"](https://stackoverflow.com/q/24010109)), so I decided to return a list.

OK for you?

Michael1993: My bad 😅

src/main/java/org/junitpioneer/internal/PioneerUtils.java
beatngu13: Alternatively recursion?

```java
public static List<ExtensionContext> findAllContexts(ExtensionContext context) {
	List<ExtensionContext> allContexts = new ArrayList<>();
	allContexts.add(context);
	allContexts.addAll(context.getParent()
			.map(PioneerUtils::findAllContexts)
			.orElse(Collections.emptyList()));
	return allContexts;
}
```

Michael1993: Very fancy alternative - in both ways (very elaborate and elegant). I actually prefer the loop, because it's simpler (if uglier). I was just complaining a little 😇

beatngu13:
> Actually, the build is platform-dependent now? How did that happen?

I forgot this:

https://github.com/junit-pioneer/junit-pioneer/blob/30cca6e81bb0cff28f8c6c5ff7d6d6553fabcf27/src/test/java/org/junitpioneer/jupiter/DefaultTimeZoneTests.java#L38-L39

Apparently, the macOS job was using UTC, so the test default resulted in UTC+12. Fix:

https://github.com/junit-pioneer/junit-pioneer/pull/496/commits/30cca6e81bb0cff28f8c6c5ff7d6d6553fabcf27

src/main/java/org/junitpioneer/internal/PioneerUtils.java
beatngu13: Indeed, let's use recursion then. 😬

beatngu13 (approved):

src/main/java/org/junitpioneer/jupiter/AbstractEntryBasedExtension.java
```diff
@@ -43,60 +45,56 @@
...
+		implements BeforeEachCallback, AfterEachCallback {
 
 	@Override
 	public void beforeEach(ExtensionContext context) {
-		clearAndSetEntries(context);
+		/*
+		 * We cannot use PioneerAnnotationUtils#findAllEnclosingRepeatableAnnotations(ExtensionContext, Class) or the
+		 * like as clearing and setting might interfere. Therefore, we have to apply the extension from the outermost
+		 * to the innermost ExtensionContext.
+		 */
```
beatngu13, line 56: Is this comment clear to everyone? Consider the following:

```java
@SetSystemProperty(key = "foo", value = "42")
class MyTests {

    void testA() {
        // ...
    }

    @ClearSystemProperty(key = "foo")
    void testB() {
        // ...
    }

}
```

Note that we first clear and then set entries (here system properties):

https://github.com/junit-pioneer/junit-pioneer/blob/0ef4ca66fe5766b3419a7846104cf6aa1b788c09/src/main/java/org/junitpioneer/jupiter/AbstractEntryBasedExtension.java#L76-L77

(If we first set and then clear, the problem still persist with the inverse example, i.e., `@ClearSystemProperty` on the class and `@SetSystemProperty` on the method.)

Now when the `BeforeEachCallback` is invoked and we use `PioneerAnnotationUtils#findAllEnclosingRepeatableAnnotations(ExtensionContext, Class)`, we first clear `foo` and then set `foo`, which is not what the user wants.

Bukama: `
> 
> Note that we first clear and then set entries (here system properties):
> 

But den quoted docs tell otherway (from outer to inner, so suggest to first set, then clear)

Bukama (commented):
See my comment on the comment-question, if I'm the only one who does it wrong, see this as approved.

src/main/java/org/junitpioneer/jupiter/AbstractEntryBasedExtension.java
beatngu13: We do both, clear and set per `ExtensionContext`. `BeforeEachCallback` from outermost to innermost, `AfterEachCallback` from innermost to outermost.

beatngu13: Side note: I meant this from an implementation perspective. This is what the code does:

https://github.com/junit-pioneer/junit-pioneer/blob/0ef4ca66fe5766b3419a7846104cf6aa1b788c09/src/main/java/org/junitpioneer/jupiter/AbstractEntryBasedExtension.java#L76-L77

First clear, then set. From a user perspective, it is, of course, different. For the given example, as we clear and set per extension context:

* `BeforeEachCallback`: from outermost to innermost:
  1. Class-level / `MyTest`:
    * Clear: n/a
    * Set: `foo = 42`
  2. Method-level / `testB()`:
    * Clear: `foo`
    * Set: n/a
* `AfterEachCallback`: from innermost to outermost:
  1. Method-level / `testB()`:
    * Restore `foo = 42`
  2. Class-level / `MyTest`:
    * Restore `foo` to pre-test state

PR #505, opened 2021-07-25
Failed test disable other tests

nipafx:
First draft - what do you think?

---
**PR checklist**

Documentation (general)

Documentation (new extension)

Code

Contributing

---

### DIFF
--- a/src/main/java/org/junitpioneer/jupiter/DisableIfTestFails.java
+++ b/src/main/java/org/junitpioneer/jupiter/DisableIfTestFails.java
@@ -20,14 +20,53 @@ import java.lang.annotation.Target;
 
 import org.junit.jupiter.api.extension.ExtendWith;
 
+/**
+ * Disables all remaining tests in a container of one of them failed.
+ *
+ * By default, all exceptions (including assertions) will be lead to disabling remaining tests.
+ * To configure this in more detail, see {@link #with()} and {@link #onAssertion()}.
+ *
+ * This annotation can be (meta-)present on classes that contain a nested class. In that case, a
+ * failing test in the outer class will disable the nested class (if it runs later) and vice versa.
+ *
+ * This annotation can be (meta-)present on a class and/or its super types (classes or interfaces).
+ * In that case, the exception types given to {@link #with()} are merged and {@link #onAssertion()}
+ * is or'ed.
+ *
+ * But if a test fails in a specific class, only other tests in the corresponding container will
+ * be disabled. That means if...
+ *
+ * <ul>
+ *     <li>a class {@code SpecificTests} implements interface {@code Tests} and</li>
+ *     <li>{@code Tests} is annotated with {@code @DisableIfTestFails} and</li>
+ *     <li>a test in {@code SpecificTests} fails</li>
+ * </ul>
+ *
+ * ... then, only remaining tests in {@code SpecificTests} are disabled and other implementations
+ * of {@code Tests} remain unaffected, i.e. their tests are not disabled.
+ */
 @Target({ TYPE, ANNOTATION_TYPE })
 @Inherited
 @Retention(RetentionPolicy.RUNTIME)
 @ExtendWith(DisableIfTestFailsExtension.class)
 public @interface DisableIfTestFails {
 
+	/**
+	 * Configure on which exceptions remaining tests are disabled (defaults to "any exception").
+	 *
+	 * If {@code @DisableIfTestFails} is present multiple times (e.g. on multiple super types),
+	 * these exceptions are collected across all annotations, meaning if any of the mentioned
+	 * exceptions are thrown, the remaining tests are disabled.
+	 */
 	Class<? extends Throwable>[] with() default {};
 
+	/**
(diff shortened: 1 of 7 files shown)


src/test/java/org/junitpioneer/jupiter/DisableIfTestFailsTests.java
```diff
@@ -0,0 +1,245 @@
...
+	// (If they're executed in parallel, all tests might already have passed the condition evaluation
+	// extension point before the first test fails and so none would be disabled.)
+	// Hence use `@Execution(SAME_THREAD)`.
+
+	@TestInstance(PER_CLASS)
+	@DisableIfTestFails(with = IOException.class)
+	@Execution(SAME_THREAD)
+	static class ThreeTestsWithSecondFailing {
+
+		private int EXECUTION_COUNT;
```
Bukama, line 105: its a variable not a constant => lowercase

src/main/java/org/junitpioneer/jupiter/DisableIfTestFailsExtension.java
```diff
@@ -0,0 +1,133 @@
...
+	private static final String DISABLED_VALUE = "";
+
+	@Override
+	public ConditionEvaluationResult evaluateExecutionCondition(ExtensionContext context) {
+		boolean disabled = context.getStore(NAMESPACE).get(DISABLED_KEY) != null;
+		if (disabled)
+			return ConditionEvaluationResult.disabled("Another test threw one of the specified exceptions.");
+		else
+			return ConditionEvaluationResult.enabled("No test threw one of the specified exceptions (yet).");
+	}
```
Bukama, line 54: Why only disabled if another test threw an exception? Why not disable if another test fails for any reason?

Bukama (changes requested):

docs/cartesian-product.adoc
```diff
@@ -412,7 +412,7 @@ void wrongOrderParameters(int i, String string) {
 }
 
 @CartesianProductTest(factory = "resolveTestReporterParam")
-void conflictingParameters(String string, TestReporter info) {
+void conflictingParameters(String string, TestReporter configuration) {
```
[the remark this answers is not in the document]
nipafx: Good point, should be removed.

src/main/java/org/junitpioneer/jupiter/DisableIfTestFailsExtension.java
nipafx: That's not a bad point, I have to think about that. First impulse: "Because this exists to deactivate tests that make "no sense to run" because some infrastructure failed." But this might not actually be a good argument. 😊

nipafx: You're right that this scenario needs to be considered. I think assertion errors should be special cased and it should be possible to configure how the extension treats them. I got a work-in-progress commit lying around locally.

src/main/java/org/junitpioneer/jupiter/DisableIfTestFails.java
```diff
@@ -0,0 +1,72 @@
...
+
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.junit.jupiter.api.extension.ExtendWith;
+
+/**
+ * Disables all remaining tests in a container of one of them failed.
```
nipafx, line 24:
```suggestion
 * Disables all remaining tests in a container if one of them failed.
```

src/main/java/org/junitpioneer/jupiter/DisableIfTestFails.java
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * By default, all exceptions (including assertions) will be lead to disabling remaining tests.
```
nipafx, line 26:
```suggestion
 * By default, all exceptions (including assertions) will lead to disabling remaining tests.
```

nipafx:
Note to self: Make sure assumptions are correctly handled (i.e. a failing assumption probably shouldn't disable all other tests).

nipafx:
Lets run ALL THE BUILDS while I'm writing the docs.

src/main/java/org/junitpioneer/jupiter/DisableIfTestFailsExtension.java
nipafx: Implemented both, thanks for spotting that!

nipafx:
@Bukama I think I implemented your change request. If you agree and this is green, feel free to merge.

nipafx:
Proposed commit message:

```
Add @DisableIfTestFails extension (#467 / #505)

This extension disables all remaining tests in a container if one of
them failed. Failed assumptions are ignored (i.e. they don't disable
other tests) but assertions aren't by default (can be configured with
the `onAssertion` attribute). All exceptions lead to disabling
remaining tests (can be configured with the `with` attribute).

Closes: #467
PR: #505
```

Bukama (approved):
Thanks for the changes. I think this extension can help some people.

PR #519, opened 2021-09-04
Decouple CartesianArgumentsProvider from the Jupiter Arguments Provider

filiphr:
As I mentioned in the last stream (2nd of September). I had some ideas about decoupling the CartesianArgumentsProvider from the ArgumentsProvider. With this PR I am showcasing my idea.

The goal is to support both ArgumentsProvider. There is a new `@CartesianArgumentsSource` that would allow users to write custom `CartesianArgumentsProvider` by using it directly on parameters (the `@ArgumentsSource` does not allow this).

Proposed commit message:

```
Decouple CartesianArgumentsProvider from the Jupiter Arguments Provider (${issues} / ${pull-request}) [max 70 characters]

The CartesianArgumentsProvider supports the Jupiter AnnotationConsumer interface.
The CartesianArgumentsProvider has a method for providing arguments 
that accepts the ExtensionContext and the Parameter for which it provides arguments.


${references}: ${issues}
PR: ${pull-request}
```

---
**PR checklist**

Documentation (general)

Documentation (new extension)

Code

Contributing

---

### DIFF
--- a/src/main/java/org/junitpioneer/jupiter/cartesian/CartesianEnumArgumentsProvider.java
+++ b/src/main/java/org/junitpioneer/jupiter/cartesian/CartesianEnumArgumentsProvider.java
@@ -25,31 +25,27 @@ import org.junit.platform.commons.support.AnnotationSupport;
 
 /**
  * This is basically an enhanced copy of Jupiter's {@code EnumArgumentsProvider},
- * except it does NOT support {@code @ParameterizedTest} and consumes a {@code Parameter}
- * instead of an annotation.
+ * except it does NOT support {@code @ParameterizedTest} and implements {@link CartesianArgumentsProvider}
+ * for use with {@code @CartesianTest}.
+ * <p>
+ * This class does not implement {@code ArgumentsProvider} since the Jupiter's {@code EnumSource} should be used for that.
  */
 class CartesianEnumArgumentsProvider implements CartesianArgumentsProvider {
 
-	private CartesianTest.Enum enumSource;
-	private Class<?> parameterType;
-
 	@Override
-	public void accept(Parameter parameter) {
-		this.parameterType = parameter.getType();
-		if (!Enum.class.isAssignableFrom(this.parameterType))
+	public Stream<? extends Arguments> provideArguments(ExtensionContext context, Parameter parameter) {
+		Class<?> parameterType = parameter.getType();
+		if (!Enum.class.isAssignableFrom(parameterType))
 			throw new PreconditionViolationException(String
 					.format(
 						"Parameter of type %s must reference an Enum type (alternatively, use the annotation's 'value' attribute to specify the type explicitly)",
-						this.parameterType));
-		this.enumSource = AnnotationSupport
+						parameterType));
+		CartesianTest.Enum enumSource = AnnotationSupport
 				.findAnnotation(parameter, CartesianTest.Enum.class)
 				.orElseThrow(() -> new PreconditionViolationException(
 					"Parameter has to be annotated with " + CartesianTest.Enum.class.getName()));
-	}
 
-	@Override
-	public Stream<? extends Arguments> provideArguments(ExtensionContext context) {
-		Set<? extends Enum<?>> constants = getEnumConstants();
+		Set<? extends Enum<?>> constants = getEnumConstants(enumSource, parameterType);
 		CartesianTest.Enum.Mode mode = enumSource.mode();
 		String[] declaredConstantNames = enumSource.names();
 		if (declaredConstantNames.length > 0) {
@@ -63,16 +59,17 @@ class CartesianEnumArgumentsProvider implements CartesianArgumentsProvider {
 		return constants.stream().map(Arguments::of);
 	}
 
-	private <E extends Enum<E>> Set<? extends E> getEnumConstants() {
-		Class<E> enumClass = determineEnumClass();
+	private <E extends Enum<E>> Set<? extends E> getEnumConstants(CartesianTest.Enum enumSource,
+			Class<?> parameterType) {
+		Class<E> enumClass = determineEnumClass(enumSource, parameterType);
 		return EnumSet.allOf(enumClass);
 	}
 
 	@SuppressWarnings({ "unchecked", "rawtypes" })
-	private <E extends Enum<E>> Class<E> determineEnumClass() {
+	private <E extends Enum<E>> Class<E> determineEnumClass(CartesianTest.Enum enumSource, Class<?> parameterType) {
 		Class enumClass = enumSource.value();
 		if (enumClass.equals(NullEnum.class)) {
-			enumClass = this.parameterType;
+			enumClass = parameterType;
 		}
 		return enumClass;
 	}
--- a/src/main/java/org/junitpioneer/jupiter/cartesian/CartesianTest.java
+++ b/src/main/java/org/junitpioneer/jupiter/cartesian/CartesianTest.java
@@ -24,7 +24,6 @@ import java.util.regex.PatternSyntaxException;
 
 import org.junit.jupiter.api.TestTemplate;
 import org.junit.jupiter.api.extension.ExtendWith;
-import org.junit.jupiter.params.provider.ArgumentsSource;
 import org.junit.platform.commons.PreconditionViolationException;
 import org.junitpioneer.jupiter.cartesian.CartesianEnumArgumentsProvider.NullEnum;
 
@@ -85,7 +84,7 @@ public @interface CartesianTest {
 
 	@Retention(RetentionPolicy.RUNTIME)
 	@Target({ ElementType.PARAMETER, ElementType.ANNOTATION_TYPE })
-	@ArgumentsSource(CartesianValueArgumentsProvider.class)
+	@CartesianArgumentsSource(CartesianValueArgumentsProvider.class)
 	@interface Values {
 
 		/**
@@ -142,7 +141,7 @@ public @interface CartesianTest {
 
 	@Retention(RetentionPolicy.RUNTIME)
 	@Target({ ElementType.PARAMETER, ElementType.ANNOTATION_TYPE })
-	@ArgumentsSource(CartesianEnumArgumentsProvider.class)
+	@CartesianArgumentsSource(CartesianEnumArgumentsProvider.class)
 	@interface Enum {
 
 		/**
--- a/src/main/java/org/junitpioneer/jupiter/cartesian/CartesianValueArgumentsProvider.java
+++ b/src/main/java/org/junitpioneer/jupiter/cartesian/CartesianValueArgumentsProvider.java
@@ -14,25 +14,30 @@ import static java.util.stream.Collectors.toList;
 
 import java.lang.reflect.Array;
 import java.lang.reflect.Parameter;
+import java.util.Arrays;
 import java.util.List;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
 
 import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.support.AnnotationConsumer;
 import org.junit.platform.commons.PreconditionViolationException;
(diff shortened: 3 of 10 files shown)


Bukama (changes requested):
Hey @filiphr,
@Michael1993 and I were looking at and talking about your PR, but we both cound't understand the problem you are trying to adress in your PR. We thinkg this has two causes: There's no issue describing the topic the PR is for and we don't see any tests which cover what (we think) you describe in your commit message PR.

As you may have read in the contribution guide, we aim to have no PR without an underlying issue, which descripes the "cause/problem" we PR aims to solve. Also we aim to document every decision in written so everyone can follow decisions and change. For example I personally wasn't attending stream last week either, so your reference to this doesn't help me do understand the fix (?) you aim for.

This saying I ask you to open an issue describing the problem you are adressing with your PR and adding some tests to the PR which show the different behaviors of the `ArgumentsProvider`

Thank you very much!

filiphr:
Thanks for the feedback @Bukama.

>There's no issue describing the topic the PR is for and we don't see any tests which cover what (we think) you describe in your commit message PR.

There are no tests in the PR since there are tests for these things already.

e.g.

* [`CartesianTestExtensionTests$CaretesianValueSourceTestCases#poeticValues`](https://github.com/junit-pioneer/junit-pioneer/blob/bd6d99478cf2c53caef592ba3c42b4abbfe47165/src/test/java/org/junitpioneer/jupiter/cartesian/CartesianTestExtensionTests.java#L649) - Uses `CartesianTest.Values` which is `CartesianArgumentsSource`
* [`CartesianTestExtensionTests$ArgumentsSourceTestCases`](https://github.com/junit-pioneer/junit-pioneer/blob/bd6d99478cf2c53caef592ba3c42b4abbfe47165/src/test/java/org/junitpioneer/jupiter/cartesian/CartesianTestExtensionTests.java#L819) - Uses the different range source and a combination of range source and `CartesianTest.Value`. The Range source is defined only through `@ArgumentsSource`. However, its implementation `RangeSourceArgumentsProvider` implements `CartesianArgumentsProvider` as well.

------

>This saying I ask you to open an issue describing the problem you are adressing with your PR and adding some tests to the PR which show the different behaviors of the `ArgumentsProvider`

The goal of this PR was to do a refactoring that would make it simpler to implement a `CartesianArgumentsProvider`, especially if you need to implement  both as in the case with `RangeSourceArgumentsProvider`. This implementation is currently stateful due to the backwards compatibility, but once that is remove the implementation can be completely stateless.

I will create an issue with some more details and will add some more tests showcasing something else.

filiphr:
I've added some test cases to show how this can be used. One of the test cases is also failing due to https://github.com/junit-pioneer/junit-pioneer/issues/520#issuecomment-917612388

src/test/java/org/junitpioneer/jupiter/cartesian/CartesianTestExtensionTests.java
```diff
@@ -353,6 +360,38 @@ void testClassWithConstructor() {
...
+			void usesCustomArgumentsProviderOnTest() {
+				ExecutionResults results = PioneerTestKit
+						.executeTestMethodWithParameterTypes(CustomCartesianArgumentsProviderTestCases.class,
+							"customArgumentProvider", String.class);
+
+				assertThat(results).hasNumberOfDynamicallyRegisteredTests(6).hasNumberOfSucceededTests(6);
+				assertThat(results)
+						.hasNumberOfReportEntries(6)
+						.withValues("first(1)", "first(2)", "second(1)", "second(2)", "third(1)", "third(2)");
+			}
```
filiphr, line 391: According to https://github.com/junit-pioneer/junit-pioneer/issues/520#issuecomment-917619959 this is not yet supported. Which means that we could remove it. I've added this only as a showcase

src/main/java/org/junitpioneer/jupiter/cartesian/CartesianArgumentsSources.java
```diff
@@ -0,0 +1,41 @@
...
+ * optional since {@code @CartesianArgumentsSource} is a {@linkplain java.lang.annotation.Repeatable
+ * repeatable} annotation.
+ *
+ * @since 5.0
+ * @see CartesianArgumentsSource
+ */
+@Target(ElementType.PARAMETER)
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+public @interface CartesianArgumentsSources {
```
nipafx, line 33: This would allow the following, right?

```java
@CartesianTest
void myTest(
    @CartesianArgumentsSource(MyArgumentsProvider.class)
    @CartesianArgumentsSource(MyOtherArgumentsProvider.class)
    String argument);
```

I.e. having several argument source annotations on the same parameter? If so, that seems like a very rare case and I wouldn't want to add any complexity to support, particularly not in this initial (refactored) release.

src/main/java/org/junitpioneer/jupiter/cartesian/CartesianEnumArgumentsProvider.java
```diff
@@ -25,31 +25,27 @@
...
- * instead of an annotation.
+ * except it does NOT support {@code @ParameterizedTest} and implements {@link CartesianArgumentsProvider}
+ * for use with {@code @CartesianTest}.
+ * <p>
+ * This class does not implement {@code ArgumentsProvider} since the Jupiter's {@code EnumSource} should be used for that.
  */
 class CartesianEnumArgumentsProvider implements CartesianArgumentsProvider {
 
-	private CartesianTest.Enum enumSource;
-	private Class<?> parameterType;
```
nipafx, line 34: I really like the statelessness, that's an big plus of this PR in my book.

src/main/java/org/junitpioneer/jupiter/cartesian/CartesianTest.java
```diff
@@ -85,7 +84,7 @@
 
 	@Retention(RetentionPolicy.RUNTIME)
 	@Target({ ElementType.PARAMETER, ElementType.ANNOTATION_TYPE })
-	@ArgumentsSource(CartesianValueArgumentsProvider.class)
+	@CartesianArgumentsSource(CartesianValueArgumentsProvider.class)
```
nipafx, line 87: No longer being usable with `@ArgumentsSource` seems like a good thing, too, because these providers couldn't work with "just" the parameterized machinery anyway and so this removes a potential mistake that the user can make.

src/main/java/org/junitpioneer/jupiter/cartesian/CartesianValueArgumentsProvider.java
```diff
@@ -14,25 +14,29 @@
...
 
 /**
- * This is a slightly modified copy of Jupiter's {@code ValueSourceArgumentsProvider},
- * except it does NOT support {@code @ParameterizedTest} and can consume a {@code Parameter}
- * instead of an annotation.
+ * This is a slightly modified copy of Jupiter's {@code ValueArgumentsProvider},
+ * except it does NOT support {@code @ParameterizedTest} and implements {@link CartesianArgumentsProvider}
+ * for use with {@code @CartesianTest}.
+ * <p>
+ * This class does not implement {@code ArgumentsProvider} since the Jupiter's {@code ValueSource} should be used for that.
```
nipafx, line 32: That last line is probably not needed.

src/test/java/org/junitpioneer/jupiter/cartesian/CartesianTestExtensionTests.java
```diff
@@ -887,6 +926,25 @@ void shouldHaveTestInfo(@CartesianTest.Values(ints = { 1, 2 }) int i,
 
 	}
 
+	static class CustomCartesianArgumentsProviderTestCases {
+
+		@CartesianTest
+		@ReportEntry("{0}({1})")
+		void twoCustomCartesianArgumentProviders(
+				@CartesianArgumentsSource(FirstCustomCartesianArgumentsProvider.class) String string,
+				@CartesianArgumentsSource(SecondCustomCartesianArgumentsProvider.class) int integer) {
```
nipafx, line 935: Looks good, although maybe it's better if we take the same route as with the other annotations and make this an inner class of `@CartesianTest`, e.g. `@CartesianTest.Source`, then users can use either of these two variants:

```java
@CartesianTest.Source(FirstCustomCartesianArgumentsProvider.class) String string,
@Source(SecondCustomCartesianArgumentsProvider.class) int integer
```

nipafx (commented):
I think this is a good idea and unless I missed something I saw only upsides:

* stateless
* no risk of accidentally applying our extensions to `@ParameterizedTests`
* good path to our variant of `@ArgumentsSource`

src/test/java/org/junitpioneer/jupiter/cartesian/CartesianTestExtensionTests.java
scordio: What about `@CartesianTest.Arguments`? Similar pattern to `@CartesianEnumSource` --> `@CartesianTest.Enum` and `@CartesianValueSource` --> `@CartesianTest.Values`

filiphr: I understand your reasoning @nipafx and @scordio. I watched the stream and read the other issue where you decided about the inner classes.

However, I personally am more in favour of the current name `@CartesianArgumentsSource`. The reason for that is that people are already familiar with `ArgumentsSource`. In addition to that the target audience for the `@CartesianArgumentsSource` annotation is other library developers that would like to provide their own `CartesianArgumentsProvider`. The example in the test case is just an example of a potential use case. 

From an API point of view I think that it is better to keep it close to what people are used to. If it really has to be an inner class I think using `@CartesianTest.ArgumentsSource`,  i.e indicating an `ArgumentsSource` for `CartesianTest` fits better, although it is longer then `@CartesianArgumentsSource` and it will clash in a standalone mode.


`@CartesianTest.Source` or `@CartesianTest.Arguments` feels more like an end user annotations where people will provide the `source` or `arguments` values themselves, i.e. like in `@CartesianTest.Values`.

src/main/java/org/junitpioneer/jupiter/cartesian/CartesianTest.java
filiphr: Actually it is usable with both. e.g. the different `XXXRangeSource` annotations only define `@ArgumentsSource`. However, the `RangeSourceArgumentsProvider` implements the `CartesianArgumentsProvider`. 

However, with the current approach you can entirely opt in to only `@CartesianTest` by using the `@CartesianArgumentsSource`. This way users won't use your `@CartesianTest` annotation in `@ParametrizedTest`

src/main/java/org/junitpioneer/jupiter/cartesian/CartesianArgumentsSources.java
filiphr: Indeed, using it like in your example was the reason for having the `@Repeatable`. I've removed it for now

src/main/java/org/junitpioneer/jupiter/cartesian/CartesianValueArgumentsProvider.java
filiphr: It was mostly there for an explanation for the maintainers why it does not implement it. However, I have removed it. Also removed it from `CartesianEnumArgumentsProvider`

src/test/java/org/junitpioneer/jupiter/cartesian/CartesianTestExtensionTests.java
nipafx: Good argument, I'm convinced.

nipafx:
@Bukama I assume you're ok with the direction this took? If so, can you mark this as accepted? :)

filiphr:
Thanks for the review and looking into this 😄

filiphr:
Just FYI I pushed another commit because I forgot to change the documentation earlier

Bukama (approved):
@nipafx Yeah I am :)

PR #540, opened 2021-11-14
Add method-level sources to CartesianTest

Michael1993:
Forget the checklist and the commit message. I'll add them later when this is actually ready for merge. For now this is just an idea I need feedback on.

---

### DIFF
--- a/docs/cartesian-product.adoc
+++ b/docs/cartesian-product.adoc
@@ -304,16 +304,18 @@ For more information, please see the link:docs/range-sources[separate documentat
 
 You can annotate your test method to supply arguments to all parameters simultaneously.
 
-=== `@CartesianTest.Factory`
+=== `@CartesianTest.MethodFactory`
 
-`@CartesianTest.Factory` can be used to name a static factory method that supplies your arguments.
+`@CartesianTest.MethodFactory` can be used to name a static factory method that supplies your arguments.
 The `value` annotation parameter is mandatory.
 Just like with JUnit's `@MethodSource`, you can specify the factory method with its fully-qualified name (including the class), e.g. `com.example.Class#factory`.
-This method must return `CartesianMethodArgumentsProvider.Sets`.
-`Sets` is a helper class, specifically for creating sets for `@CartesianTest`.
-To create the test data, instantiate a `new Sets()` then use the `add()` method to register the values for the parameters.
-When calling `add`, you have to pass in all the arguments you need for the corresponding parameter.
-For convenience, all `add()` methods return with your `Sets` instance, so you can chain `add()` calls.
+This method must return `ArgumentSets`.
+
+`ArgumentSets` is a helper class, specifically for creating sets for `@CartesianTest`.
+To create the test data, instantiate with the static factory method `argumentsForFirstParameter`, then call the `addValuesForNextParameter` method once per additional parameter in the order in which the they appear in the test method.
+In each call, pass in all values for the corresponding parameter.
+For convenience, all methods return with your `ArgumentSets` instance, so you can chain `add()` calls.
+If you want to create an initially-empty `ArgumentSets`, call the static factory method `create()`.
 
 Let's look at an example.
 
@@ -322,16 +324,16 @@ Let's look at an example.
 
 class MyCartesianTestClass {
 	@CartesianTest
-	@CartesianTest.Factory("setFactory")
+	@CartesianTest.MethodFactory("setFactory")
 	void testMethod(String string, Class<?> clazz, TimeUnit unit) {
 		// passing test code
 	}
 
-	static Sets setFactory() {
-		return new Sets()
-			.add("Alpha", "Omega")
-			.add(Runnable.class, Cloneable.class, Predicate.class)
-			.add(TimeUnit.DAYS, TimeUnit.HOURS);
+	static ArgumentSets setFactory() {
+		return ArgumentSets
+				.argumentsForFirstParameter("Alpha", "Omega")
+				.argumentsForNextParameter(Runnable.class, Cloneable.class, Predicate.class)
+				.argumentsForNextParameter(TimeUnit.DAYS, TimeUnit.HOURS);
 	}
 }
 ----
@@ -366,21 +368,21 @@ You can reuse the same argument provider method multiple times.
 ----
 class MyCartesianTestClass {
 	@CartesianTest
-	@CartesianTest.Factory("provideArguments")
+	@CartesianTest.MethodFactory("provideArguments")
 	void testNeedingArguments(String string, int i) {
 		// passing test code
 	}
 
 	@CartesianTest
-	@CartesianTest.Factory("provideArguments")
+	@CartesianTest.MethodFactory("provideArguments")
 	void testNeedingSameArguments(String string, int i) {
 		// different passing test code
 	}
 
-	static CartesianProductTest.Sets provideArguments() {
-		return new CartesianProductTest.Sets()
-			.add("Mercury", "Earth", "Venus")
-			.add(1, 12, 144);
+	static ArgumentSets provideArguments() {
+		return ArgumentSets
+			.argumentsForFirstParameter("Mercury", "Earth", "Venus")
+			.argumentsForNextParameter(1, 12, 144);
 	}
 }
 ----
@@ -392,11 +394,11 @@ There are multiple requirements the static factory method has to fulfill to qual
 - must have the same name as the test method (or its name must be specified via the `factory` attribute)
 - must be `static`
 - must have **no** parameters
-- must return `CartesianProductTest.Sets`
+- must return `ArgumentSets`
 - must register values for every parameter exactly once
 - must register values in order
 
-==== Returning wrong `Sets` in the static factory method
+==== Returning wrong `ArgumentSets` in the static factory method
 
 If you register too few, too many, or conflicting parameters, you will get an https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/extension/ParameterResolutionException.html[`ParameterResolutionException`].
 "Conflicting parameters" means your test method has a parameter that should be injected by JUnit (e.g.: `TestReporter`) but you also try to inject it.
@@ -407,39 +409,40 @@ Examples of badly configured tests/static factory method:
 ----
 class BadExamples {
 	@CartesianTest
-	@CartesianTest.Factory("resolveParameters")
+	@CartesianTest.MethodFactory("resolveParameters")
 	void tooFewParameters(String string, int i, boolean b) {
 		// fails because the boolean parameter is not resolved
 	}
 
 	@CartesianTest
-	@CartesianTest.Factory("resolveParameters")
+	@CartesianTest.MethodFactory("resolveParameters")
 	void tooManyParameters(String string) {
 		// fails because we try to supply a non-existent integer parameter
 	}
 
 	@CartesianTest
-	@CartesianTest.Factory("resolveParameters")
+	@CartesianTest.MethodFactory("resolveParameters")
 	void wrongOrderParameters(int i, String string) {
 		// fails because the static factory method declared parameter sets in the wrong order
 	}
 
 	@CartesianTest
-	@CartesianTest.Factory("resolveTestReporterParam")
+	@CartesianTest.MethodFactory("resolveTestReporterParam")
 	void conflictingParameters(String string, TestReporter info) {
 		// fails because both the factory method and JUnit tries to inject TestReporter
 	}
 
-	static Sets resolveParameters() {
-		return new Sets()
-			.add("A", "B", "C")
-			.add(1, 2, 3);
+	static ArgumentSets resolveParameters() {
+		return ArgumentSets
+			.argumentsForFirstParameter("A", "B", "C")
+			.argumentsForNextParameter(1, 2, 3);
 	}
 
-	static Sets resolveTestReporterParam() {
-		return new Sets()
-			.add("A", "B", "C")
-			.add(new MyTestReporter()); // in this case MyTestReporter implements TestReporter
+	static ArgumentSets resolveTestReporterParam() {
+		return ArgumentSets
+			.argumentsForFirstParameter("A", "B", "C")
+			// in this case MyTestReporter implements TestReporter
+			.argumentsForNextParameter(new MyTestReporter());
 	}
 }
 ----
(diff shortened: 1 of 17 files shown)


Bukama:
At our team meeting we talked about some examples when to use which annotation and if they are combinable.

@nipafx and @beatngu13 also wanted to have a deeper look at the code, if I remember correctly.

nipafx:
Can you describe the idea in a few words? :)

src/main/java/org/junitpioneer/jupiter/cartesian/CartesianEnumArgumentsProvider.java
```diff
@@ -30,7 +30,7 @@
  * @implNote This class does not implement {@code ArgumentsProvider} since the Jupiter's {@code EnumSource}
```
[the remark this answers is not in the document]
nipafx: Oh, I proposed to make `CartesianArgumentsProvider` package-visible. Then the comment could go into `package-info.java`. Having read through `CartesianMethodArgumentsProvider` and `CartesianParameterArgumentsProvider`, I don't think we should stress the connection to `ArgumentsProvider` too much or even encourage people to implement both.

src/main/java/org/junitpioneer/jupiter/cartesian/CartesianTestExtension.java
```diff
@@ -76,6 +77,22 @@ private static void ensureNoInputConflicts(List<?> methodSources, List<?> parame
...
+				"Only one method-level arguments source can be used with @CartesianTest");
+	}
+
+	private List<List<?>> getSetsFromMethodArgumentsSource(Annotation argumentsSource, ExtensionContext context) {
+		try {
+			CartesianMethodArgumentsProvider provider = initializeMethodArgumentsProvider(argumentsSource,
+				context.getRequiredTestMethod());
+			// @formatter:off
+			return provider.provideArguments(context).get();
+			// @formatter:on
```
nipafx, line 91: I think the formatter is fine with the code as is. 😉

src/main/java/org/junitpioneer/jupiter/cartesian/CartesianTestExtension.java
```diff
@@ -90,39 +107,58 @@ private static void ensureNoInputConflicts(List<?> methodSources, List<?> parame
...
 			throw new ExtensionConfigurationException("Could not provide arguments because of exception.", ex);
 		}
 	}
 
-	private CartesianArgumentsProvider<?> initializeArgumentsProvider(Annotation source, Parameter parameter) {
+	private CartesianMethodArgumentsProvider initializeMethodArgumentsProvider(Annotation source, Method method) {
+		CartesianArgumentsSource providerAnnotation = AnnotationSupport
+				.findAnnotation(method, CartesianArgumentsSource.class)
+				// never happens, we already know these annotations are annotated with @ArgumentsSource
+				.orElseThrow(() -> new PreconditionViolationException(format(
```
nipafx, line 122: Shouldn't this then be an `IllegalStateException`? I understand `PreconditionViolationException` to mean that the user did something wrong.

src/main/java/org/junitpioneer/jupiter/cartesian/CartesianTestExtension.java
```diff
@@ -90,39 +107,58 @@ private static void ensureNoInputConflicts(List<?> methodSources, List<?> parame
...
 				.findAnnotation(parameter, CartesianArgumentsSource.class);
-
 		if (cartesianProviderAnnotation.isPresent()) {
-			return AnnotationConsumerInitializer
-					.initialize(parameter, ReflectionSupport.newInstance(cartesianProviderAnnotation.get().value()));
+			providerClass = cartesianProviderAnnotation.get().value();
+		} else {
+			ArgumentsSource providerAnnotation = AnnotationSupport
+					.findAnnotation(parameter, ArgumentsSource.class)
+					.orElseThrow(() -> new PreconditionViolationException(
```
nipafx, line 142: As above.

src/main/java/org/junitpioneer/jupiter/cartesian/CartesianTest.java
```diff
@@ -310,4 +310,13 @@ void validate(CartesianTest.Enum enumSource, Set<? extends java.lang.Enum<?>> co
 
 	}
 
+	@Retention(RetentionPolicy.RUNTIME)
+	@Target({ ElementType.METHOD, ElementType.ANNOTATION_TYPE })
+	@CartesianArgumentsSource(CartesianFactoryArgumentsProvider.class)
+	@interface Factory {
```
nipafx, line 316: Should we call this `MethodSource` in line with Jupiter or is there a good reason to have a different name (e.g. to avoid collisions)?

src/main/java/org/junitpioneer/jupiter/cartesian/CartesianMethodArgumentsProvider.java
```diff
@@ -0,0 +1,66 @@
...
+	 * to be used for the {@code @CartesianTest}.
+	 * For more information, see
+	 * <a href="https://junit-pioneer.org/docs/cartesian-product/" target="_top">the Cartesian product documentation</a>.
+	 *
+	 * @param context the current extension context; never {@code null}
+	 * @return a {@code Sets} object; never {@code null}
+	 */
+	Sets provideArguments(ExtensionContext context) throws Exception;
+
+	class Sets {
```
nipafx, line 43: This class and its methods need Javadoc.

docs/cartesian-product.adoc
```diff
@@ -300,37 +300,190 @@ To demonstrate with a table:
...
+You can annotate your test method to supply arguments to all parameters simultaneously.
+
+=== `@CartesianTest.Factory`
+
+`@CartesianTest.Factory` can be used to name a static factory method that supplies your arguments.
+The `name` annotation parameter is mandatory.
+Just like with JUnit's `@MethodSource`, you can specify the factory method with its fully-qualified name (including the class), e.g. `com.example.Class#factory`.
+This method must return `CartesianMethodArgumentsProvider.Sets`.
+`Sets` is a helper class, specifically for creating sets for `@CartesianTest`.
+To create the test data, instantiate a `new Sets()` then use the `add()` method to register the values for the parameters.
```
nipafx, line 314: It should be made clearer that `add` needs to be called once per parameter with all arguments for that parameter.

docs/cartesian-product.adoc
```diff
@@ -300,37 +300,190 @@ To demonstrate with a table:
...
 For more information, please see the link:docs/range-sources[separate documentation about range sources].
 
+== Annotating your test method
+
+You can annotate your test method to supply arguments to all parameters simultaneously.
+
+=== `@CartesianTest.Factory`
+
+`@CartesianTest.Factory` can be used to name a static factory method that supplies your arguments.
+The `name` annotation parameter is mandatory.
```
nipafx, line 310: Is this line still up to date? I can't see `name` on `Factory`.

docs/cartesian-product.adoc
```diff
@@ -300,37 +300,190 @@ To demonstrate with a table:
...
+
+	static CartesianProductTest.Sets provideArguments() {
+		return new CartesianProductTest.Sets()
+			.add("Mercury", "Earth", "Venus")
+			.add(1, 12, 144);
+	}
+}
+----
+
+==== Conditions for the static factory method
```
nipafx, line 387: Personally, I'd prefer "requirements", but I'll leave that to you.

docs/cartesian-product.adoc
```diff
@@ -300,37 +300,190 @@ To demonstrate with a table:
...
+
+There are multiple conditions the static factory method has to fulfill to qualify:
+
+- must have the same name as the test method (or its name must be specified via the `factory` attribute)
+- must be `static`
+- must have **no** parameters
```
nipafx, line 393: What if we could inject parameters into factory methods, though?! 🤡

docs/cartesian-product.adoc
```diff
@@ -300,37 +300,190 @@ To demonstrate with a table:
...
-Because `CartesianTest` works with parameters, you will have to use `CartesianArgumentsProvider` instead of `ArgumentsProvider`.
-This will let you initialize your argument source with it's corresponding parameter.
+You might find that the available annotations do not fit your need.
+For a `@ParameterizedTest`, you can write a custom https://junit.org/junit5/docs/current/api/org.junit.jupiter.params/org/junit/jupiter/params/provider/ArgumentsSource.html::[`@ArgumentsSource`] and a corresponding https://junit.org/junit5/docs/current/api/org.junit.jupiter.params/org/junit/jupiter/params/provider/ArgumentsProvider.html::[`ArgumentsProvider`].
+You can do the same thing for `@CartesianTest`, with the following caveats:
+
+ - Instead of `@ArgumentsSource`, you have to use https://javadoc.io/static/org.junit-pioneer/junit-pioneer/1.5.0/org/junitpioneer/jupiter/cartesian/CartesianArgumentsSource.html::[`@CartesianArgumentsSource`]
+ - Instead of `ArgumentsProvider`, you have to use one of two interfaces, each explored in their own section.
```
nipafx, line 453: There's a later comment in this review where I give an opinion on when and how to mention `ArgumentsProvider`. As to how that applies here: I think this explanation here is good and should stay as is, but all other mentions of `ArgumentsProvider` in this documentation should be removed.

src/main/java/org/junitpioneer/jupiter/cartesian/CartesianTest.java
Michael1993: No, we should not. For several reasons:
 - In `@CartesianProductTest` this was a property called `factory`.
 - It avoids collision, but more importantly it avoids confusion. This annotation is only for `@CartesianTest`.
 - We decided to use short names for simplicity. Even though this annotation doesn't go on a parameter, I'd still keep it short.
 - I don't want to change it. 😭

src/main/java/org/junitpioneer/jupiter/cartesian/CartesianEnumArgumentsProvider.java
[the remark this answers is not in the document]
Michael1993: Tbh. there is actually no real connection to `ArgumentsProvider`...

docs/cartesian-product.adoc
Michael1993: Ha-ha, very funny.

docs/cartesian-product.adoc
Michael1993: I think this is the _only_ explanation that's needed.

docs/cartesian-product.adoc
Michael1993: Updated it a bit.

nipafx: Me too. 😁

src/main/java/org/junitpioneer/jupiter/cartesian/CartesianEnumArgumentsProvider.java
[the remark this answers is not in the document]
nipafx: Good thing it's gone now. :)

src/main/java/org/junitpioneer/jupiter/cartesian/CartesianEnumArgumentsProvider.java
```diff
@@ -30,7 +30,7 @@
...
  * should be used for that.
  */
-class CartesianEnumArgumentsProvider<E extends Enum<E>> implements CartesianArgumentsProvider<E> {
+class CartesianEnumArgumentsProvider<E extends Enum<E>> implements CartesianParameterArgumentsProvider<E> {
 
 	@Override
 	public Stream<E> provideArguments(ExtensionContext context, Parameter parameter) {
```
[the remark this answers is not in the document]
nipafx: > the exception message is from the future

Lol.

Since I'm already editing, I'll replace the message.

src/main/java/org/junitpioneer/jupiter/cartesian/CartesianTest.java
nipafx: I have to think about it. If I feel we need to change it, I can do the work. 😉

nipafx:
Looks great! Things left to do:

* make sure correct exceptions are thrown (I see a strong preference of `PreconditionViolationException`)
* further eradicate mentions of `ArgumentsProvider`
* consider renaming/refactoring `Sets` for better ease of use

I can tackle those on stream in an hour.

docs/cartesian-product.adoc
```diff
@@ -300,37 +300,191 @@ To demonstrate with a table:
...
+	@CartesianTest.Factory("provideArguments")
+	void testNeedingSameArguments(String string, int i) {
+		// different passing test code
+	}
+
+	static CartesianProductTest.Sets provideArguments() {
+		return new CartesianProductTest.Sets()
+			.add("Mercury", "Earth", "Venus")
+			.add(1, 12, 144);
+	}
```
filiphr, line 384: I don't understand why you don't piggy back on what is already available there and use `MethodSource`. I can write this like:

```
static Stream<Arguments> provideArguments() {
    return Stream.of(
        Arguments.of("Mercury", "Earth", "Venus"),
        Arguments.of(1, 12, 144),
    )
}
```

You will benefit of a lot of simplifications:

1. No need to do magic with reflection and you piggy back on the logic done by JUnit Jupiter (not sure if you can get an instance of the provider though)
2. You get nice auto completion and methods marked as used in IntelliJ

nipafx: First and foremost, because `Arguments` is not the right abstraction. [Its Javadoc](https://junit.org/junit5/docs/current/api/org.junit.jupiter.params/org/junit/jupiter/params/provider/Arguments.html) says:

> `Arguments` is an abstraction that provides access to an array of objects to be used for invoking a `@ParameterizedTest` method. 

What we need, though, is one "array of objects" per parameter (not per execution) and we're not invoking anything with them (they get folded into other "arrays", which then dictate the number of executions), let alone a `@ParameterizedTest`.

Re 1.: Unless we're reflectively accessing `MethodArgumentsProvider`, we still have to write our own reflective code to get these arguments. Even if we copy its source code, we'd still have to maintain it ourselves.

Re 2.: True, but I don't think that's worth the trade off.

nipafx:
All that's needed is a commit message, right?

Michael1993:
> All that's needed is a commit message, right?

Yes. Well, also approvals from maintainers. 😄

nipafx (approved):

PR #546, opened 2021-11-22
Add NotWorking test extension (#551 / #546)

Marcono1234:
Preliminary remark: I wanted to implement this either way, therefore I directly created a pull request. However, if you prefer I can also create a GitHub issue where this can be discussed first.

Adds an extension with annotation `@NotWorking` for marking test methods which are 'not working'; quoting from the documentation added by this pull request:
> Often tests fail due to a bug in the tested application or in used dependencies. Traditionally such a test method would be annotated with JUnit's `@Disabled`. However, this has the following disadvantages when the bug causing the test failure is fixed:
> * the developer might not notice the existing test method and create a new one
> * the existing test method might not be noticed and remains disabled for a long time after the bug has been fixed, adding no value for the project
> 
> `@NotWorking` tries to solve these issues. Unlike `@Disabled` it still executes the annotated test method but aborts the test if a test failure or error occurs. However, if the test is executed successfully the `@NotWorking` annotation will cause a test failure because the test _is working_. This lets the developer know that they have fixed a bug (possibly by accident) and that they can now remove the `@NotWorking` annotation from the test method.

Originally I proposed this in https://github.com/junit-team/junit5/issues/2023#issuecomment-846474693, but I am not sure if by accident I 'hijacked' that issue, and whether I described my intentions there clear enough.
Therefore I first wanted to propose this here because I hope it has higher chances being accepted here.

I am not very familiar with JUnit extension development yet, therefore I am not sure if the way I implemented this is a clean way or if it can be improved. Any feedback is appreciated!

Proposed commit message:

```
Add NotWorking test extension (#551 / #546)

PR: #546
```

---
**PR checklist**

Documentation (general)

Documentation (new extension)

Code

Contributing

---


Michael1993:
Could you tell me what the purpose of the extension is? When would I use this over `@Disabled`?

Marcono1234:
> What is the difference between a failing test running but not breaking and a test not running at all?

Let's imagine there is a bug in your application:
1. You mark the corresponding test method with `@Disabled`
2. At some later point you fix the bug (possibly by accident, e.g. due to refactoring)
3. The test is still disabled, because you are not aware of its existence or you forgot that you disabled it
4. In the worst case you write a new test which covers the exact same scenario covered by the disabled test; this adds maintenance burden

With the proposed `@NotWorking` extension you would immediately after the bug was fixed (point 2 in the list above) notice the existing test. This avoids
- duplicating tests
- having disabled tests lingering around for a long time after the issue why they were disabled has been solved

src/test/java/org/junitpioneer/jupiter/NotWorkingExtensionTests.java
```diff
@@ -0,0 +1,225 @@
...
+
+		@NotWorking
+		@Test
+		void test() {
+		}
+
+	}
+
+	/**
+	 * {@link BeforeEach} should be considered by {@link NotWorking} because it
```
Michael1993, line 174: Did you mean to write `@AfterEach`?

src/main/java/org/junitpioneer/jupiter/NotWorkingExtension.java
```diff
@@ -0,0 +1,101 @@
...
+	@Override
+	public void afterEach(ExtensionContext context) throws Exception {
+		NotWorking annotation = getNotWorkingAnnotation(context);
+
+		// null if extension should ignore test result
+		Boolean didExceptionOccur = getStore(context).remove(EXCEPTION_OCCURRED_STORE_KEY, Boolean.class);
+		if (Boolean.TRUE.equals(didExceptionOccur)) {
+			String message = annotation.value();
+			if (message.isEmpty()) {
+				message = "Test marked as 'not working' failed";
```
Michael1993, line 69:
```suggestion
				message = "Test marked as 'not working' failed as expected";
```
I think this is clearer, what do you think?

src/main/java/org/junitpioneer/jupiter/NotWorkingExtension.java
```diff
@@ -0,0 +1,101 @@
...
+	@Override
+	public void beforeEach(ExtensionContext context) throws Exception {
+		getStore(context).put(EXCEPTION_OCCURRED_STORE_KEY, false);
+	}
+
+	@Override
+	public void afterEach(ExtensionContext context) throws Exception {
+		NotWorking annotation = getNotWorkingAnnotation(context);
+
+		// null if extension should ignore test result
```
Michael1993, line 64: Don't really see the thought process behind this statement. I thought this value can never be `null` because for the extension to get invoked during the `afterEach` phase it _must_ have been invoked during the `beforeEach` phase.

Michael1993: Oh, right - this is because of assumptions? In which case it makes me wonder why aren't you storing the exception object itself instead of a `boolean`?

src/main/java/org/junitpioneer/jupiter/NotWorkingExtension.java
```diff
@@ -0,0 +1,101 @@
...
+class NotWorkingExtension implements Extension, TestExecutionExceptionHandler, LifecycleMethodExecutionExceptionHandler,
+		BeforeEachCallback, AfterEachCallback {
+
+	private static final Namespace NAMESPACE = Namespace.create(NotWorkingExtension.class);
+	private static final String EXCEPTION_OCCURRED_STORE_KEY = "exceptionOccurred";
+
+	public NotWorkingExtension() {
+	}
+
+	private Store getStore(ExtensionContext context) {
```
Michael1993, line 34: To be honest, I'd inline this method, but that's just my personal preference.

Bukama:
We talked about this at our last team meeting and want you to invite to take part in a discussion to think about this and similar use cases in a more global view.

Feel free to write down your thoughts and ideas in #550 - thank you!

Bukama (commented):
Thank you for the PR.
I really like the documentation text about when to use it and especially when not. Maybe a link to `assertThrows` can be added.

@Michael1993 has already stated his thoughts about the technical implementation.

One minor formal thing: Some time ago we decided that we always want to have an issue on which PRs are based on to avoid discussing non-implementation related things in the issue instread in the PR.

edit: I've created #551 for this.

src/main/java/org/junitpioneer/jupiter/NotWorkingExtension.java
Marcono1234: If you don't mind I would like to keep the method for now to avoid having to repeat the `getStore(NAMESPACE)` part, which is a bit more verbose.

src/main/java/org/junitpioneer/jupiter/NotWorkingExtension.java
Marcono1234: Currently both `handleTestExecutionException` and `handleAfterEachMethodExecutionException` check the exception type and for test abort exceptions directly throw them.

If the exception object was put in the store, the abort exception check would have to be repeated in the `afterEach` method. I am not sure if removing the check from `handleTestExecutionException` and `handleAfterEachMethodExecutionException` would be a good idea, maybe JUnit behaves differently depending on from where an exception is thrown (i.e. `handleTestExecutionException`, `handleAfterEachMethodExecutionException` or `afterEach`).

I have added some documentation to the `EXCEPTION_OCCURRED_STORE_KEY` field, does that help? Or would you still prefer to store the exception object?

docs/not-working-tests.adoc
```diff
@@ -22,8 +22,9 @@ in most cases; exceptions for this are for example 'flaky' tests.
 
 *Important*: This annotation is *not* intended as a way to mark test methods
 which intentionally cause exceptions.
-Such test methods should use JUnit's `assertThrows` or similar means to explicitly
-test for a specific exception class being thrown by a specific action.
+Such test methods should use https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/Assertions.html#assertThrows(java.lang.Class,org.junit.jupiter.api.function.Executable)[JUnit's `assertThrows`]
```
Marcono1234, line 25: I am not completely sure how well AsciiDoc handles this URL with parentheses, but at least in the preview in Eclipse it seems to work.

@Bukama, is that how you wanted the link to the JUnit docs?

Marcono1234:
Thanks a lot to all of you for being so welcoming and providing such helpful review comments (here and on my other pull requests)! I have tried to address most of the feedback, but left a comment for the refactoring of the value stored in the JUnit Store.

I also don't mind if you want to keep this pull request open until the discussion in #551 has come to a conclusion.

Happy Christmas Holidays! (and no hurry to respond to my comment)

src/main/java/org/junitpioneer/jupiter/NotWorking.java
```diff
@@ -0,0 +1,70 @@
...
+ * Implementation note:
+ * Only supports METHOD as target but not test classes because there it is not clear
+ * what the 'correct' behavior would be when only a few test methods execute successfully.
+ * Would the developer then have to remove the @NotWorking annotation from the test
+ * class and annotate methods individually?
+ * Additionally it is rather unlikely that a complete test class is 'not working'.
+ */
+@Documented
+@Retention(RUNTIME)
+@Target(METHOD)
```
nipafx, line 59: It should also be possible to annotate other annotations, so users can create their own aggregate annotations. For that you need to change how you find the annotations, more on that below.

src/main/java/org/junitpioneer/jupiter/NotWorkingExtension.java
```diff
@@ -0,0 +1,113 @@
...
+		return context.getStore(NAMESPACE);
+	}
+
+	private static NotWorking getNotWorkingAnnotation(ExtensionContext context) {
+		return context.getRequiredTestMethod().getDeclaredAnnotation(NotWorking.class);
```
nipafx, line 51: If you use Jupiter's `AnnotationSupport::findAnnotation`, it will also look for meta annotations. It returns an `Optional`, which we usually get out of with `.orElseThrow(() -> new IllegalStateException("@NotWorking is missing."))`.

src/main/java/org/junitpioneer/jupiter/NotWorkingExtension.java
```diff
@@ -0,0 +1,113 @@
...
+import org.junit.jupiter.api.extension.Extension;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.extension.ExtensionContext.Namespace;
+import org.junit.jupiter.api.extension.ExtensionContext.Store;
+import org.junit.jupiter.api.extension.LifecycleMethodExecutionExceptionHandler;
+import org.junit.jupiter.api.extension.TestExecutionExceptionHandler;
+import org.opentest4j.TestAbortedException;
+
+class NotWorkingExtension implements Extension, TestExecutionExceptionHandler, LifecycleMethodExecutionExceptionHandler,
+		BeforeEachCallback, AfterEachCallback {
```
nipafx, line 26: I was surprised by how many extension points are involved here and checked [the existing extension points](https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/extension/Extension.html), looking for a simpler approach. I think I found it with `InvocationInterceptor`, where all you need to do is proceed with the invocation and then invert the result, i.e. fail on a successful execution and abort on failure. Here's a proof of concept:

```java
import org.junit.jupiter.api.Test;
import org.junit.jupiter.api.extension.ExtendWith;
import org.junit.jupiter.api.extension.ExtensionContext;
import org.junit.jupiter.api.extension.InvocationInterceptor;
import org.junit.jupiter.api.extension.ReflectiveInvocationContext;
import org.opentest4j.AssertionFailedError;
import org.opentest4j.TestAbortedException;

import java.lang.reflect.Method;

import static org.junit.jupiter.api.Assertions.fail;

public class LabTests {

	@Test
	@ExtendWith(InvertResultExtension.class)
	void test() {
		System.out.println("Hi!");
		fail();
	}

	private static class InvertResultExtension implements InvocationInterceptor {

		@Override
		public void interceptTestMethod(Invocation<Void> invocation, ReflectiveInvocationContext<Method> invocationContext, ExtensionContext extensionContext) throws Throwable {
			invert(invocation);
		}

		@Override
		public <T> T interceptTestFactoryMethod(Invocation<T> invocation, ReflectiveInvocationContext<Method> invocationContext, ExtensionContext extensionContext) throws Throwable {
			return invert(invocation);
		}

		private <T> T invert(Invocation<T> invocation) {
			try {
				invocation.proceed();
			} catch (Throwable ex) {
				throw new TestAbortedException("Test failed as expected.", ex);
			}
			throw new AssertionFailedError("Test unexpectedly passed.");
		}

	}

}
```

src/test/java/org/junitpioneer/jupiter/NotWorkingExtensionTests.java
```diff
@@ -0,0 +1,225 @@
...
+	void doesNotAbortOnAfterAllTestFailure() {
+		ExecutionResults results = PioneerTestKit.executeTestMethod(DummyTestClassFailureAfterAll.class, "test");
+		assertThat(results)
+				.hasSingleStartedTest()
+				.whichFailed()
+				.withExceptionInstanceOf(AssertionError.class)
+				.hasMessage("Test marked as 'not working' succeeded; remove @NotWorking from it");
+	}
+
+	static class DummyTestClass {
```
nipafx, line 120: Can you rename this to `NotWorkingTestCases` or something else that ends with `TestCases`. 😉

src/test/java/org/junitpioneer/testkit/assertion/TestCaseAssertBase.java
```diff
@@ -47,7 +48,7 @@
 	@Override
 	public TestCaseFailureAssert whichFailed() {
 		assertThat(actual.failed().count()).isEqualTo(1);
-		return this;
+		return new TestCaseAssertBase(actual.failed());
```
nipafx, line 51: Took me a while to understand this, but I think it makes sense and actually makes the code a bit clearer. Thanks!

nipafx (commented):
Thank you @Marcono1234, good work! It's great to see PRs that cover all the bases (from code to tests to documentation, etc.). There are a bunch of comments below. I'm also not too happy with the name - we can discuss that on the issue.

src/main/java/org/junitpioneer/jupiter/NotWorkingExtension.java
```diff
@@ -0,0 +1,113 @@
...
+	 *   the test to be skipped
+	 *   <li>{@code false}: No exception occurred; the extension should cause the test to fail
+	 *   <li>{@code null}: A regular test abort exception was thrown by user code; the
+	 *   extension should do nothing
+	 * </ul>
+	 */
+	private static final String EXCEPTION_OCCURRED_STORE_KEY = "exceptionOccurred";
+
+	public NotWorkingExtension() {
+	}
```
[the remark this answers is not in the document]
Marcono1234: I mainly added it to make it explicit that a no-args constructor is required for JUnit to create an instance. Often having an implicit default no-args constructor is considered bad practice.
Should I remove it nonetheless?

docs/not-working-tests.adoc
```diff
@@ -0,0 +1,56 @@
...
+annotated test method but aborts the test if a test failure or error occurs.
+However, if the test is executed successfully the `@NotWorking` annotation will cause
+a test failure because the test _is working_.
+This lets the developer know that they have fixed a bug (possibly by accident) and that
+they can now remove the `@NotWorking` annotation from the test method.
+
+It is therefore recommended to prefer using `@NotWorking` over `@Disabled`
+in most cases; exceptions for this are for example 'flaky' tests.
+
+*Important*: This annotation is *not* intended as a way to mark test methods
```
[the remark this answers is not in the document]
Marcono1234: Thanks for the hint! I was not aware of this.

src/main/java/org/junitpioneer/jupiter/NotWorkingExtension.java
Marcono1234: That seems to work for nearly all cases except for exceptions thrown in an `@AfterEach` method, which would then probably require similarly complex logic as it is currently implemented.

Should I switch to `InvocationInterceptor` nonetheless?

src/main/java/org/junitpioneer/jupiter/NotWorking.java
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * <p>For more details and examples, see
+ * <a href="https://junit-pioneer.org/docs/not-working-tests/" target="_top">the documentation on <code>@NotWorking</code></a>.
+ * </p>
+ *
+ * @see org.junit.jupiter.api.Disabled
+ */
+/*
+ * Implementation note:
+ * Only supports METHOD as target but not test classes because there it is not clear
```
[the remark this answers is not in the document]
Marcono1234: Also mentioned that it can be used as meta-annotation, and that it has to be used in combination with a "testable" annotation (e.g. `@Test`).

Marcono1234:
Thanks for the feedback! I tried to address most of it and commented in the other cases, hopefully that is what you had in mind.

src/test/java/org/junitpioneer/testkit/assertion/TestCaseAssertBase.java
Marcono1234: Sorry for unresolving this comment. I was unsure whether this change could lead to any incorrect usage (similar to how `CONTRIBUTING.md` shows `hasNumberOfFailedTests(3).hasSingleSucceededTest()`), but this does not seem to be an issue here. The returned `TestCaseFailureAssert` only allows performing assertions on the exception. Therefore I resolved this comment again.

Michael1993:
@Marcono1234 I don't think you need to worry about `@BeforeEach` or `@AfterEach`, since exceptions occurring there are not really part of the tested behaviour. In fact, they explicitly _shouldn't_ be testing behaviour and so annotating them with `@NotWorking` (or `@HasToBeFixed`) is a mistake.

What do you think?

Marcono1234:
> I don't think you need to worry about `@BeforeEach` or `@AfterEach`, since exceptions occurring there are not really part of the tested behaviour.

I was thinking of the case where the cleanup in `@AfterEach` fails due to the bug which also causes the test to fail. But I did not intend the annotation to be used on the `@AfterEach` method itself.
Do you think it is worth keeping this functionality, or should it be dropped from this pull request?

The extension implements `BeforeEachCallback` to set up the context store, but it intentionally does not support detecting exceptions from a `@BeforeEach` method.

Michael1993:
Hey @Marcono1234 

Regarding this PR: we would like you to switch to `InvocationInterceptor` and fix the merge conflicts - if you are still available to work on this.

Marcono1234:
As mentioned in my comment above `InvocationInterceptor` might not work for the `@AfterEach` case. In case that is not possible, do you want me to use `InvocationInterceptor` nonetheless even though this means the extension will have reduced functionality?

However, there is a `InvocationInterceptor.interceptAfterEachMethod` method. Not sure if I overlooked it back then, or if even with that method the extension would have to remember whether an exception occurred to support the `@AfterEach` case.

This would also include renaming the extension to `@ExpectedToFail` (https://github.com/junit-pioneer/junit-pioneer/issues/551#issuecomment-1132048284), right?

Michael1993:
> As mentioned in my comment above `InvocationInterceptor` might not work for the `@AfterEach` case. In case that is not possible, do you want me to use `InvocationInterceptor` nonetheless even though this means the extension will have reduced functionality?

Yes, please use `InvocationInterceptor`, it's much more fitting for this use-case.

> However, there is a `InvocationInterceptor.interceptAfterEachMethod` method. Not sure if I overlooked it back then, or if even with that method the extension would have to remember whether an exception occurred to support the `@AfterEach` case.

Feel free to investigate this and we will discuss it with the other maintainers when it's part of the PR.

> This would also include renaming the extension to `@ExpectedToFail` ([#551 (comment)](https://github.com/junit-pioneer/junit-pioneer/issues/551#issuecomment-1132048284)), right?

Yes, that'd be super. Thank you!

Marcono1234:
Because this renaming and usage of `InvocationInterceptor` is quite different from what I originally implemented here, I am closing this and have opened a separate PR for this: #676

That PR is based on the commits of this one which hopefully makes reviewing the changes easier. I hope that is alright.

PR #565, opened 2021-12-25
Fix most compiler warnings

Marcono1234:
Proposed commit message:

```
Fix most compiler warnings (#565)

Fixes most compiler warnings about unchecked casts and raw types.

PR:  #565
```

In the main sources there is still one warning left in `RangeSourceArgumentsProvider` for the import of the deprecated `CartesianAnnotationConsumer`. However, apparently that warning for the import itself cannot be suppressed.

In the test sources there are warnings left for usage of `TestPlan.add(TestIdentifier)` in `IssueExtensionExecutionListenerTests`. I have not suppressed them because eventually it might be necessary to find a replacement for that. Currently these calls are only made against a mock instance, so it is most likely not an issue that the method implementation was changed recently to always throw an exception (https://github.com/junit-team/junit5/commit/8b910ecefff4dff177867db812c2fb28bb274f34).

---
**PR checklist**

Documentation (general)

Documentation (new extension)

Code

Contributing

---


src/test/java/org/junitpioneer/jupiter/CartesianProductTestExtensionTests.java
```diff
@@ -29,7 +29,6 @@
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ValueSource;
 import org.junit.platform.commons.PreconditionViolationException;
-import org.junitpioneer.jupiter.CartesianEnumSource.Mode;
```
Marcono1234, line 32: Have removed this import to avoid a (non-suppressable) deprecation warning for it. However, that has the disadvantage that `Mode` usage in the tests below is a bit verbose.

Should I revert this?

src/test/java/org/junitpioneer/jupiter/issue/IssueExtensionExecutionListenerTests.java
```diff
@@ -92,8 +92,6 @@ void abortedIssueTestCaseCreated() {
 		assertAll(() -> assertThat(issueTestSuite.issueId()).isEqualTo("#123"),
 			() -> assertThat(issueTestSuite.tests().size()).isEqualTo(1));
 
-		IssueTestCase testCase = issueTestSuite.tests().get(0);
```
Marcono1234, line 95: Variable was unused and assertion below seems to cover that already.

src/test/java/org/junitpioneer/testkit/assertion/PioneerAssert.java
```diff
@@ -266,9 +266,9 @@ private TestSuiteContainersFailureAssert hasNumberOfSpecificContainers(long cont
 
 	@SafeVarargs
 	@Override
-	public final ListAssert<String> withExceptionInstancesOf(Class<? extends Throwable>... exceptionTypes) {
```
Marcono1234, line 269: See comment in `TestSuiteFailureAssert.java` below.

src/test/java/org/junitpioneer/testkit/assertion/suite/TestSuiteFailureAssert.java
```diff
@@ -24,7 +24,12 @@
 	 * @param exceptionTypes the exception types you want to check
 	 * @return a {@link ListAssert} for asserting exception messages
 	 */
-	ListAssert<String> withExceptionInstancesOf(Class<? extends Throwable>... exceptionTypes);
+	/*
+	 * Note: Parameter type should be `Class<? extends Throwable>...`,
+	 * but @SafeVarargs is currently not permitted on abstract methods,
+	 * which causes annoying warnings on every usage of this method
+	 */
+	ListAssert<String> withExceptionInstancesOf(Class<?>... exceptionTypes);
```
Marcono1234, line 32: Please let me know if this should be reverted. The alternative would be to use `@SuppressWarnings("unchecked")` for every caller of this method.

Michael1993:
I don't know what to think about this, my gut reaction is that it's not really necessary.

Marcono1234:
> my gut reaction is that it's not really necessary

Why not? The current amount of warnings makes it more difficult to find newly introduced issues, and some of the warnings (at least the unused variable one in `IssueExtensionExecutionListenerTests`) are legit and should be fixed.
For example `CartesianProductTestExtensionTests` has dozens of warnings due to usage of the deprecated class. This makes it rather distracting (at least in Eclipse). Of course you could edit the IDE settings to ignore certain warning types, but ideally these false positive warnings should be suppressed so that `javac` is not creating warnings for it either.

Michael1993:
Personally, I am fine with these changes but I'd like someone with more experience to okay them.

src/test/java/org/junitpioneer/testkit/assertion/suite/TestSuiteFailureAssert.java
nipafx: I understand the impetus, but being able to pass arbitrary classes is weird. I think there's a better solution: All callers of this method only pass one argument, so let's refactor to:

```java
	/*
	 * Note: We avoid a varargs-variant of this method to prevent heap-pollution warnings.
	 * If you need a method with n exception types, create n-1 `withExceptionInstancesOf`
	 * (note the plural) overloads with [2, n] parameters.
	 */
	ListAssert<String> withExceptionInstanceOf(Class<? extends Throwable> exceptionType);
```

(Note the method name's singular.)

nipafx (changes requested):
Thanks @Marcono1234, I really like the PR. Fewer warnings are always a good thing in my book and I'd like to take this even further and see whether we can get it to zero warnings, so we can turn them into errors with the compiler flag `-Werror`. More on that in another comment.

nipafx:
> In the main sources there is still one warning left in `RangeSourceArgumentsProvider` for the import of the deprecated `CartesianAnnotationConsumer`. However, apparently that warning for the import itself cannot be suppressed.

It can't. In Java 9+ it ceases to be a warning, but as long as we're on 8, we'll have to stick with it. What works, paradoxically, is to replace each use of the class with its fully qualified name (thus allowing us to remove the import) and then suppress the resulting warnings in the methods.

> In the test sources there are warnings left for usage of `TestPlan.add(TestIdentifier)` in `IssueExtensionExecutionListenerTests`.

I didn't look into that (and don't have time to do it right now), but can we switch to an alternative API? If that requires a newer Jupiter version, that's no problem - there's a PR floating around somewhere that ups it to the most recent release (5.8.2, at the time of writing). If that helps, we can update in this PR and switch to the new API to get rid of the deprecation warning.

src/main/java/org/junitpioneer/internal/PioneerAnnotationUtils.java
```diff
@@ -60,6 +60,7 @@ private PioneerAnnotationUtils() {
 	 * <em>enclosing-present</em> on the test element (method or class) belonging to the
 	 * specified {@code context}.
 	 */
+	@SafeVarargs
 	public static boolean isAnyAnnotationPresent(ExtensionContext context,
 			Class<? extends Annotation>... annotationTypes) {
```
Michael1993, line 65:
```suggestion
	public static boolean isAnnotationPresent(ExtensionContext context,
			Class<? extends Annotation> annotationType) {
```

src/main/java/org/junitpioneer/internal/PioneerAnnotationUtils.java
```diff
@@ -75,6 +76,7 @@ public static boolean isAnyAnnotationPresent(ExtensionContext context,
 	 * <em>enclosing-present</em> on the test element (method or class) belonging to the specified
 	 * {@code context}.
 	 */
+	@SafeVarargs
 	public static boolean isAnyRepeatableAnnotationPresent(ExtensionContext context,
 			Class<? extends Annotation>... annotationTypes) {
```
Michael1993, line 81:
```suggestion
	public static boolean isRepeatableAnnotationPresent(ExtensionContext context,
			Class<? extends Annotation> annotationType) {
```

Michael1993 (changes requested):
Since the vararg parameter was unnecessary for the `withExceptionInstanceOf`, let's change it for these as well, since they are also always invoked with a single argument.

src/test/java/org/junitpioneer/testkit/assertion/suite/TestSuiteFailureAssert.java
Marcono1234: Have kept the plural form name <code>withExceptionInstance<ins>s</ins>Of</code> because the method can still return multiple exceptions in case they all have the same type.

src/test/java/org/junitpioneer/jupiter/CartesianProductTestExtensionTests.java
```diff
@@ -1208,32 +1217,35 @@ void allValuesWithMultipleOmittedTypes(TestEnum e1, AnotherTestEnum e2) {
 		}
 
 		@CartesianProductTest
-		@CartesianEnumSource(value = TestEnum.class, names = { "ONE", "TWO" }, mode = Mode.INCLUDE)
-		@CartesianEnumSource(value = AnotherTestEnum.class, names = { "BETA", "GAMMA" }, mode = Mode.EXCLUDE)
+		@CartesianEnumSource(value = TestEnum.class, names = { "ONE", "TWO" }, mode = CartesianEnumSource.Mode.INCLUDE)
+		@CartesianEnumSource(value = AnotherTestEnum.class, names = { "BETA",
+				"GAMMA" }, mode = CartesianEnumSource.Mode.EXCLUDE)
```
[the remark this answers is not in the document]
Marcono1234: Is there anything I can / should do about this?

Marcono1234:
> Since the vararg parameter was unnecessary for the `withExceptionInstanceOf`, let's change it for these as well, since they are also always invoked with a single argument.

This and the other one are actually fine (and don't cause any warnings) because `@SafeVarargs` can be used on them, and this seems to be a proper use case for this annotation here, because the array is not modified.

For `withExceptionInstancesOf` the issue was that `@SafeVarargs` could not be used because it is not allowed on methods which can be overridden.

Do you want this to be changed nonetheless?

src/test/java/org/junitpioneer/jupiter/CartesianProductTestNameFormatterTests.java
```diff
@@ -19,6 +19,7 @@
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtensionConfigurationException;
 
+@SuppressWarnings("deprecation") // For CartesianProductTestNameFormatter
```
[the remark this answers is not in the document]
Marcono1234: This annotation is already on the entire class; or do I misunderstand your comment?

Marcono1234:
> What works, paradoxically, is to replace each use of the class with its fully qualified name (thus allowing us to remove the import) and then suppress the resulting warnings in the methods.

Have done this. Now the main sources compile without warnings.

---

> > In the test sources there are warnings left for usage of `TestPlan.add(TestIdentifier)` in `IssueExtensionExecutionListenerTests`.
> 
> I didn't look into that (and don't have time to do it right now), but can we switch to an alternative API? If that requires a newer Jupiter version, that's no problem - there's a PR floating around somewhere that ups it to the most recent release (5.8.2, at the time of writing). If that helps, we can update in this PR and switch to the new API to get rid of the deprecation warning.

The issue is that `TestPlan` is an internal JUnit class and it changed in r5.8.0-RC1, see [commit](https://github.com/junit-team/junit5/commit/9604d8eeafff81c4ca90fece1b0e5e446597c74a#diff-9f26cdfdb6e74804dd2eddb0eb902ed4b0391e67b972c1dde3e630b8ff41e293R87) (as indicated by [this comment in `IssueExtensionExecutionListenerTests`](https://github.com/junit-pioneer/junit-pioneer/blob/fc94b3ea2f7595e65851ebddc192885c6dadf4df/src/test/java/org/junitpioneer/jupiter/issue/IssueExtensionExecutionListenerTests.java#L38-L39)). I assume that is why the factory method cannot be used, and why the deprecated `add` method is used. But maybe these `add` calls are redundant, because `TestPlan` is already mocked, so the `add` calls don't seem to do anything. The tests still pass without them.

JUnit Pioneer currently uses the version matrix `[ '5.7.2', '5.8.2' ]` so it is not possible to use the static factory method without causing issues for either of these versions.

It would be good if one of you check this as well (when you have time) because you can better judge what the correct behavior should be.

src/test/java/org/junitpioneer/jupiter/CartesianProductTestExtensionTests.java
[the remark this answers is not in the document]
nipafx: No, was just a comment to express my annoyance. 😉

src/test/java/org/junitpioneer/jupiter/CartesianProductTestNameFormatterTests.java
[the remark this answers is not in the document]
nipafx: The test class itself can be deprecated.

nipafx (approved):

Michael1993 (approved):

nipafx:
Unfortunately, we can't merge this as is because there are three spurious warnings/errors:

* the use of `NullEnum.class` as attribute default in `CartesianEnumSource` (both deprecated) creates a warning that can't be suppressed - works on JDK 9+ without suppression
* there are two rawtype warnings that can probably be suppressed - works on JDK 10+ without suppression

Looks like we need to update our build to JDK 11 soon.

src/test/java/org/junitpioneer/jupiter/issue/IssueExtensionIntegrationTests.java
```diff
@@ -62,7 +62,7 @@ private static void assertStatus(IssueTestCase testCase) {
 			assertThat(testCase.result()).isEqualTo(Status.FAILED);
 	}
 
-	static class IssueIntegrationCases {
+	static class IssueIntegrationTestCase {
```
Marcono1234, line 65: Should this be "...TestCase<ins>s</ins>"? (with "s" at the end)

build.gradle.kts
```diff
@@ -215,14 +215,18 @@ tasks {
 
 	compileJava {
 		options.encoding = "UTF-8"
-		options.compilerArgs.add("-Werror")
-		options.compilerArgs.add("-Xlint:all")
+		if (!modularBuild.toBoolean()) {
+			options.compilerArgs.add("-Werror")
+			options.compilerArgs.add("-Xlint:all")
+		}
```
Marcono1234, line 221: @Michael1993, instead of completely ignoring warnings, you could probably in general ignore warnings for the `exports` group, see [`javac` documentation](https://docs.oracle.com/en/java/javase/17/docs/specs/man/javac.html#option-Xlint-custom):
```
options.compilerArgs.add("-Xlint:all,-exports")
```

However, except for the `...Extension.class` usage in annotations, many of these warnings seem legit. But it might make sense to address this in a separate pull request.

build.gradle.kts
```diff
@@ -215,14 +215,18 @@ tasks {
...
 	}
 
 	compileTestJava {
 		options.encoding = "UTF-8"
-		options.compilerArgs.add("-Werror")
-		options.compilerArgs.add("-Xlint:all")
+		if (!modularBuild.toBoolean()) {
```
Marcono1234, line 226: This is probably not necessary. The test sources don't seem to cause any problems for modular builds.

src/test/java/org/junitpioneer/jupiter/issue/IssueExtensionIntegrationTests.java
nipafx: Yes, good catch. 👍🏾

build.gradle.kts
nipafx: After going over the warnings, I decided, they're not helpful (see 5425bc5), so we ignore them.

Bukama:
Not `merge-ready` as there are compilation errors of test classes

Marcono1234:
Thank you for your assitance with this, or rather for taking over this pull request. Though you apparently removed GitHub's `Co-authored-by` from the commit message so the complete changes in `main` are now attributed to me 😅

src/main/module/module-info.java
```diff
@@ -1,4 +1,16 @@
+/**
+ * JUnit Pioneer provides extensions for <a href="https://github.com/junit-team/junit5/">JUnit 5</a>
+ * and its Jupiter API.
+ *
+ * <p>Pioneer does not limit itself to proven ideas with wide application but is purposely open to
+ * experiments. It aims to spin off successful and cohesive portions into sibling projects or back
+ * into the JUnit 5 code base.
+ *
+ * <p>The dependencies on Jupiter modules could be marked as <code>transitive</code> but that would
+ * allow users who depend on this module to not `require` org.junit.*, which would be backwards.
```
Marcono1234, line 10: @nipafx, should this use `{@code ...}`?
```suggestion
 * <p>The dependencies on Jupiter modules could be marked as {@code transitive} but that would
 * allow users who depend on this module to not {@code require} org.junit.*, which would be backwards.
```

PR #566, opened 2021-12-25
Improve Javadoc creation

### LINKED ISSUE #522: Generate Javadoc using JDK 17
~It allows to have the Javadoc from JDK 17.~

~Blocked by #521~

Advantages: https://github.com/junit-pioneer/junit-pioneer/issues/522#issuecomment-985885702

Marcono1234:
Resolves #522

Proposed commit message:

```
Improve Javadoc creation (#522 / #566)

Resolves: #522
PR: #566
```

Improves Javadoc creation by:
- Using [Gradle toolchains](https://docs.gradle.org/current/userguide/toolchains.html) to create the Javadoc with JDK 17
- Add external links to the JUnit Javadoc
- (Implicitly) add external links to the JDK Javadoc
- Enable doclint, except for the `missing` group

Future suggested work:
I currently do not completely understand why the project is built multiple times with different target Java versions and with different modularity settings. In my opinion (but maybe I am overlooking something):
- The project should always be built modular, but 

---
**PR checklist**

Documentation (general)

Documentation (new extension)

Code

Contributing

---

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -27,6 +27,8 @@ val modularBuild : String by project
 val experimentalJavaVersion : String? by project
 val experimentalBuild: Boolean = experimentalJavaVersion?.isNotEmpty() ?: false
 
+val targetJavaVersion = JavaVersion.VERSION_1_8
+
 java {
 	if (experimentalBuild) {
 		toolchain {
@@ -36,7 +38,7 @@ java {
 		sourceCompatibility = if (modularBuild.toBoolean()) {
 			JavaVersion.VERSION_11
 		} else {
-			JavaVersion.VERSION_1_8
+			targetJavaVersion
 		}
 	}
 	withJavadocJar()
@@ -97,7 +99,7 @@ jacoco {
 }
 
 sonarqube {
-	// If you want to use this logcally a sonarLogin has to be provide, either via Username and Password
+	// If you want to use this logically a sonarLogin has to be provided, either via Username and Password
 	// or via token, https://docs.sonarqube.org/latest/analysis/analysis-parameters/
 	properties {
 		// Default properties if somebody wants to execute it locally
@@ -225,12 +227,26 @@ tasks {
 	}
 
 	javadoc {
-		// of all the javadoc checks (accessibility, html, missing, reference, syntax; see
-		// https://docs.oracle.com/javase/8/docs/technotes/tools/unix/javadoc.html#BEJEFABE)
-		// disable the warning for missing comments and tags because they spam the output
-		// (it does often not make sense to comment every tag; e.g. the @return tag on annotations)
-		(options as CoreJavadocOptions).addStringOption("Xdoclint:accessibility,html,syntax,reference", "-quiet")
-		options.encoding = "UTF-8"
+		// Exclude internal implementation package from javadoc
+		exclude("org/junitpioneer/internal")
+
+		options {
+			// Cast to standard doclet options, see https://github.com/gradle/gradle/issues/7038#issuecomment-448294937
+			this as StandardJavadocDocletOptions
+
+			links = listOf("https://junit.org/junit5/docs/current/api/")
+
+			// Set javadoc `--release` flag (affects which warnings and errors are reported)
+			addStringOption("-release", targetJavaVersion.getMajorVersion())
+
+			// Enable doclint, but ignore warnings for missing tags, see
+			// https://docs.oracle.com/en/java/javase/17/docs/specs/man/javadoc.html#additional-options-provided-by-the-standard-doclet
+			// The Gradle option methods are rather misleading, but a boolean `true` value just makes sure the flag
+			// is passed to javadoc, see https://github.com/gradle/gradle/issues/2354
+			addBooleanOption("Xdoclint:all,-missing", true)
+			encoding = "UTF-8"
+		}
+
 		shouldRunAfter(test)
 	}
 
@@ -270,4 +286,11 @@ tasks {
 		githubToken = generateChangelogTask.githubToken
 		newTagRevision = System.getenv("GITHUB_SHA")
 	}
+}
+
+tasks.withType<Javadoc>().configureEach {
+	javadocTool.set(javaToolchains.javadocToolFor {
+		// Create Javadoc with newer JDK to get the latest features, e.g. search bar
+		languageVersion.set(JavaLanguageVersion.of(17))
+	})
 }
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -233,7 +233,11 @@ tasks {
 		})
 
 		// Exclude internal implementation package from javadoc
-		exclude("org/junitpioneer/internal")
+		/*
+		 * TODO: Currently disabled because modular Gradle build fails when this package is excluded
+		 * See https://stackoverflow.com/q/32785002 and slightly related Gradle issue https://github.com/gradle/gradle/issues/14066
+		*/
+		//exclude("org/junitpioneer/internal")
 
 		options {
 			// Cast to standard doclet options, see https://github.com/gradle/gradle/issues/7038#issuecomment-448294937
@@ -244,7 +248,8 @@ tasks {
 
 			// Set javadoc `--release` flag (affects which warnings and errors are reported)
 			// (Note: Gradle adds one leading '-' to the option on its own)
-			addStringOption("-release", targetJavaVersion.majorVersion)
+			// Have to use at least Java 9 to support modular build
+			addStringOption("-release", maxOf(9, targetJavaVersion.majorVersion.toInt()).toString())
 
 			// Enable doclint, but ignore warnings for missing tags, see
 			// https://docs.oracle.com/en/java/javase/17/docs/specs/man/javadoc.html#additional-options-provided-by-the-standard-doclet
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -99,7 +99,7 @@ jacoco {
 }
 
 sonarqube {
-	// If you want to use this logically a sonarLogin has to be provided, either via Username and Password
+	// If you want to use this locally a sonarLogin has to be provided, either via Username and Password
 	// or via token, https://docs.sonarqube.org/latest/analysis/analysis-parameters/
 	properties {
 		// Default properties if somebody wants to execute it locally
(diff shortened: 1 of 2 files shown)


build.gradle.kts
```diff
@@ -271,3 +287,10 @@ tasks {
 		newTagRevision = System.getenv("GITHUB_SHA")
 	}
 }
+
+tasks.withType<Javadoc>().configureEach {
```
Marcono1234, line 291: Ideally this would be part of the `javadoc` task definition above, but then it failed to find functions, possibly because the 'wrong' `this` was in scope.
Would be great if someone more familiar with the Gradle Kotlin DSL could give it a try.

Alternatively the complete `javadoc` block above could be put inside `tasks.javadoc { ... }` including the toolchain definition below, but then the tasks configurations are separated which might decrease readability.

build.gradle.kts
```diff
@@ -225,12 +227,26 @@ tasks {
...
+		exclude("org/junitpioneer/internal")
+
+		options {
+			// Cast to standard doclet options, see https://github.com/gradle/gradle/issues/7038#issuecomment-448294937
+			this as StandardJavadocDocletOptions
+
+			links = listOf("https://junit.org/junit5/docs/current/api/")
+
+			// Set javadoc `--release` flag (affects which warnings and errors are reported)
+			addStringOption("-release", targetJavaVersion.getMajorVersion())
```
Michael1993, line 240: Comment doesn't seem to match code here.

build.gradle.kts
```diff
@@ -225,12 +227,26 @@ tasks {
...
+
+			// Enable doclint, but ignore warnings for missing tags, see
+			// https://docs.oracle.com/en/java/javase/17/docs/specs/man/javadoc.html#additional-options-provided-by-the-standard-doclet
+			// The Gradle option methods are rather misleading, but a boolean `true` value just makes sure the flag
+			// is passed to javadoc, see https://github.com/gradle/gradle/issues/2354
+			addBooleanOption("Xdoclint:all,-missing", true)
```
Michael1993, line 246: I wonder if the same is true here... Is this `--missing` or just `-missing`?

build.gradle.kts
Marcono1234: Do you mean because of `--` vs. `-`? Gradle seems to prepend the options with a `-`, so `-release` becomes `--release`.
`javadoc` [expects `--release`](https://docs.oracle.com/en/java/javase/17/docs/specs/man/javadoc.html#options-for-javadoc) which is why I wrote that in the comment.

build.gradle.kts
Marcono1234: It is `-missing`. That is part of the `-Xdoclint` option, it is not a standalone option.
See https://docs.oracle.com/en/java/javase/17/docs/specs/man/javadoc.html#additional-options-provided-by-the-standard-doclet (also linked in the comment in the source above)

build.gradle.kts
Marcono1234: Also feel free to test the changes locally. For example without the `--release` option, if one of the classes would for example contain `private int _;`, a `javadoc` error would be shown because `_` is a reserved identifier for Java 17.
However, with `--release 8` option, it only causes a warning because Java 8 still permits it.

build.gradle.kts
Marcono1234: Also feel free to test the changes locally. Only using `Xdoclint:all` will show the undesired warnings about missing javadoc tags (e.g. for parameters). With `-missing` it only shows the other warnings, currently that the code does not use modules but links against the JUnit doc which uses modules, that can be ignored and could also be solved in the future with #567.

build.gradle.kts
```diff
@@ -225,12 +227,27 @@ tasks {
...
+
+			// Set javadoc `--release` flag (affects which warnings and errors are reported)
+			// (Note: Gradle adds one leading '-' to the option on its own)
+			addStringOption("-release", targetJavaVersion.getMajorVersion())
+
+			// Enable doclint, but ignore warnings for missing tags, see
+			// https://docs.oracle.com/en/java/javase/17/docs/specs/man/javadoc.html#additional-options-provided-by-the-standard-doclet
+			// The Gradle option methods are rather misleading, but a boolean `true` value just makes sure the flag
+			// is passed to javadoc, see https://github.com/gradle/gradle/issues/2354
+			addBooleanOption("Xdoclint:all,-missing", true)
```
beatngu13, line 247: Why `addBooleanOption` instead of `addStringOption`?

Also, the [Javadoc manual](https://docs.oracle.com/en/java/javase/17/docs/specs/man/javadoc.html#additional-options-provided-by-the-standard-doclet) says:

```text
-Xdoclint:(all|none|[-]group)
```

I'm not that familiar with the Javadoc CLI, but does it really make sense to combine `all` and `missing` (especially with a leading `-`)?

```diff
-Xdoclint:accessibility,html,syntax,reference
+Xdoclint:all,-missing
```

build.gradle.kts
```diff
@@ -225,12 +227,27 @@ tasks {
...
+			encoding = "UTF-8"
```
beatngu13, line 248: I would move this up just before `links`. Then, the `options` block becomes:

```kotlin
this as StandardJavadocDocletOptions

encoding = "UTF-8"
links = listOf("https://junit.org/junit5/docs/current/api/")

// options
// ...
```

IMHO a bit easier to read, what do you think?

beatngu13 (changes requested):
Another great PR, thank you!

build.gradle.kts
Marcono1234: > Why `addBooleanOption` instead of `addStringOption`?

This is explained a bit in the Gradle issue linked in the comment; the name and behavior of these methods is rather confusing. Apparently Gradle considers a "boolean option" an option without separate value, e.g. `-option`. Whereas a "string option" would be an option with separate value, e.g. `-option value`. Since the doclint setting is part of the same argument it has to be a "boolean option".
However, using a "string option" with `null` value (explicitly or when calling the `addStringOption` method without separate value parameter), the option seems to be simply omitted from the command line arguments (see [Gradle implementation](https://github.com/gradle/gradle/blob/fc26c4c3e3edb3ef3572a4b663fc452689f2ad19/subprojects/language-java/src/main/java/org/gradle/external/javadoc/internal/StringJavadocOptionFileOption.java#L28-L30)).

---

> I'm not that familiar with the Javadoc CLI, but does it really make sense to combine `all` and `missing` (especially with a leading `-`)?

The leading `-` excludes the group. I tried to explain that with the comment "Enable doclint, but ignore warnings for missing tags" a few lines above, but maybe that was not obvious; do you have any suggestions for improving the comment?
This configuration is mentioned by the following bullet point in the linked documentation:
> <code>-Xdoclint all,-<i>group</i></code>: Enables all checks except <i>group</i> checks

beatngu13: Thank you for that great explanation!

build.gradle.kts
Marcono1234: Thanks, that worked!

Michael1993 (approved):

beatngu13 (approved):

beatngu13:
@Marcono1234 after adding the `merge-ready` label, which enables the `full-featured` matrix build, Javadoc generation fails with the modular build on Java 17:

https://github.com/junit-pioneer/junit-pioneer/runs/4731184102?check_suite_focus=true

Would you mind having a look?

Marcono1234:
It looks like the issue is:
>  \> Task :javadoc FAILED
> error: option --module-path not allowed with target 8

The issue occurred when a modular build (`-PmodularBuild=true`) was performed. I have changed the Javadoc task to now always at least use Java 9 as `--release` version. However, that has the disadvantage that exclusion of packages does not work anymore (I have commented it out). The proper solution would likely be to use the `-exclude` Javadoc option; not to be confused with Gradle's `exclude` functions which are currently used in the script and which exclude source files completely (causing Javadoc failures for modular builds due to unresolved imports), instead of just skipping Javadoc generation for them. Unfortunately using `-exclude` does not seem to work due to how Gradle invokes the Javadoc tool, see comments I added to the build script. `-exclude` would probably only work if Gradle used `-subpackages`, but it does not use that at the moment.

I have also updated JaCoCo because the previous version did not support Java 17 which caused errors to be printed to the console. Though if you want I can also create a separate pull request for that.

build.gradle.kts
```diff
@@ -93,11 +95,11 @@ yamlValidator {
 }
 
 jacoco {
-	toolVersion = "0.8.6"
+	toolVersion = "0.8.7"
 }
 
 sonarqube {
-	// If you want to use this logcally a sonarLogin has to be provide, either via Username and Password
+	// If you want to use this logically a sonarLogin has to be provided, either via Username and Password
```
Michael1993, line 102: Wait, isn't this supposed to be "locally"? 😅

Marcono1234: Ah, that makes a lot more sense 😅
Have fixed it now, sorry.

build.gradle.kts
```diff
@@ -225,12 +227,37 @@ tasks {
...
+		javadocTool.set(project.javaToolchains.javadocToolFor {
+			// Create Javadoc with newer JDK to get the latest features, e.g. search bar
+			languageVersion.set(JavaLanguageVersion.of(17))
+		})
+
+		// Exclude internal implementation package from javadoc
+		/*
+		 * TODO: Currently disabled because modular Gradle build fails when this package is excluded
+		 * See https://stackoverflow.com/q/32785002 and slightly related Gradle issue https://github.com/gradle/gradle/issues/14066
+		*/
```
beatngu13, line 239: Missing alignment

```suggestion
		 */
```

build.gradle.kts
```diff
@@ -225,12 +227,37 @@ tasks {
...
+		//exclude("org/junitpioneer/internal")
```
beatngu13, line 240: Would this help?

```suggestion
		if (!modularBuild.toBoolean()) {
			exclude("org/junitpioneer/internal")
		}
```

beatngu13 (changes requested):

build.gradle.kts
Marcono1234: Thanks, have fixed it.

build.gradle.kts
Marcono1234: Yes that works, and is probably a better solution for now, have changed this.

beatngu13 (approved):

beatngu13:
Thanks for your perseverance – great PR!

PR #578, opened 2022-01-10
Fix tests not reverting system property and environment variable changes

Marcono1234:
Some of the tests related to the system property and environment variable extension were changing system properties and environment variables manually, but were not reverting the changes afterwards.

I have only added the necessary calls to remove the custom entries. A more extensive solution would be to store the existing value (if any) in a field and restore that value afterwards. However, because custom system property and environment variable names are used, there are most likely no existing old values, so that extra logic would probably be redundant.

Also, should these test classes be annotated with `@WritesSystemProperty` and `@WritesEnvironmentVariable`, or was that intentionally not done to prevent spurious test success in case these annotations behave incorrectly?
(And do the tests currently rely on the fact that the used system property and environment variable names are unique per test?)

Proposed commit message:

```
Fix tests not reverting system property and env variable changes (#578)

PR: #578
```

---
**PR checklist**

Documentation (general)

Documentation (new extension)

Code

Contributing

---

### DIFF
--- a/src/test/java/org/junitpioneer/jupiter/AbstractEntryBasedExtensionTests.java
+++ b/src/test/java/org/junitpioneer/jupiter/AbstractEntryBasedExtensionTests.java
@@ -12,6 +12,7 @@ package org.junitpioneer.jupiter;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
@@ -37,6 +38,13 @@ class AbstractEntryBasedExtensionTests {
 
 		System.clearProperty(CLEAR_SYSPROP_KEY);
 		System.setProperty(SET_SYSPROP_KEY, SET_SYSPROP_ORIGINAL_VALUE);
+	}
+
+	@AfterEach
+	void tearDown() {
+		// Revert changes to not affect other tests
+		EnvironmentVariableUtils.clear(SET_ENVVAR_KEY);
+		System.clearProperty(SET_SYSPROP_KEY);
 	}
 
 	@Test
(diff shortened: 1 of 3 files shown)


Michael1993 (commented):
Hm, I don't think this is necessary. `AbstractEntryBasedExtension` already implements `AfterEachCallback` and `AfterAllCallback` with resetting behaviour.

Marcono1234:
> `AbstractEntryBasedExtension` already implements `AfterEachCallback` and `AfterAllCallback` with resetting behaviour

But that is only used when you actually use the extension (e.g. through `@SetSystemProperty`). The changes done by this PR are for unit tests which explicitly set system properties and environment variables using `System.setProperty` and `EnvironmentVariableUtils.set`. And therefore, if I understand it correctly, they will by default not be cleaned up again and be visible to other unrelated tests.

src/test/java/org/junitpioneer/jupiter/AbstractEntryBasedExtensionTests.java
```diff
@@ -39,6 +40,13 @@ void setUp() {
 		System.setProperty(SET_SYSPROP_KEY, SET_SYSPROP_ORIGINAL_VALUE);
 	}
 
+	@AfterEach
+	void tearDown() {
+		// Revert changes to not affect other tests
```
beatngu13, line 45: I think you can remove that comment.

Also, isn't it now possible to remove the clear calls from the `@BeforeEach`?

beatngu13 (changes requested):

src/test/java/org/junitpioneer/jupiter/AbstractEntryBasedExtensionTests.java
Marcono1234: The clear calls are for different variables though (`CLEAR_...`) whereas these this is for the `SET_...` variables. Or do you mean something else?

nipafx:
@Marcono1234 I just noticed that you're not mentioned in the README under _Contributors_. We'll gladly have you there! If you're ok with that, feel free to add your various contributions in this PR.

Marcono1234:
Unless you really want me to be listed there, I don't mind not being listed there for my previous contributions.

nipafx:
It's totally up to you. Just know that you can add yourself in any PR if you want to.

nipafx (approved):

Issue #587, opened 2022-02-10
Improve GitHub Actions to provide more Feedback within Pull Requests


aepfli:
Our GitHub Actions do not provide a lot of information to the user why they are failing. There is always a second click needed to get to the action log.

Instead of this, we want to have more Feedback on the Pull Request:
[](url)

1. Use a reporter to post Unit Test Reports directly onto the Pull Request:
    - [JUnit Report Action](https://github.com/marketplace/actions/junit-report-action)
    - [Test Reporter](https://github.com/marketplace/actions/test-reporter)
    - [...](https://github.com/marketplace?category=&query=junit+sort%3Apopularity-desc&type=&verification=)
2. Add an own Action for just Spotless
3. Remove spotless from the test-executing builds

This should increase the GitHub Action overview within our Pull Request and reduce the amount of clicks to reach a certain stage.

p1729:
@aepfli can I give this a try ?

aepfli:
> @aepfli can I give this a try ? 

Sure give it a try

p1729:
@aepfli Could you please let me know for which test-executing job(s) we need the test result reporter action. Currently, we have 4 test-executing jobs in all. Will it be fine if we add this for the `basic` job?

beatngu13:
#611 improved the GHA build, but didn't address the lack of feedback mentioned in this issue, therefore, reopened.

beatngu13:
While I'm here: the new job summary feature might be helpful.

https://github.blog/changelog/2022-05-09-github-actions-enhance-your-actions-with-job-summaries/

nipafx:
@aepfli @p1729 or really @ anyone who's interested: Please let me know if you will work on this in the near future. If not, I might kick it out of the 2.0 milestone to get that one closer to the finishing line.

Bukama:
On our today's team meeting @beatngu13 and I decided to remove this issue form the 2.0 milestone as we don't want to block the release and as it is a feature not directly related to the final jar, but "only" to our build pipeline.

PR #611, opened 2022-03-18
Improve GitHub actions

p1729:
Proposed commit message:

```
Improve GitHub actions (#587 / #611)

Adds an own Action for just Spotless
Removes spotless from the test-executing builds

Closes: #587
PR: #611 
```

---
**PR checklist**

Documentation (general)

Documentation (new extension)

Code

Contributing

---


p1729:
@aepfli could you please review this PR for #587

aepfli:
Ah I see you used the annotation action, and already adapted the name. So this seems to be fine. ;)

.github/workflows/build.yml
```diff
@@ -78,13 +78,39 @@ jobs:
...
-          arguments: --refresh-dependencies --stacktrace --scan clean build
+          arguments: --refresh-dependencies --stacktrace --scan clean build -x spotlessCheck
       - name: SonarCloud analysis
         if: ${{ matrix.os == 'ubuntu' }}
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           SONAR_TOKEN: "61ab2579215aa8a0024a2f9368fc1298fdecfd18"
         run: ./gradlew jacocoTestReport sonarqube --stacktrace -i
+      - name: Test report
+        uses: mikepenz/action-junit-report@v2
```
beatngu13, line 89: Current version seems to be [v3.0.0](https://github.com/mikepenz/action-junit-report/releases/tag/v3.0.0):

```suggestion
        uses: mikepenz/action-junit-report@v3
```

.github/workflows/build.yml
```diff
@@ -78,13 +78,39 @@ jobs:
...
+        if: always() # always run even if the previous step fails
+        with:
+          check_name: "Test report for Java ${{ matrix.java }} on ${{ matrix.os }}"
+          report_paths: '**/build/test-results/test/TEST-*.xml'
+
+  # Code format check
+  code-format-check:
+    runs-on: ubuntu-latest
+    needs: validation
```
beatngu13, line 98: Maybe we now want to run this cheap `code-format-check` before `basic`, i.e. let `basic` need this?

.github/workflows/build.yml
```diff
@@ -78,13 +78,39 @@ jobs:
...
+          arguments: --refresh-dependencies --stacktrace --scan clean build -x spotlessCheck
       - name: SonarCloud analysis
         if: ${{ matrix.os == 'ubuntu' }}
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           SONAR_TOKEN: "61ab2579215aa8a0024a2f9368fc1298fdecfd18"
         run: ./gradlew jacocoTestReport sonarqube --stacktrace -i
+      - name: Test report
+        uses: mikepenz/action-junit-report@v2
+        if: always() # always run even if the previous step fails
```
beatngu13, line 90:
```suggestion
        if: always() # always report JUnit test result

File: .github/workflows/build.yml [hunk 4]
```diff
@@ -78,13 +78,39 @@ jobs:
...
+    needs: validation
+    timeout-minutes: 15
+    name: Code format check
+    steps:
+      - name: Check out repo
+        uses: actions/checkout@v2
+      - name: Set up Java
+        uses: actions/setup-java@v2
+        with:
+          java-version: 8
```
beatngu13, line 107: Meanwhile we use Java 11:

```suggestion
          java-version: 11
```

beatngu13 (changes requested):

beatngu13:
Nice PR, thank you!

Feel free to add yourself here: https://github.com/junit-pioneer/junit-pioneer#contributors

.github/workflows/build.yml
aepfli: The idea was to run all on parallel, so the feedback for the contributor is faster.

So it can be done in one swoop, and not two build iterations.

But what I thought is, maybe run spotless and one is, instead of 3 - and continue after those 2 build steps. Would this be a good idea?

The issue I see, is that it would make the waiting time for contributors longer, to get a full feedback on their contribution from CI system

beatngu13: But is the feedback faster than before? Currently we have:

```text
validation --> basic (incl. Spotless) --> ...
```

With this PR:

```text
validation --> basic (excl. Spotless) --> ...
            |
            -> code-format-check
```

Before, `basic` would fail (fast) in case of a Spotless error. Now, `basic` passes but `code-format-check` would fail, probably just as fast as `basic` before?

The motivation behind my suggestion was to save computation from our (I believe) 2,000 minutes/month limit. That is, don't run the `basic` matrix until spotless errors have been resolved.

aepfli: But from a contributor perspective there are two different paths:

Approach with separate spotless:
1. Open pr
2. Builds fails on spotless
3. Fix code
4. Wait for build
5. Build might fail for unit tests
6. Fix unit tests

Versus
In parallel:
1. Open pr
2. Build fails on spotless
3. Build fails on tests
4. Fix both

If we want to trim it, I would suggest to have just one basic job just for one environment in parallel with spotless. Most likely the tests will not fail on other OS (besides some special extension) and run them afterwards.

I think this would unify both worlds, WDYT? Less computation time, and fast feedback for all basic needs.

beatngu13: I see your point, it's a trade-off.

When you say "just one basic job", you mean something like?

```
validation --> basic (no matrix, ubuntu only) --> full-featured (merge-ready label)
            |
            -> code-format-check
```

?

aepfli: sorry i was only on my mobile so i had issue properly formatting. but yes something like you posted or:

```
validation --> basic (no matrix, ubuntu only) -
            |                                                          |-> matrix build (3os) --> full-featured (merge-ready label)
            -> code-format-check                      -
```

so 3 phase:
1. Security (gradle wrapper validation)
2. Basic Verification and checks (currently spotless and basic build, but could be increases)
3. Extended Verification
4. Fully Fledged ultimative pre merge stage

beatngu13: How would `basic` and the Ubuntu-part of `matrix build` differ? Would it be sufficient to run the matrix only with macOS and Windows, if `basic` already covers Ubuntu? Example:

```mermaid
flowchart LR
    v[validation]
    b[basic]
    c[code-format-check]
    x[x-os]
    f[full-featured]

    v --> b
    v --> c
    b --> x
    c --> x
    x --> f
```

(Just rembered that GitHub now supports [Mermaid](https://mermaid-js.github.io/).)

aepfli: i do think that this might be the way to go:

- we gain fast feedback as a contributor (to be honest there are seldomly issues with other OS's)
- we reduce the build times, we just have 3 jobs which will do all kind of verifications - and the rest at a later stage
- we still have full verification at the end but fast and bundled feedback

beatngu13: I will mark this as resolved, since there is now #614 for further discussion.

beatngu13 (approved):

Bukama:
I'll add the merge-ready tag, to see if the full build also works

nipafx:
Looks like the build department 😁 likes this PR, thanks for that @p1729! What about this comment, though? (cc @aepfli)

> But, it seems like we need to investigate into the token usage, as it is not working for forked repositories, this might be not as useful as intended.

p1729:
@nipafx @aepfli I was switching between `annotate_only` and checks creation 😄 but as @aepfli pointed out there are issues with GitHub permissions for forked repositories for creating new checks in the build by the action. Also, since this is my first contribution I was not able to get the feedback early based on my changes on this PR as it needs the approval to run the build. So, I started creating a PR on my forked repo and intentionally failing tests to verify them as annotations. 

Currently, I am not able to see the test failure annotations on the PR. Hence, I am following up with an [issue](https://github.com/mikepenz/action-junit-report/issues/550).

aepfli:
Maybe we should remove the annotating part for now, and handle this in a separate task. The split into smaller steps is already great and provides enough feedback to the contributors.

also maybe worth a shot, because it was also twitter by github themselves: https://github.com/test-summary/action

p1729:
@aepfli I have removed the test report action from this PR. We can take that up later as you suggested.

/cc: @nipafx

aepfli (approved):
lgtm

nipafx:
@p1729 I think you need to merge `main` into this branch to get the build green.

beatngu13 (approved):

beatngu13:
@p1729 thx again for that PR!

Changed the "closes" to "relates to" #587 since the reporting issue still exists.

Issue #594, opened 2022-02-17
Update JUnit to 5.8


nipafx:
The temporary directory/resource extension proposed in #348 / #491 requires parameter-level extension registration, which [was added in JUnit Jupiter 5.8.0](https://junit.org/junit5/docs/5.8.0/release-notes/index.html). Updating from 5.7 to 5.8 requires no code changes.

As per [contribution guidelines](/https://github.com/junit-pioneer/junit-pioneer/blob/main/CONTRIBUTING.md#junit-5), this update should be done in a separate issue (this one) and PR (to be opened). If no other features requires 5.8 before #348, a PR can be proposed and merged right before #491 is merged - use the following commit message (add correct PR number in header and body):

```
Update JUnit to 5.8 (#594 / #XYZ)

The temporary directory/resource extension proposed in #348 / #491
requires parameter-level extension registration, which was added in
JUnit Jupiter 5.8.0[1].

[1]: https://junit.org/junit5/docs/5.8.0/release-notes/index.html

Closes: #594
PR: #XYZ
```

PR #491, opened 2021-05-29
Create new temporary directory/resource extension

jbduncan:
Proposed commit message:

```
${action} (${issues} / ${pull-request}) [max 70 characters]

${body} [max 70 characters per line]

${references}: ${issues}
PR: ${pull-request}
```

---
**PR checklist**

The following checklist shall help the PR's author, the reviewers and maintainers to ensure the quality of this project.
It is based on our contributors guidelines, especially the ["writing code" section](https://github.com/junit-pioneer/junit-pioneer/blob/main/CONTRIBUTING.md#writing-code).
It shall help to check for completion of the listed points.
If a point does not apply to the given PR's changes, the corresponding entry can be simply marked as done. 

Documentation (general)
* [x] There is documentation (Javadoc and site documentation; added or updated)
* [x] There is implementation information to describe _why_ a non-obvious source code / solution got implemented
* [x] Site documentation has its own `.adoc` file in the `docs` folder, e.g. `docs/report-entries.adoc`
* [x] Only one sentence per line (especially in `.adoc` files)
* [x] Javadoc uses formal style, while sites documentation may use informal style

Documentation (new extension)
* [x] The `docs/docs-nav.yml` navigation has an entry for the new extension
* [x] The `package-info.java` contains information about the new extension

Code
* [x] Code adheres to code style, naming conventions etc.
* [x] Successful tests cover all changes
* [x] There are checks which validate correct / false usage / configuration of a functionality and there are tests to verify those checks
* [x] Tests use [AssertJ](https://assertj.github.io/doc/) or our own [PioneerAssert](https://github.com/junit-pioneer/junit-pioneer/blob/main/CONTRIBUTING.md#assertions) (which are based on AssertJ)

Contributing
* [ ] A prepared commit message exists
* [ ] The list of contributions inside `README.md` mentions the new contribution (real name optional) 

---

I hereby agree to the terms of the [JUnit Pioneer Contributor License Agreement](https://github.com/junit-pioneer/junit-pioneer/blob/main/CONTRIBUTING.md#junit-pioneer-contributor-license-agreement).

Issue #348, opened 2020-09-29
Create new temporary directory/resource extension


nipafx:
After Marc Philipp implemented the `@TempDir` extension in Pioneer, it was later also added to Jupiter. It didn't (and as of September 2020 doesn't) support custom file systems, but other than that is more feature-rich than Pioneer's variant. This creates an unpleasant situation where there are two extensions that solve the same problem with the same approach, but slightly different feature sets (and an identical name to boot).

That said, there are a lot of feature requests for Jupiter's variant (⚠️ _some of these are already implemented!_ ⚠️):

* [Multiple declarations of `@TempDir` in same parameter list or class declaration should fail](https://github.com/junit-team/junit5/issues/2219)
* [Support @TempDir on constructor parameters](https://github.com/junit-team/junit5/issues/1754)
* [Introduce cleanup mode in @TempDir](https://github.com/junit-team/junit5/issues/1814)
* [Multiple @TempDirs should result in multiple directories](https://github.com/junit-team/junit5/issues/1967)
* [Provide support for setting a custom parent directory for @TempDir](https://github.com/junit-team/junit5/issues/2088)
* [Support disabling cleanup for @TempDir](https://github.com/junit-team/junit5/issues/2159)
* [Support custom file system for `@TempDir`](https://github.com/junit-team/junit5/issues/2400)
* [Make `@TempDir` restore mint permissions before closing context to avoid `AccessDeniedException`](https://github.com/junit-team/junit5/issues/2609)

After evaluating the situation in #277, we decided to remove the extension before the 1.0 release with the goal to look for a better solution to the problem in the future (i.e. this issue). So the task is to come up with a temporary directory (or, more generally temporary resource) extension that covers _more_ than what Jupiter does, possibly with an approach that makes different trade-offs (so users of Jupiter and Pioneer have more choices to solve their specific problem).

One interesting proof of concept to look at is @sormuras': [1](https://github.com/sormuras/brahms#temporary-directory-as-resource), [2](https://github.com/sormuras/brahms/blob/master/src/test/java/integration/resource/ChickenAndEggTests.java)

nipafx:
For starters, we're just looking for a design proposal or proof of concept that shows the outline of the extension and how it could provide the features requested above.

jbduncan:
I just wanted to say that this issue is of interest to me, and that I've just signed up for hack.commit.push so that I can dedicate some time to work on it.

[...]

PR #612, opened 2022-03-27
Update JUnit to 5.8 (#594 / #612)

Bukama:
The temporary directory/resource extension proposed in #348 / #491
requires parameter-level extension registration, which was added in
JUnit Jupiter 5.8.0[^1].

[^1]: https://junit.org/junit5/docs/5.8.0/release-notes/index.html

Closes: #594
PR: #612

---

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -98,7 +98,7 @@ jobs:
     strategy:
       matrix:
         java: [ 11, 17 ]
-        junit-version: [ '5.7.2', '5.8.2' ]
+        junit-version: [ '5.8.2' ]
         modular: [true, false]
         os: [ubuntu, macos, windows]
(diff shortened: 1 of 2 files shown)


aepfli:
What is our future plan, do we want to only support one version in the future? I am just asking out of curiosity and we could reduce complexity within out pipeline, when we remove one axis from the matrix (the junit version)

aepfli (approved):
LGTM, but see my earlier comment

beatngu13 (approved):

Michael1993 (approved):

nipafx:
> What is our future plan, do we want to only support one version in the future?

We should build against our baseline and at least also against the newest released Jupiter version, so we know early when problems come down the line.

Issue #567, opened 2021-12-25
Simplify Gradle build


Marcono1234:
It appears currently the project is built multiple times by CI with different JDK versions and different modularity settings. Maybe I am overlooking something, but I think the following should be done to simplify the build and to also increase its reproducibility:
- The source and target Java version (or better the `--release` version) should be fixed and independent of the JDK version a user (or the CI server) has installed; [Gradle toolchains](https://docs.gradle.org/current/userguide/toolchains.html) could be used to ensure this
- Instead there could be different test executions which alter the JUnit version and the runtime Java version. For that Gradle toolchains and possibly the incubating [test suites feature](https://docs.gradle.org/7.3.3/release-notes.html#test-suites) could be used (maybe there are other solutions as well).
- The `module-info.class` should always be created, but it should be placed in a Multi-Release JAR folder. There are cases where build tools or plugins have issues with `module-info.class` being in the top-level directory of a Java 8 JAR file. The ModiTect plugin supports Multi-Release JARs by setting the `jvmVersion` value (should be set to 9 since that is the next higher version >= 9 after version 8 targeted by this project), see [documentation](https://github.com/moditect/moditect-gradle-plugin#addmainmoduleinfo).
This would also allow running the Javadoc task with `--release 9` then to have it generate a module overview, see also #566.

What do you think?

Bukama:
@aepfli @beatngu13

aepfli:
i took a look at the tool [Toolchains](https://docs.gradle.org/current/userguide/jvm_test_suite_plugin.html#test_suite_target) - and based on

> For the initial release of this plugin, each test suite has a single target. This results in a 1:1:1 relationship between test suite, test suite target and a matching [Test](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.testing.Test.html) task. The name of the Test task is derived from the suite name. Future iterations of the plugin will allow defining multiple targets based other attributes, such as a particular JDK/JRE runtime.

We might be a little bit early for jdk changes, but it could for sure work for the JUnit versions - but still seems to be a little bit tideous. i would wait for gradle to improve the build.

```
val list: List<String> = listOf("5.8", "5.8.1")
            list.forEach {
                val testSuite = register("testWithJava$it", JvmTestSuite::class) {
                    useJUnitJupiter()

                    dependencies {
                        implementation(project)
                        implementation(project.dependencies.platform("org.junit:junit-bom:$junitVersion"))
[...]

PR #614, opened 2022-03-28
Draft: Simplify Gradle Build (#567)

aepfli:
Utilizing `jvm-test-suite` to its full potential.
Removing deprecated configuration, and updating
Gradle.

Proposed commit message:

```

${references}: ${issues}
PR: ${pull-request}
```

TODO:
- improve toolchain usage - this is just a first quiet simply hack, and not in a final stage
- check if there are additional configurations needed for the test suites to be run, and also how to export the test results
- evaluate if there are any issues if we are using toolchain for source compatibility etc. (we can use the release option all the time as fallback)
- write some good test verification - as this is fiddeling with the whole build, we want to be sure to test as much as possible.


aepfli:
@beatngu13 @Marcono1234  i am curious, what you think about such an approach - it is for now just an experiment from my side - do you think it makes sense to pursue this?

aepfli:
adding label `merge-ready` to see the time impact this has

.github/workflows/build.yml
```diff
@@ -98,10 +98,12 @@ jobs:
     strategy:
       matrix:
         java: [ 11, 17 ]
-        junit-version: [ '5.7.2', '5.8.2' ]
         modular: [true, false]
         os: [ubuntu, macos, windows]
-    name: with Java ${{ matrix.java }}, JUnit ${{ matrix.junit-version }}, Modular ${{ matrix.modular }} on ${{ matrix.os }}
+        exclude:
+          - modular: false # this was run already in the previous stage, no need for a duplicated run
+            java: 11
```
aepfli, line 105: i will extract this into an own improvement, we can save 3 jobs by this :)

build.gradle.kts
```diff
@@ -23,23 +24,16 @@ plugins.withType<JavaPlugin>().configureEach {
 group = "org.junit-pioneer"
 description = "JUnit 5 Extension Pack"
 
-val modularBuild : String by project
-val experimentalJavaVersion : String? by project
+val modularBuild: String by project
+val experimentalJavaVersion: String? by project
```
aepfli, line 28: in the future we could use this to actually provide our targeted version - and utilize toolchains fully - we have to set the --release option properly

build.gradle.kts
```diff
@@ -188,78 +187,113 @@ nexusPublishing {
...
-				"-XX:+IgnoreUnrecognizedVMOptions",
-				"--add-opens=java.base/java.util=ALL-UNNAMED")
-	}
 
-	testing {
-		suites {
-			val test by getting(JvmTestSuite::class) {
+		supportedJUnitVersions.split(",").filter { it != junitVersion }.forEach {
+			val testSuite = register("testWithJUnit$it", JvmTestSuite::class) {
 				useJUnitJupiter()
```
Marcono1234, line 242: Should this pass the JUnit version as argument to this call?

build.gradle.kts
```diff
@@ -188,78 +187,113 @@ nexusPublishing {
...
+							filter {
+								includeTestsMatching("*Tests")
+							}
+						}
 					}
-				} } }
+				}
 			}
+
+			versionTest.dependsOn(testSuite)
```
Marcono1234, line 277: It looks like currently Gradle aborts as soon as one test suite fails. Maybe it would be useful to have `versionTest` run all test suites and fail only afterwards. What do you think? And are you aware of a way in which this could be implemented?

build.gradle.kts
```diff
@@ -188,78 +187,113 @@ nexusPublishing {
...
+						srcDir("src/test/java")
+					}
+					resources {
+						srcDir("src/test/resources")
+					}
+				}
+
+				targets {
+					all {
+						testTask.configure {
```
Marcono1234, line 266: It looks like you could also specify the Java version to run the test here:
```kotlin
javaLauncher.set(javaToolchains.launcherFor {
    languageVersion.set(JavaLanguageVersion.of(javaVersion))
})
```

Then it might not be necessary anymore to run the complete build with different JDK versions (unless that is desired).

Marcono1234 (commented):
That looks pretty interesting already! Hopefully the other maintainers of you think this is going in the right way as well.

In the future it might also be possible to always perform a modular build, and then add blackbox module testing with a separate test suite (and separate test sources) as well.

Though please note that I am not that familiar with Gradle, so maybe there are better solutions to what I write in my review comments.

gradle.properties
```diff
@@ -1,6 +1,7 @@
-junitVersion=5.7.2
+junitVersion=5.8.2
 
 # Ensure sufficient heap size, especially for Sonar.
 org.gradle.jvmargs=-Xmx8g
 modularBuild=false
 org.gradle.java.installations.fromEnv=JDK_EXPERIMENTAL
+supportedJUnitVersions=5.8.1,5.8.2
```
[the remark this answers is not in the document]
aepfli: correct for testing

build.gradle.kts
```diff
@@ -188,78 +187,113 @@ nexusPublishing {
...
+
+	sourceSets {
+		main {
+			if (modularBuild.toBoolean())
+				java.srcDir("src/main/module")
 		}
 	}
 
+	// All compile Tasks should now use UTF-8 not just JavaCompile
+	withType<JavaCompile>().configureEach {
```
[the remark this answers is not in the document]
aepfli: before we set the utf-8 for javacompile and javatestcompile separately - but now we also do have other testcompiles etc. And all of them should be utf8 encoded - with this, we have it once and for all

build.gradle.kts
```diff
@@ -188,78 +187,113 @@ nexusPublishing {
...
-		// IF YOU ADD MORE OPTIONS; CONSIDER REPLACING `-XX:+IgnoreUnrecognizedVMOptions WITH A CONDITIONAL
-		jvmArgs(
-				"-XX:+IgnoreUnrecognizedVMOptions",
-				"--add-opens=java.base/java.util=ALL-UNNAMED")
-	}
 
-	testing {
-		suites {
-			val test by getting(JvmTestSuite::class) {
+		supportedJUnitVersions.split(",").filter { it != junitVersion }.forEach {
```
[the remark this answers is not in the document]
aepfli: it is a thing - but  there might be a nicer version.

Generally speaking, this reduced the CI minutes - we only have 1 job to prep and boot up - instead of one per junit verison. the whole orchestration is removed, and we are down to just a simple additional test execution. So in the end it makes our builds faster.

Additionally i am a fan of, having less configuration in CI and make it easy for people to use ci like feature locally - with this, we have an own gradle target to test multiple versions at once. So i do think this makes the local development process easier and nice

.github/workflows/build.yml
aepfli: i was talking about the exclude - we run those 3 jobs within our standard phase, we do not need to run them for java 11 and modular false again. 3 job less in the whole pipeline

build.gradle.kts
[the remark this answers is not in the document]
Bukama: > with this, we have an own gradle target to test multiple versions at once. So i do think this makes the local development process easier and nice

Sounds to me like: You can test multiple versions locally, but our CI does not anymore.
This would further mean that we force the users to run their changes locally instead of seeing impacts on Github and decide there what (not) to do?!?

nipafx:
@aepfli Since the goal is to simplify the build, let's go all in. Let's assume Pioneer 2.0 will target Java 11 and that this PR will be merged last thing before its release. Then we can get rid of:

* JDK 8 build
* non-modular build

This would also mean that you can copy the module declaration into the source folder.

Hm, now that I wrote that, I wonder whether it wouldn't be cleaner to first merge the new baseline(11 + module) and then soon after this PR. Either way, you can already make the changes here.

Bukama:
With PR #627 the Java version would be 11 and modular build will be default. I already removed the parts of the build flow there. So I think this PR needs a rebase after #627 is merged

nipafx:
#627 won't be merged for a while (because we can't release a 1.x version after that anymore), so this PR here is hopefully merged before that. @aepfli Can you take a look at this (and #611) so we make progress on the build? :)

build.gradle.kts
aepfli: looks good, i added that

build.gradle.kts
aepfli: also added something to achieve this behaviour, although it is not as clean with kotlin dsl as with groovy one

aepfli:
> #627 won't be merged for a while (because we can't release a 1.x version after that anymore), so this PR here is hopefully merged before that. @aepfli Can you take a look at this (and #611) so we make progress on the build? :)

I would actually get first rid of all the modularity etc. and than continue from there, it is just a lot of bloat and will most likely need additional efforts, which is not really worth it, if we plan to upgrade to 11 anyways

nipafx:
Since this branch already has and will soon have way more conflicts with main, I will close the PR for now. We can reopen it once 2.0 is out and the conflicts have been resolved.

PR #619, opened 2022-04-10
Update AssertJ

Michael1993:
Proposed commit message:

```
${action} (${issues} / ${pull-request}) [max 70 characters]

${body} [max 70 characters per line]

${references}: ${issues}
PR: ${pull-request}
```

---
**PR checklist**

The following checklist shall help the PR's author, the reviewers and maintainers to ensure the quality of this project.
It is based on our contributors guidelines, especially the ["writing code" section](https://github.com/junit-pioneer/junit-pioneer/blob/main/CONTRIBUTING.md#writing-code).
It shall help to check for completion of the listed points.
If a point does not apply to the given PR's changes, the corresponding entry can be simply marked as done. 

Documentation (general)
* [ ] There is documentation (Javadoc and site documentation; added or updated)
* [ ] There is implementation information to describe _why_ a non-obvious source code / solution got implemented
* [ ] Site documentation has its own `.adoc` file in the `docs` folder, e.g. `docs/report-entries.adoc`
* [ ] Site documentation in `.adoc` file references demo in `src/demo/java` instead of containing code blocks as text
* [ ] Only one sentence per line (especially in `.adoc` files)
* [ ] Javadoc uses formal style, while sites documentation may use informal style

Documentation (new extension)
* [ ] The `docs/docs-nav.yml` navigation has an entry for the new extension
* [ ] The `package-info.java` contains information about the new extension

Code
* [ ] Code adheres to code style, naming conventions etc.
* [ ] Successful tests cover all changes
* [ ] There are checks which validate correct / false usage / configuration of a functionality and there are tests to verify those checks
* [ ] Tests use [AssertJ](https://assertj.github.io/doc/) or our own [PioneerAssert](https://github.com/junit-pioneer/junit-pioneer/blob/main/CONTRIBUTING.md#assertions) (which are based on AssertJ)

Contributing
* [ ] A prepared commit message exists
* [ ] The list of contributions inside `README.md` mentions the new contribution (real name optional)

Issue #618, opened 2022-04-10
Update AssertJ to a newer version


Bukama:
Our assertion library `AssertJ` should be updated to a newer version. But not to just bumping the version number, but make then proper / improving the use of generic `SELF` assertions.

nipafx:
Thanks to the harsh compiler flags, the build fails after the update with raw types in one of our classes. I didn't investigate but it feels like we can improve the API by returning a more specific type, maybe `SELF`?

Michael1993:
I'm not really sure what improvements could be made. Could you give me an example @Bukama ?

scordio:
Gradle noob question: how to reproduce the offending warnings/errors? I did a few builds but didn't spot anything strange with the branch from #619.

PR #625, opened 2022-04-21
Return type-aware assertion on `withExceptionInstanceOf`

scordio:
This targets the branch of #619.

I'm not sure if this is what @nipafx meant at https://github.com/junit-pioneer/junit-pioneer/issues/618#issuecomment-1094266234, but it's an improvement I would add with the new version of AssertJ.

A benefit of this change is chaining `extracting` calls that would target the throwable concrete type (some examples at assertj/assertj-core#2242).

I also replaced `new ThrowableAssert<>` with `assertThat` - no obvious difference here but I suggest sticking to the factory methods instead of instantiating assertion classes directly.

### DIFF
--- a/src/test/java/org/junitpioneer/testkit/assertion/TestCaseAssertBase.java
+++ b/src/test/java/org/junitpioneer/testkit/assertion/TestCaseAssertBase.java
@@ -17,7 +17,7 @@ import java.util.function.Consumer;
 import java.util.function.Predicate;
 
 import org.assertj.core.api.AbstractThrowableAssert;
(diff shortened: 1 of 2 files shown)


Michael1993 (approved):

Michael1993 (approved):

Michael1993:
Let me know when you are finished so I can merge this into the draft PR - don't forget to add your contribution to the README

scordio:
Finished! Spotless and I are on the same page now 😄

No need to add an entry to the README, it's a tiny thing but happy to help if more comes out of #618.

PR #628, opened 2022-05-08
Allow non-static Cartesian factory methods with PER_CLASS lifecycle

robtimus:
```
Allow non-static Cartesian factory methods with PER_CLASS (#628)

JUnit allows non-static factory methods as long as the test class is
annotated with `@Testinstance(Lifecycle.PER_CLASS)`. The same is
currently not true for `@CartesianTest.MethodFactory`.
This commit fixes that.

The existing non-static test is unmodified, indicating that
non-static is still not allowed if the test class is not annotated
with `@Testinstance(Lifecycle.PER_CLASS)`.

PR: #628
```

---
**PR checklist**

Documentation (general)

Code
* [X] There are checks which validate correct / false usage / configuration of a functionality and there are tests to verify those checks
* [X] Tests use [AssertJ](https://assertj.github.io/doc/) or our own [PioneerAssert](https://github.com/junit-pioneer/junit-pioneer/blob/main/CONTRIBUTING.md#assertions) (which are based on AssertJ)

Contributing
* [X] The list of contributions inside `README.md` mentions the new contribution (real name optional)

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,7 @@ The least we can do is to thank them and list some of their accomplishments here
 
 * [Pankaj Kumar](https://github.com/p1729) contributed towards improving GitHub actions (#587 / #611)
 * [Filip Hrisafov](https://github.com/filiphr) contributed the [JSON Argument Source](https://junit-pioneer.org/docs/json-argument-source/) support (#101 / #492)
+* [Rob Spoor](https://github.com/robtimus) made it possible to use non-static factory methods for `@CartesianTest.MethodFactory`, as long as the test class is annotated with `@TestInstance(Lifecycle.PER_CLASS)` (#628)
 
 #### 2021
 
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ The least we can do is to thank them and list some of their accomplishments here
 
 * [Pankaj Kumar](https://github.com/p1729) contributed towards improving GitHub actions (#587 / #611)
 * [Filip Hrisafov](https://github.com/filiphr) contributed the [JSON Argument Source](https://junit-pioneer.org/docs/json-argument-source/) support (#101 / #492)
-* [Rob Spoor](https://github.com/robtimus) made it possible to use non-static factory methods for `@CartesianTest.MethodFactory`, as long as the test class is annotated with `@TestInstance(Lifecycle.PER_CLASS)` (#628)
+* [Rob Spoor](https://github.com/robtimus) enabled non-static factory methods for `@CartesianTest.MethodFactory` (#628)
 
 #### 2021
 
--- a/docs/cartesian-product.adoc
+++ b/docs/cartesian-product.adoc
@@ -288,12 +288,19 @@ You can reuse the same argument provider method multiple times.
 include::{demo}[tag=cartesian_argument_sets_reuse]
 ----
 
+You can use an argument provider method in nested classes, as long as these are annotated with `@TestInstance(Lifecycle.PER_CLASS)`.
+
+[source,java,indent=0]
+----
+include::{demo}[tag=cartesian_argument_sets_with_nested_classes]
+----
+
 ==== Requirements for the factory method
 
 There are multiple requirements the factory method has to fulfill to qualify:
 
 - must have the same name as the test method (or its name must be specified via the `factory` attribute)
-- must be `static` (unless the test class is annotated with `@TestInstance(Lifecycle.PER_CLASS)`)
+- must be `static` (unless the test class is annotated with `@TestInstance(Lifecycle.PER_CLASS)` and the factory method is defined in the test class)
 - must have **no** parameters
 - must return `ArgumentSets`
 - must register values for every parameter exactly once
--- a/docs/cartesian-product.adoc
+++ b/docs/cartesian-product.adoc
@@ -16,7 +16,7 @@ From Wikipedia:
 What does all this mean?
 
 The Cartesian product of sets is all the possible combinations where you take a single element from each set.
-If you have two sets, `{ 1, 2 }` and `{ 3, 4 }`, their cartesian product is `{ { 1, 3 }, { 1, 4 }, { 2, 3 }, { 2, 4 } }`.
+If you have two sets, `{ 1, 2 }` and `{ 3, 4 }`, their Cartesian product is `{ { 1, 3 }, { 1, 4 }, { 2, 3 }, { 2, 4 } }`.
 
 Sometimes it's useful to test all possible combinations of parameter sets.
 Normally, this results in a lot of written test data parameters.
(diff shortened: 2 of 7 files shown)


Michael1993 (changes requested):
Hi! 👋 
Thank you for this, looks great so far! Could you please also add an example with a non-static factory method to the documentation? Either to showcase the wrong usage (i.e. with a missing `@TestInstance` annotation) or the correct usage.

Michael1993:
Don't forget spotless! 👿

rob-spoor:
> Don't forget spotless! 👿

The errors are an update in the year in the licenses. I think that the spotless configuration needs to be updated, because the second year should be 2022, not 2021. Spotless suggest changing it back to 2021.

Michael1993:
Please rebase or merge main.

Michael1993:
> > Don't forget spotless! 👿
> 
> The errors are an update in the year in the licenses. I think that the spotless configuration needs to be updated, because the second year should be 2022, not 2021. Spotless suggest changing it back to 2021.

Whoops. 😅

robtimus:
The rebase is complete, I'll work on the documentation shortly.

robtimus:
I've added the requested example.

While working on the example I found a case where my change would have caused failures - using a non-static factory method from another class. I've improved the check that determines whether or not the factory method must be required, and added some tests. There are now three, in addition to the unmodified tests that were already on `main`:
* Using a non-static factory method from another class in a class annotated with `@TestInstance(PER_CLASS)` still fails.
* Using a static factory method from another class in a class annotated with `@TestInstance(PER_CLASS)` still works.
* Using a non-static factory method from the same class in a class annotated with `@TestInstance(PER_CLASS)` works (I already added this test before).

As a bonus I added support for nested test classes in `PioneerTestKit`. I made those changes generic enough to also support entire nested test classes and no-argument methods in nested test classes.

Michael1993 (approved):
LGTM 👍 

Thank you!

docs/cartesian-product.adoc
```diff
@@ -387,23 +387,46 @@ class MyCartesianTestClass {
...
-==== Requirements for the static factory method
+You can use an argument provider method in nested classes, as long as these are annotated with `@TestInstance(Lifecycle.PER_CLASS)`.
 
-There are multiple requirements the static factory method has to fulfill to qualify:
+[source,java]
+----
+class MyCartesianTestClass {
+	@Nested
+	@TestInstance(Lifecycle.PER_CLASS)
+	MyNestedCartesianTestClass {
```
nipafx, line 397:
```suggestion
	class MyNestedCartesianTestClass {
```

Right?

docs/cartesian-product.adoc
```diff
@@ -387,23 +387,46 @@ class MyCartesianTestClass {
...
+		@CartesianTest
+		@CartesianTest.MethodFactory("provideArguments")
+		void testNeedingArguments(String string, int i) {
+			// passing test code
+		}
+
+		ArgumentSets provideArguments() {
```
nipafx, line 404:
```suggestion
		// this provider method doesn't have to be static
		ArgumentSets provideArguments() {
```

nipafx (commented):
I started a review a week ago and forgot to send it, stupid me. Just added another comment - can you take a look at this?

docs/cartesian-product.adoc
```diff
@@ -387,23 +387,46 @@ class MyCartesianTestClass {
...
+
+There are multiple requirements the factory method has to fulfill to qualify:
 
 - must have the same name as the test method (or its name must be specified via the `factory` attribute)
-- must be `static`
+- must be `static` (unless the test class is annotated with `@TestInstance(Lifecycle.PER_CLASS)` and the factory method is defined in the test class)
 - must have **no** parameters
 - must return `ArgumentSets`
 - must register values for every parameter exactly once
 - must register values in order
```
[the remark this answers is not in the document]
robtimus: Didn't I say that:

>  and the factory method is defined in the test class

docs/cartesian-product.adoc
robtimus: Right.

I did a rebase on main, and extracted the example to `CartesianTestExtensionDemo`, so these errors should now be caught..

docs/cartesian-product.adoc
robtimus: In fact, it can't even be static because it's in a non-static nested class. I've added a comment about that, as well as why the `@TestInstance` annotation is necessary.

robtimus:
@nipafx can you check again? I've rebased on main and split off the example, as was recently done on main for other Cartesian test examples.

docs/cartesian-product.adoc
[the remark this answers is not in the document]
nipafx: You did. 🤦🏾

src/demo/java/org/junitpioneer/jupiter/cartesian/CartesianTestExtensionDemo.java
```diff
@@ -279,4 +282,30 @@ void testWithCustomDisplayName(@Values(strings = { "0", "1" }) String a,
...
+		@TestInstance(Lifecycle.PER_CLASS)
+		class MyNestedCartesianTestClass {
+
+			@CartesianTest
+			@CartesianTest.MethodFactory("provideArguments")
+			void testNeedingArguments(String string, int i) {
+				// passing test code
+			}
+
+			// this provider method cannot be static because it is in a non-static nested class
```
nipafx, line 299: Since Java 16, this is not longer correct. To make records (which are always static) easier to use in non-static nested classes, this limitation was lifted and non-static nested classes can now have static members. See [this compiling example](https://github.com/nipafx/demo-java-x/blob/main/src/main/java/dev/nipafx/demo/java16/lang/staticinner/OuterClass.java).

I think it's enough to point out that the method doesn't have to be static, but if you want, you can also add that it couldn't even be until Java 16.

README.md
```diff
@@ -119,6 +119,7 @@ The least we can do is to thank them and list some of their accomplishments here
 #### 2022
 
 * [Filip Hrisafov](https://github.com/filiphr) contributed the [JSON Argument Source](https://junit-pioneer.org/docs/json-argument-source/) support (#101 / #492)
+* [Rob Spoor](https://github.com/robtimus) made it possible to use non-static factory methods for `@CartesianTest.MethodFactory`, as long as the test class is annotated with `@TestInstance(Lifecycle.PER_CLASS)` (#628)
```
nipafx, line 122:
```suggestion
* [Rob Spoor](https://github.com/robtimus) enabled non-static factory methods for `@CartesianTest.MethodFactory` (#628)
```

No reason to go into detail here.

robtimus: Updated

src/demo/java/org/junitpioneer/jupiter/cartesian/CartesianTestExtensionDemo.java
robtimus: I wasn't aware of this. I've used your original suggestion.

nipafx:
Merged - thank you for your contribution @robtimus! :heart:

robtimus:
You're welcome 😄

PR #629, opened 2022-05-11
Register Jackson modules in JacksonJsonConverter (624 / 629)

robtimus:
The default JacksonJsonConverter does not support classes from
java.time when using @JsonSource (and probably also @JsonFileSource
and @JsonClassPathSource) in combination with POJO method arguments.
This is the case even if JavaTimeModule is available on the class
path.

By calling findAndRegisterModules(), any module that's available on
the class path and is registered as a service, will be available for
JSON processing.

Fixes: 624
PR: 629


Bukama (approved):
Thank you for the PR.

Again on of the little details one has to know with modules :O

Michael1993:
🤔

I'm not sure if this is an actual solution...? I think that while our tests pass with flying colors, other projects depending on Pioneer might still fail if we only add the jsr310 as a test time dependency.

Michael1993:
Spotless, ruining the fun for everyone (again). 😔

Michael1993:
Please rebase or merge main.

README.md
```diff
@@ -119,6 +119,7 @@ The least we can do is to thank them and list some of their accomplishments here
 #### 2022
 
 * [Filip Hrisafov](https://github.com/filiphr) contributed the [JSON Argument Source](https://junit-pioneer.org/docs/json-argument-source/) support (#101 / #492)
+* [Rob Spoor](https://github.com/robtimus) added registration of Jackson modules for use in `@JsonSource`, `@JsonFileSource` and `@JsonClasspathSource`. This allows modules like `JavaTimeModule` to be used (#624 / #629)
```
nipafx, line 122:
```suggestion
* [Rob Spoor](https://github.com/robtimus) added registration of Jackson modules (e.g. `JavaTimeModule`) JSON sources (#624 / #629)
```

nipafx (commented):
Thanks for the contribution @robtimus! ❤️ Two small changes.

build.gradle.kts
```diff
@@ -53,7 +53,8 @@ repositories {
 }
 
 val junitVersion : String by project
-val jacksonVersion: String = "2.13.2.2"
+val jacksonVersion: String = "2.13.2"
+val jacksonDatabindVersion: String = jacksonVersion + ".2"
```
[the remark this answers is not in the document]
robtimus: At the moment there need to be two versions. The reason is that only Jackson databind uses version 2.13.2.2. All other Jackson artifacts still use 2.13.2.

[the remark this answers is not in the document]
robtimus: Anyway, I'll update this, as the two lines are still next to each other.

README.md
robtimus: I felt that it didn't read really well. I've set changed it to "added registration of Jackson modules (...) *for* JSON sources"

robtimus:
> 🤔
> 
> I'm not sure if this is an actual solution...? I think that while our tests pass with flying colors, other projects depending on Pioneer might still fail if we only add the jsr310 as a test time dependency.

@Michael1993 I've used `findAndRegisterModules()`, so *any* module that is available through the ServiceLoader mechanism (which Jackson uses for this method) will be found. That's also why I didn't add the JSR310 dependency as a proper dependency. The Pioneer user can add any module they want themselves, and these will become available.

filiphr:
I agree with @robtimus regarding the JSR310 dependency. It is up to the users to decide which jackson modles they want to have. If the goal would be to have out of the box support for certain things in the [`JacksonNode`](https://github.com/junit-pioneer/junit-pioneer/blob/main/src/main/java/org/junitpioneer/jupiter/json/JacksonNode.java) from JUnit Pioneer.

@robtimus have you looked in a solution similar to how Spring does it in their [`Jackson2ObjectMapperBuilder#registerWellKnownModulesIfAvailable`](https://github.com/spring-projects/spring-framework/blob/a6cd8a78e2b0242bb45035e322c16ee98acf88e6/spring-web/src/main/java/org/springframework/http/converter/json/Jackson2ObjectMapperBuilder.java#L816)? Where JUnit Pioneer will register the well known modules in its created `ObjectMapper`?

robtimus:
I did not, but a quick look shows it's an alternative to `findModulesViaServiceLoader`. If that would be used as solution, then there would be support for `java.time`, `Optional`, and things from Kotlin, provided that the necessary dependencies are present. However, if the user wants some other third-party library that's not "well-known", they won't be able to use it. By using `findAndRegisterModules`, the user gets more control.

Bukama:
> I did not, but a quick look shows it's an alternative to `findModulesViaServiceLoader`. If that would be used as solution, then there would be support for `java.time`, `Optional`, and things from Kotlin, provided that the necessary dependencies are present. However, if the user wants some other third-party library that's not "well-known", they won't be able to use it. By using `findAndRegisterModules`, the user gets more control.

I like this approach as the user is more flexible. Please add this information into the docs.

filiphr:
>By using `findAndRegisterModules`, the user gets more control.

Actually with this the user gets less control. This means that any dependency that has a module registered through the Service Loader can cause a problem. Registering well known dependency guarantees that things will work with the well known dependencies.

I would suggest adding an SPI that would allow anyone to provide a custom `JacksonJsonConverter` and Pioneer will use it to get the `ObjectMapper` that it needs. This would actually be similar to the [`Jackson2ObjectMapperProvider`](https://github.com/lukas-krecan/JsonUnit/blob/master/json-unit-core/src/main/java/net/javacrumbs/jsonunit/providers/Jackson2ObjectMapperProvider.java) from the JsonUnit project.

nipafx:
As I understand this exchange, Pioneer has three options to register JSON modules itself:

* none (current implementation)
* `findModulesViaServiceLoader` (only "well-known" JSON modules; whatever makes them "well-known")
* `findAndRegisterModules` (all JSON modules)

None of that gives the user any configuration options withing Pioneer, so beyond that, we could offer:

* a way to pick one of the three options
* registering a "Jackson configurer" (which isn't a word 😊)
* registering a custom converter

I think the second part (offering more custom configuration) could be considered but should definitely be its own issue. I'll leave it an interested party to create one.

So let's focus on the first part and decide which default behavior makes the most sense. I'm torn because I have experienced more than once that some dependency registered a service that I don't care about, which then wrecks some other mechanism that uses that service. And while this is super frustrating, it's pretty rare. Intentionally adding a JSON module that Pioneer then ignores is a much more common albeit somewhat less frustrating scenario and I think it has the overall larger frustration integral. 😉 

So I'm inclined to go with `findAndRegisterModules`. I'm with @Bukama that this should be documented.

@robtimus: I just merged an improved documentation of the JSON source into `main`. If you merge or rebase this branch, you'll see that `json-argument-source.adoc` contains a section _JSON Parser Integration_ with a list-of-one that contains Jackson. Mentioning the module registration there (ideally in the form of a link to the documentation that goes into details) would be appropriate.

README.md
nipafx: Yeah, my proposal didn't grammar. 😉

filiphr:
>As I understand this exchange, Pioneer has three options to register JSON modules itself:
>
> * none (current implementation)
> * `findModulesViaServiceLoader` (only "well-known" JSON modules; whatever makes them "well-known")
> * `findAndRegisterModules` (all JSON modules)

* `findModulesViaServiceLoader` doesn't exist. I was referring to an example that Spring does, by checking if certain known modules are available (if the class is accessible) and registering them. 

* `findAndRegisterModules` are all JSON modules that are registered through the Service Loader

I am really against `findAndRegisterModules`, considering the fact that I have had numerous problems with some unexpected dependencies registering their own Jackson modules and wrecking havoc. If we look at the Spring example we can see that the known modules are the `Jdk8Module` and the `JavaTimeModule` (which are part of the jackson-databind) and the `KotlinModule` which is part of another dependency. Considering the fact that Pioneer is already Java 8 based the `Jdk8Module` and `JavaTimeModule` can be manually registered automatically already in the current implementation. 

Another problem with the `findAndRegisterModules` is the fact that it is a greedy approach, which means that if you want to change it at any time it would be considered as a breaking change.

nipafx:
(Thanks for explaining `findModulesViaServiceLoader` - got it now.)

I can't say how numerous problems with "random" JSON modules are, so maybe my estimation of the overall frustration of `findAndRegisterModules` was off. But wouldn't you still say that it would be more frustrating to intentionally add a JSON module that then gets ignored? To get a bit more input, [I asked on Twitter](https://twitter.com/nipafx/status/1526116179273732097).

Re breaking changes: Every change to the resolution strategy would be breaking, even the one currently proposed is, so I don't think that impacts which specific strategy is chosen.

robtimus:
The current vote count is in favour of selected modules, so I don't think we should continue with `findAndRegisterModules`.

Maybe I need to close this PR and create a new one, that gives control to the caller without giving much for free. The way to do this is to let the caller define their own `JsonConverter` class.

To be able to do this, I would make the following changes:
* Add annotation `@JsonConverter` (but differently named for obvious reasons, suggestions please! Maybe `@JsonConverterType`?) with one required `value` property of type `Class<? extends JsonConverter>` (the interface).
* Make `JsonConverter` and `Node` public. They're needed by the annotation.
* Make `JacksonJsonConverter` public (but not `JacksonNode`). It can either be made final, in which case callers would need to encapsulate it in their own converter implementations, or only its methods need to be made final so callers can extend it with a predefined `ObjectMapper`. I'd prefer to forbid callers to override the `toNode` methods.

Here's how this annotation would look like on a method:

```
@ParameterizedTest
@JsonClasspathSource(value = "luke.json", data = "vehicles")
@JsonConverter(MyCustomJacksonConverter.class)
```

---

To make life easier for callers, the `@JsonConverter` annotation could target both methods and classes (with `@Inherited`), with methods taking precedence of course. That way, the caller can annotate the entire class instead of every single method. The lookup in `JsonConverterProvider.getJsonConverter` would first check the test method (if present); then the test class (if present); then the current logic.

---

In addition, I'd change the current `JacksonJsonConverter` `INSTANCE` field to register the three modules that were discussed: Jdk8, JavaTime and Kotlin.

robtimus:
As the majority vote wants selected modules only, I'm refactoring this PR to only register the well-known modules: Jdk8, JavaTime and Kotlin. However, my lack of Gradle knowledge is acting up now (I'm a Maven user).

I don't think that we want to have all three modules loaded just because someone says `requireCapability("org.junit-pioneer:junit-pioneer-jackson")`. So how can we add the three modules separately? I now have this:
```
"jacksonImplementation"(group = "com.fasterxml.jackson.core", name = "jackson-databind", version = jacksonVersion)
"jacksonImplementation"(group = "com.fasterxml.jackson.datatype", name = "jackson-datatype-jdk8", version = jacksonVersion)
"jacksonImplementation"(group = "com.fasterxml.jackson.datatype", name = "jackson-datatype-jsr310", version = jacksonVersion)
"jacksonImplementation"(group = "com.fasterxml.jackson.module", name = "jackson-module-kotlin", version = jacksonVersion)
```
So how can we change this so callers have to use `requireCapability("org.junit-pioneer:junit-pioneer-jackson-jdk8")`, `requireCapability("org.junit-pioneer:junit-pioneer-jackson-jsr310")` or  `requireCapability("org.junit-pioneer:junit-pioneer-jackson-kotlin")`? Or do we actually _want_ Jdk8 and JavaTime to come "for free", seeing as Pioneer now requires Java 11+ anyway, and these two modules should always work? For Kotlin I'm adding a custom version of Spring's `KotlinDetector.isKotlinPresent()` before trying to add the module, so is it bad if the dependency is added when it's not used when Kotlin is not present?

And yes, there is a single Jackson version again - 2.13.3.

PR #631, opened 2022-05-13
Improve documentation of JSON argument source

nipafx:
Proposed commit message:

```
Improve documentation of JSON argument source (#621 / #631)

* mention that text blocks work
* reduce code snippet line length to max ~60, so they fit into the
  narrow code boxes on the website (i.e. no horizontal scroll bar)
* explain parser integration work:
	* so far only Jackson
	* optional dependency
	* force module resolution

Closes: #621
PR: #631
```
---
**PR checklist**

Documentation (general)

Documentation (new extension)

Code

Contributing

### DIFF
--- a/docs/json-argument-source.adoc
+++ b/docs/json-argument-source.adoc
@@ -150,9 +150,9 @@ This parametrized test will generate the following tests:
 To not make users' dependency management more complex, JUnit Pioneer has no run-time dependencies.
 At the same time, it's not parsing JSON itself and relies on third-party libraries for that.
 For this extension that means that projects who want to use it need to pull in a JSON parser themselves.
-This is the list of supported parsers (with the version used when building Pioneer although compatible versions should work as well):
+This is the list of supported parsers:
 
-* Jackson: com.fasterxml.jackson.core:jackson-databind:2.13.2.2
+* https://search.maven.org/artifact/com.fasterxml.jackson.core/jackson-databind[Jackson]
 
 If you need support for another parser, please https://github.com/junit-pioneer/junit-pioneer/issues/new/choose[open an issue].
 If your project does not already depend on a supported JSON parser, you can add it as follows.
@@ -173,7 +173,7 @@ testRuntimeOnly("org.junit-pioneer:junit-pioneer") {
 Alternatively, the dependency can be added directly:
 
 ```kotlin
-testImplementation("com.fasterxml.jackson.core:jackson-databind:$VERSION")
+testImplementation("com.fasterxml.jackson.core:jackson-databind:$CURRENT_VERSION")
 ```
 
 === Maven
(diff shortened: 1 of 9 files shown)


nipafx:
@filiphr You wanna take a look?

docs/json-argument-source.adoc
```diff
@@ -145,6 +145,59 @@ This parametrized test will generate the following tests:
 * [1] Snowspeeder, 4.5
 * [2] Imperial Speeder Bike, 3
 
+== JSON Parser Integration
+
+To not make users' dependency management more complex, JUnit Pioneer has no run-time dependencies.
+At the same time, it's not parsing JSON itself and relies on third-party libraries for that.
+For this extensions that means that projects who want to use it need to pull in a JSON parser themselves.
```
filiphr, line 152:
```suggestion
For this extension that means that projects who want to use it need to pull in a JSON parser themselves.
```

docs/json-argument-source.adoc
```diff
@@ -145,6 +145,59 @@ This parametrized test will generate the following tests:
...
+This is the list of supported parsers (with the version used when building Pioneer although compatible versions should work as well):
+
+* Jackson: com.fasterxml.jackson.core:jackson-databind:2.13.2.2
```
filiphr, line 155: I do not think that it is needed to add the exact specific version of Jackson for this. We are not using any specific features from Jackson for this. I would only mention that Jackson is used

filiphr (commented):
Looks good to me in general. I think that I found one small typo and I do not agree that a specific version of Jackson needs to be mentioned (in my opinion) it will confuse people.

Will leave it to you to decide what you want to do with the comments.

docs/json-argument-source.adoc
nipafx: I wasn't comfortable with that either, but I thought it outweighed the issue of users having to figure out the version for themselves. Because in a paragraph that tells you to add a Jackson dependency, wouldn't you expect it to tell you which version?

What if I change the sentence in 153 to:

> This is the list of supported parsers - with the version used when building Pioneer although most or even all versions should work because we're not using any of their features directly (you are) and just configure them:

filiphr: The Maven snippet uses only `$VERSION`, which means that users cannot copy paste it anyways (and I agree with the fact that they need to put a specific version). 

What if we change sentence in 153 to:

>This is the list of supported parsers:

and the line in 155 with a link to https://search.maven.org/artifact/com.fasterxml.jackson.core/jackson-databind? 

> * [Jackson](https://search.maven.org/artifact/com.fasterxml.jackson.core/jackson-databind)

nipafx: I didn't want to repeat the version to make updating it less error prone. I'm still not sure whether eliding it completely is such a good idea, but let's give it a try.

PR #632, opened 2022-05-13
Migrate to AsciiDoc snippets (#595 / #632)

### LINKED ISSUE #595: Migrate to AsciiDoc snippets
Once #334 gets merged, the rest of the documentation should be migrated to use the new approach.

Bukama:
Proposed commit message:

```
Migrate to AsciiDoc snippets (#595 / #632)

Migrate all demo code to AciiDoc snippets to let them be compiled.
Intentionally failing tests are moved to inner classes to prevent
test execution.

closes #595 
PR: #632
```

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -90,7 +90,7 @@ spotless {
 }
 
 checkstyle {
-	toolVersion = "7.8.2"
+	toolVersion = "10.2"
 	configDirectory.set(rootProject.file(".infra/checkstyle"))
 }
 
@@ -100,7 +100,7 @@ yamlValidator {
 }
 
 jacoco {
-	toolVersion = "0.8.7"
+	toolVersion = "0.8.8"
 }
 
 sonarqube {
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -260,6 +260,7 @@ tasks {
 				dependencies {
 					implementation(project)
 					implementation("com.fasterxml.jackson.core:jackson-databind:$jacksonVersion")
+					implementation("org.assertj:assertj-core:3.22.0")
 				}
 
 				sources {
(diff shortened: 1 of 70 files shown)


build.gradle.kts
```diff
@@ -260,6 +260,7 @@ tasks {
 				dependencies {
 					implementation(project)
 					implementation("com.fasterxml.jackson.core:jackson-databind:$jacksonVersion")
+					implementation("org.assertj:assertj-core:3.22.0")
```
Bukama, line 263: This seems not to work as `assertThat` is not found during compile (but by IntelliJ) in demo sources

nipafx:
Fixed it. Now the build fails because some of the demos fail at run time. Intentionally, I think, but that means those failing demos must be pulled into nested classes that aren't marked as `@Nested` and hence not executed by Jupiter. To be super-thorough, you could run them indirectly (like our tests do) and then assert the correct behavior. You could probably copy-paste that from the actual tests.

Bukama:
> Fixed it. Now the build fails because some of the demos fail at run time. Intentionally, I think, but that means those failing demos must be pulled into nested classes that aren't marked as `@Nested` and hence not executed by Jupiter. To be super-thorough, you could run them indirectly (like our tests do) and then assert the correct behavior. You could probably copy-paste that from the actual tests.

Thank you. I moved the failing tests to inner classes and the build successfully runs now (at least local).

One thing I noticed: As for the demo the same spotless/checkstyle rules are applied the code is intended. These intention is then also applied to the documentation. Same for all other formatting, which results in longer code lines than we had before after applying #317.

nipafx:
You need to add `indent=0` to the include's option list, e.g.:

```adoc
include::{json-demo}[tag=classpath_source,indent=0]
```

nipafx:
Thank you very much, Bish! ❤️

Issue #623, opened 2022-04-12
Execution order of extensions changed with 1.7.0


JoergSiebahn:
Since 1.7.0 some extension annotations at class level are not applied before `@RegisterExtension` any more.

Specifically `@SetSystemProperty` is using `beforeEach` now but used `beforeAll` and has been executed before `@RegisterExtension` annotated extension fields in the test class.

We are using [DropwizardAppExtension](https://github.com/dropwizard/dropwizard/blob/master/dropwizard-testing/src/main/java/io/dropwizard/testing/junit5/DropwizardAppExtension.java) to bootstrap an application for all tests in the test class. In our case, the application needs the system property on startup to be configured for the test.

This change has been introduced with #496.

We see [our test](https://github.com/SDA-SE/sda-dropwizard-commons/blob/2.48.6/sda-commons-server-dropwizard/src/test/java/org/sdase/commons/server/dropwizard/bundles/DefaultLoggingConfigurationBundleWithRequestLogFilterTest.java) failing with the [update to 1.7.0](https://github.com/SDA-SE/sda-dropwizard-commons/pull/1555).

The [PR](https://github.com/junit-pioneer/junit-pioneer/pull/496) and associated issues refer to alignment and consistent behavior. That sounds reasonable at a first glance. My personal assumption would be that something at class level is wrapped around something at method level and is applied as a class extension.

Do you have any suggestion how a test can be configured with 1.7.0 when it uses a static class extension that needs a system property to be configured for the test. In [our specific case ](https://github.com/SDA-SE/sda-dropwizard-commons/blob/2.48.6/sda-commons-server-dropwizard/src/test/java/org/sdase/commons/server/dropwizard/bundles/DefaultLoggingConfigurationBundleWithRequestLogFilterTest.java) we could try to bootstrap the application in the test method and shut it down in a `finally` statement. But in general that is not suitable as it should be started only once for many test methods for performance reasons.

beatngu13:
Hey @JoergSiebahn,

thank you for the detailed issue description! 🙏 [We actually considered this a bug](https://github.com/junit-pioneer/junit-pioneer/issues/479#issuecomment-1078361099), but your use case is definitely valid.

We will look into this and try to come up with something.

JoergSiebahn:
Hey @beatngu13,

[...]

PR #641, opened 2022-05-21
Add configuration option to entry-based extensions

Michael1993:
Related to #623 so we can move forward with it.

@JoergSiebahn could you share your opinion about this approach? Would this work for you?

Proposed commit message:

```

${references}: ${issues}
PR: ${pull-request}
```

---
**PR checklist**

Documentation (general)

Documentation (new extension)

Code

Contributing

### DIFF
--- a/src/main/java/org/junitpioneer/jupiter/ApplyMode.java
+++ b/src/main/java/org/junitpioneer/jupiter/ApplyMode.java
@@ -1,0 +1,15 @@
+/*
+ * Copyright 2016-2022 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junitpioneer.jupiter;
+
+public enum ApplyMode {
+	TEST, CLASS
+}
--- a/src/test/java/org/junitpioneer/jupiter/SystemPropertyExtensionTests.java
+++ b/src/test/java/org/junitpioneer/jupiter/SystemPropertyExtensionTests.java
@@ -236,7 +236,12 @@ class SystemPropertyExtensionTests {
 		@Nested
 		@SetSystemProperty(key = "A", value = "newer A")
 		@TestInstance(TestInstance.Lifecycle.PER_CLASS)
-		class ResettingSystemPropertyNestedTests {
+		class ResettingSystemPropertyAfterEachNestedTests {
+
+			@BeforeAll
+			void changeShouldNotBeVisible() {
+				assertThat(System.getProperty("A")).isEqualTo("old A");
+			}
 
 			@Test
 			@SetSystemProperty(key = "A", value = "newest A")
@@ -248,6 +253,30 @@ class SystemPropertyExtensionTests {
 			@ReadsSystemProperty
 			void resetAfterTestMethodExecution() {
 				assertThat(System.getProperty("A")).isEqualTo("old A");
+			}
+
+		}
+
+		@Nested
+		@SetSystemProperty(key = "A", value = "newer A", mode = ApplyMode.CLASS)
+		@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+		class ResettingSystemPropertyAfterAllNestedTests {
+
+			@BeforeAll
+			void changeShouldBeVisible() {
+				assertThat(System.getProperty("A")).isEqualTo("newer A");
+			}
+
+			@Test
+			@SetSystemProperty(key = "A", value = "newest A")
+			void setForTestMethod() {
+				assertThat(System.getProperty("A")).isEqualTo("newest A");
+			}
+
+			@AfterAll
+			@ReadsSystemProperty
+			void resetAfterTestMethodExecution() {
+				assertThat(System.getProperty("A")).isEqualTo("newer A");
 			}
 
 		}
--- a/src/main/java/org/junitpioneer/jupiter/AbstractEntryBasedExtension.java
+++ b/src/main/java/org/junitpioneer/jupiter/AbstractEntryBasedExtension.java
@@ -25,9 +25,12 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.function.Function;
+import java.util.function.Predicate;
 import java.util.stream.Stream;
 
+import org.junit.jupiter.api.extension.AfterAllCallback;
 import org.junit.jupiter.api.extension.AfterEachCallback;
+import org.junit.jupiter.api.extension.BeforeAllCallback;
 import org.junit.jupiter.api.extension.BeforeEachCallback;
 import org.junit.jupiter.api.extension.ExtensionConfigurationException;
 import org.junit.jupiter.api.extension.ExtensionContext;
@@ -46,7 +49,14 @@ import org.junitpioneer.internal.PioneerUtils;
  * @param <S> The set annotation type.
  */
 abstract class AbstractEntryBasedExtension<K, V, C extends Annotation, S extends Annotation>
-		implements BeforeEachCallback, AfterEachCallback {
+		implements BeforeEachCallback, AfterEachCallback, BeforeAllCallback, AfterAllCallback {
+
+	@Override
+	public void beforeAll(ExtensionContext context) {
+		List<ExtensionContext> contexts = PioneerUtils.findAllContexts(context);
+		Collections.reverse(contexts);
+		contexts.forEach(c -> clearAndSetEntries(c, ApplyMode.CLASS));
+	}
 
 	@Override
 	public void beforeEach(ExtensionContext context) {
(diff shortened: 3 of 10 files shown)


beatngu13:
Hmmm, I thought we want to avoid this (https://github.com/junit-pioneer/junit-pioneer/issues/623#issuecomment-1098911367):

>If at all possible, I'd like to avoid an option because it makes the user choose.

And have settled on (https://github.com/junit-pioneer/junit-pioneer/issues/623#issuecomment-1099404175):

>After today's maintainer meeting: we would first try to keep the current/new behavior, but make the extension additionally visible in `@BeforeAll`.

?

Michael1993:
I don't think it makes the user choose? By default it's always test-scoped with an optional configuration option to make it class scoped.

Michael1993:
Also I don't know what this means:

> but make the extension additionally visible in @BeforeAll.

How do you want to achieve that?

beatngu13:
>I don't think it makes the user choose? By default it's always test-scoped with an optional configuration option to make it class scoped.

"configuration option", read it again, slowly. 😬 It is very similar to the [first draft](https://github.com/junit-pioneer/junit-pioneer/issues/623#issuecomment-1098411733), to which @nipafx responded as above.

>Also I don't know what this means:
>
>>but make the extension additionally visible in @BeforeAll.
>
>How do you want to achieve that?

The [second draft](https://github.com/junit-pioneer/junit-pioneer/issues/623#issuecomment-1099404175) already works locally:

grafik

Diff (as I said, locally and early stage 😉):

```diff
--- a/src/main/java/org/junitpioneer/jupiter/AbstractEntryBasedExtension.java
+++ b/src/main/java/org/junitpioneer/jupiter/AbstractEntryBasedExtension.java
@@ -27,7 +27,9 @@ import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Stream;
 
+import org.junit.jupiter.api.extension.AfterAllCallback;
 import org.junit.jupiter.api.extension.AfterEachCallback;
+import org.junit.jupiter.api.extension.BeforeAllCallback;
 import org.junit.jupiter.api.extension.BeforeEachCallback;
 import org.junit.jupiter.api.extension.ExtensionConfigurationException;
 import org.junit.jupiter.api.extension.ExtensionContext;
@@ -46,7 +48,12 @@ import org.junitpioneer.internal.PioneerUtils;
  * @param <S> The set annotation type.
  */
 abstract class AbstractEntryBasedExtension<K, V, C extends Annotation, S extends Annotation>
-               implements BeforeEachCallback, AfterEachCallback {
+               implements BeforeEachCallback, AfterEachCallback, BeforeAllCallback, AfterAllCallback {
+
+       @Override
+       public void beforeAll(ExtensionContext context) {
+               beforeEach(context);
+       }
 
        @Override
        public void beforeEach(ExtensionContext context) {
@@ -138,7 +145,12 @@ abstract class AbstractEntryBasedExtension<K, V, C extends Annotation, S extends
        }
 
        @Override
-       public void afterEach(ExtensionContext context) throws Exception {
+       public void afterAll(ExtensionContext context) {
+               afterEach(context);
+       }
+
+       @Override
+       public void afterEach(ExtensionContext context) {
                // apply from innermost to outermost
                PioneerUtils.findAllContexts(context).forEach(this::restoreOriginalEntries);
        }
```

Some existing tests are failing due to backup/reset issues. This is the hard part, and I haven't had time so far to wrap my head around.

But personally I also like the configuration option, mainly because of two reasons:

1. It reminds me of [`TestInstance.Lifecycle`](https://junit.org/junit5/docs/5.0.1/api/org/junit/jupiter/api/TestInstance.Lifecycle.html).
2. The implementation is much, much easier.

So, I'm happy to take a step back and again talk about the design/approach (preferably not here in the PR, but in the issue and/or during a maintainer meeting).

Michael1993:
> "configuration option", read it again, slowly. 😬 It is very similar to the first draft, to which @nipafx responded as above.

I thought there is a difference between "making" the user choose and "letting" the user choose. I think this approach just adds more options to our users to configure the extensions. It's also very simple and straightforward (except if someone tries to set a method annotation to `mode = CLASS`).

src/main/java/org/junitpioneer/jupiter/ApplyMode.java
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright 2016-2022 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v20.html
+ */
+
```
Bukama, line 10: Javadoc :)

Michael1993: Javadoc and normal documentation is pending, I just wanted to check if this approach is OK, first.

Michael1993:
> > @JoergSiebahn could you share your opinion about this approach? Would this work for you?
> 
> Thanks for asking. For me the [second draft](https://github.com/junit-pioneer/junit-pioneer/issues/623#issuecomment-1099404175) that applies class level annotations in `beforeAll` and `beforeEach` feels more reasonable but also has a drawback:
> 
> * If I'm not missing something it is compatible to release 1.7.0 and (nearly compatible) to versions before 1.7.0. („nearly“ = see downside below)
> * My personal expectation is that a class level annotation is applied in `beforeAll`.
> * To have independent test cases, applying it in `beforeEach` again, is arguable.
> * The downside is that a wanted or needed dependency between test methods is hard to achieve as `test2` in the second draft shows. Imho, this downside exists in 1.7.0 as well. That may be mitigated by adding an optional config option for the class level annotation like `CLASS_LEVEL_ONLY`. I'm not sure how understandable such a setting is for the users.
> 
> However, I'm fine with the current approach based on the [first draft](https://github.com/junit-pioneer/junit-pioneer/issues/623#issuecomment-1098411733). It's not much that has to be changed in client code after upgrade.

Would using an array for `mode` work for you? I.e:
```java
@SetSystemProperty(key = "", value = "", mode = { ApplyMode.TEST, ApplyMode.CLASS })

**Comment by JoergSiebahn** [comment]
> Would using an array for mode work for you?

Based on the proposed changes, an array would give another option. To me it seems like it would cover an edge case where the property is required for the class and must be reset for a test because it is changed in another test. I'm not sure if the additional complexity is worth it. Assuming the scenario is an edge case, you could achieve the same with annotating the class and the test method. Maybe I'm missing some other scenarios.

When using `TEST` and `CLASS` as a default for class level annotations, you need different defaults based on the location (class or method). That may be hard to understand without looking in the code.

So these are just some thoughts from my point of view. In the end any solution discussed so far will solve our case.

**Comment by Michael1993** [comment]
> > Would using an array for mode work for you?
> 
> Based on the proposed changes, an array would give another option. To me it seems like it would cover an edge case where the property is required for the class and must be reset for a test because it is changed in another test. I'm not sure if the additional complexity is worth it. Assuming the scenario is an edge case, you could achieve the same with annotating the class and the test method. Maybe I'm missing some other scenarios.
> 
> When using `TEST` and `CLASS` as a default for class level annotations, you need different defaults based on the location (class or method). That may be hard to understand without looking in the code.
> 
> So these are just some thoughts from my point of view. In the end any solution discussed so far will solve our case.

Well, actually it doesn't make sense in general. Since the annotations are repeatable, you could just add one for `TEST` and one for `CLASS`.

**Comment by Michael1993** [comment]
Going back to your earlier comment @beatngu13 - should we actually ***use*** [`TestInstance.Lifecycle`](https://junit.org/junit5/docs/5.0.1/api/org/junit/jupiter/api/TestInstance.Lifecycle.html) (or, alternatively, its enum values)?

**Review by Bukama — APPROVED** [review]
I'm a bit curious if you really need the mode, as the unique id should be unique anyway, isn't it?
If you provide a good explaining commit message, you can merge :)

File: src/test/java/org/junitpioneer/jupiter/SystemPropertyExtensionTests.java [hunk 2]
```diff
@@ -236,7 +233,36 @@ class ResettingSystemPropertyTests {
 		@Nested
 		@SetSystemProperty(key = "A", value = "newer A")
 		@TestInstance(TestInstance.Lifecycle.PER_CLASS)
-		class ResettingSystemPropertyNestedTests {
+		class ResettingSystemPropertyAfterEachNestedTests {
```
beatngu13, line 236: What is the difference between this test class and `ResettingSystemPropertyAfterAllNestedTests`? (Same for `EnvironmentVariableExtensionTests.java`.)

src/test/java/org/junitpioneer/jupiter/SystemPropertyExtensionTests.java
```diff
@@ -247,15 +273,15 @@ void setForTestMethod() {
 			@AfterAll
 			@ReadsSystemProperty
 			void resetAfterTestMethodExecution() {
-				assertThat(System.getProperty("A")).isEqualTo("old A");
+				assertThat(System.getProperty("A")).isEqualTo("newer A");
```
beatngu13, line 276: Isn't this (and line 258) a breaking change? (Same for `EnvironmentVariableExtensionTests.java`.)

beatngu13 (changes requested):
I also did three commits, feel free to revert if you dislike them and/or I missed something.

src/test/java/org/junitpioneer/jupiter/SystemPropertyExtensionTests.java
Michael1993: Probably? 🤷 

I would like to pick this up again, but I completely forgot where we stand.

beatngu13: Finally a good topic to discuss during a maintainer meeting, I guess? 😅

nipafx:
[This conversation](https://github.com/junit-pioneer/junit-pioneer/pull/641#discussion_r901997436) left us with the open question of whether this PR can be merged before 2.0.

Con:

* it changes behavior and might force people to change their code

Pro:

* pushing it to 2.0 not only delays it into the future, it also requires people to upgrade to Java 11, which can be a big ask

Bukama:
After the documentation is updated (see @nipafx request above) I'm open to merge into merge for 1.x release to make it (also) available for Java 8 users. Yes behavior is changed, but for me bugfixes may change behavior which requires users to change there code. But we should not force user to upgrade to Java 11.

beatngu13 (approved):

JoergSiebahn:
Thanks a lot for the fix. 1.9.0 worked out of the box.

Issue #652, opened 2022-07-14
Include JUnit 5.9 in build


beatngu13:
JUnit 5.9.0-RC1 has been released:

https://junit.org/junit5/docs/snapshot/release-notes/#release-notes-5.9.0-RC1

So the final release is probably just around the corner.

Including 5.9.0-RC1 into our build:

https://github.com/junit-pioneer/junit-pioneer/pull/651/commits/12edd783323165fd6b514942a909a691bc2c2dca

Currently causes the following error:

```text
/home/runner/work/junit-pioneer/junit-pioneer/src/test/java/org/junitpioneer/internal/TestExtensionContext.java:33: error: TestExtensionContext is not abstract and does not override abstract method getExecutableInvoker() in ExtensionContext
public class TestExtensionContext implements ExtensionContext {
```

We should fix this and include JUnit 5.9 in our build.

Michael1993:
I guess we could implement it without the `@Override` annotation? Is that what you were thinking?

Bukama:
Final version of 5.9.0 was already released

Bukama:
I'll give a try in spare time

PR #651, opened 2022-07-14
Upgrade Gradle Wrapper and build against Java 18

beatngu13:
**EDIT:** Updated based on [Nicolai's comment](https://github.com/junit-pioneer/junit-pioneer/pull/651#issuecomment-1246523498).

~~Draft because [this](https://github.com/junit-pioneer/junit-pioneer/issues/613#issuecomment-1184802435) is still open for discussion.~~

Closes #613, #659.

Proposed commit message:

```
Upgrade Gradle, build against Java 18/19, improve modular tests (#613, #659 / #651)

(Yes, this change does too much, but it all hangs together.)

Updates the Gradle wrapper version from 7.4 to 7.5 because that comes
with support for Java 18, which is also added to the build pipeline.
The experimental Java version is now (for another few days), Java 19.

The modular build is properly configured to really test from the
module path.

Furthermore, replaces the archived/deprecated GitHub action
sormuras/download-jdk with oracle-actions/setup-java.

Closes: #613, #659
PR: #651
```

---
**PR checklist**

The following checklist shall help the PR's author, the reviewers and maintainers to ensure the quality of this project.
It is based on our contributors guidelines, especially the ["writing code" section](https://github.com/junit-pioneer/junit-pioneer/blob/main/CONTRIBUTING.md#writing-code).
It shall help to check for completion of the listed points.
If a point does not apply to the given PR's changes, the corresponding entry can be simply marked as done. 

Documentation (general)
* [ ] There is documentation (Javadoc and site documentation; added or updated)
* [ ] There is implementation information to describe _why_ a non-obvious source code / solution got implemented
* [ ] Site documentation has its own `.adoc` file in the `docs` folder, e.g. `docs/report-entries.adoc`
* [ ] Site documentation in `.adoc` file references demo in `src/demo/java` instead of containing code blocks as text
* [ ] Only one sentence per line (especially in `.adoc` files)
* [ ] Javadoc uses formal style, while sites documentation may use informal style

Documentation (new extension)
* [ ] The `docs/docs-nav.yml` navigation has an entry for the new extension
* [ ] The `package-info.java` contains information about the new extension

Code
* [ ] Code adheres to code style, naming conventions etc.
* [ ] Successful tests cover all changes
* [ ] There are checks which validate correct / false usage / configuration of a functionality and there are tests to verify those checks
[...]

PR #657, opened 2022-08-13
Update to JUnit 5.9 (#652 / 657)

Bukama:
Proposed commit message:

```
Update to JUnit Jupiter 5.9 (#652 / #657)

Updates the used JUnit Jupiter version

closes: #652
PR: #657
```

---
**PR checklist**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -147,7 +147,7 @@ jobs:
     timeout-minutes: 15
     strategy:
       matrix:
-        junit-version: [ '5.8.2', '5.9.0' ]
+        junit-version: [ '5.9.0' ]
         modular: [true, false]
         os: [ubuntu, macos, windows]
     name: Experimental build with newest JDK early-access build and Gradle release candidate
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -118,7 +118,7 @@ jobs:
     strategy:
       matrix:
         java: [ 11, 17 ]
-        junit-version: [ '5.8.2', '5.9.0' ]
+        junit-version: [ '5.9.0' ]
         modular: [true, false]
         os: [ubuntu, macos, windows]
     name: with Java ${{ matrix.java }}, JUnit ${{ matrix.junit-version }}, Modular ${{ matrix.modular }} on ${{ matrix.os }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -118,7 +118,7 @@ jobs:
     strategy:
       matrix:
         java: [ 11, 17 ]
-        junit-version: [ '5.8.2' ]
+        junit-version: [ '5.8.2', '5.9.0' ]
         modular: [true, false]
         os: [ubuntu, macos, windows]
     name: with Java ${{ matrix.java }}, JUnit ${{ matrix.junit-version }}, Modular ${{ matrix.modular }} on ${{ matrix.os }}
@@ -147,7 +147,7 @@ jobs:
     timeout-minutes: 15
     strategy:
       matrix:
-        junit-version: [ '5.8.2' ]
+        junit-version: [ '5.8.2', '5.9.0' ]
         modular: [true, false]
         os: [ubuntu, macos, windows]
     name: Experimental build with newest JDK early-access build and Gradle release candidate
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
-junitVersion=5.8.2
+junitVersion=5.9.0
 
 # Ensure sufficient heap size, especially for Sonar.
 org.gradle.jvmargs=-Xmx8g
--- a/src/test/java/org/junitpioneer/internal/TestExtensionContext.java
+++ b/src/test/java/org/junitpioneer/internal/TestExtensionContext.java
@@ -18,6 +18,7 @@ import java.util.Set;
 import java.util.function.Function;
 
 import org.junit.jupiter.api.TestInstance;
+import org.junit.jupiter.api.extension.ExecutableInvoker;
 import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.jupiter.api.extension.TestInstances;
 import org.junit.jupiter.api.parallel.ExecutionMode;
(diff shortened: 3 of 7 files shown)


gradle.properties
```diff
@@ -1,4 +1,4 @@
-junitVersion=5.8.2
+junitVersion=5.9.0
```
beatngu13, line 1: As you know, this actually bumps the JUnit version we use under hood. It not just incorporates it in our build (as suggested in #652). From our [contributing guidelines](https://github.com/junit-pioneer/junit-pioneer/blob/main/CONTRIBUTING.md#updating-junit-5):

>As [documented](https://github.com/junit-pioneer/junit-pioneer/blob/main/README.md#dependencies) Pioneer aims to use the lowest JUnit 5 version that supports Pioneer's feature set. At the same time, there is no general reason to hesitate with updating the dependency if a new feature requires a newer version or the old version has a severe bug.

Our code currently only fails with JUnit 5.9 due to test internals, not because we need a certain feature or fix. I'm cool with it, I just wonder if we can and/or want to stay on 5.8.2?

In any case, please also update the build accordingly:

* https://github.com/junit-pioneer/junit-pioneer/blob/main/.github/workflows/build.yml#L121
* https://github.com/junit-pioneer/junit-pioneer/blob/main/.github/workflows/build.yml#L150

beatngu13 (changes requested):

gradle.properties
Bukama: You are totally right (e.g. #320 was the issue about upgrading to 5.6.0). But I understood your comment in the issue (#652) as a "we need it" and therefore created this PR. I'll check if I can get the build work without updating. If not I would call a vote on next maintainer meeting about either upgrading the Jupiter version or not adding it to the build matrix to avoid failing builds.

P.S. Will update the two sections of the workflow script you've linked while doing this.

Bukama: Okay in 5.9.0 the interface returns `org.junit.jupiter.api.extension.ExecutableInvoker`, but in 5.8.2 the class is placed `org.junit.jupiter.engine.execution.ExecutableInvoker`. So even if I "overwrite" the `getExecutableInvoker` Methode it would return the false interface.

src/test/java/org/junitpioneer/internal/TestExtensionContext.java
```diff
@@ -42,11 +43,16 @@ public TestExtensionContext(Class<?> testClass, Method testMethod) {
 		this.testMethod = testMethod;
 	}
 
-	// @Override once we baseline against 5.8
+	@Override
 	public ExecutionMode getExecutionMode() {
 		throw NOT_SUPPORTED_IN_TEST_CONTEXT;
 	}
 
+	@Override
```
Michael1993, line 51:
```suggestion
	// @Override once we baseline against 5.9
```
We don't have to update, so we should not.

.github/workflows/build.yml
```diff
@@ -118,7 +118,7 @@ jobs:
     strategy:
       matrix:
         java: [ 11, 17 ]
-        junit-version: [ '5.8.2' ]
+        junit-version: [ '5.8.2', '5.9.0' ]
```
beatngu13, line 121: You can / have to remove 5.8.2, as our test code is now incompatible.

```suggestion
        junit-version: [ '5.9.0' ]
```

.github/workflows/build.yml
```diff
@@ -147,7 +147,7 @@ jobs:
     timeout-minutes: 15
     strategy:
       matrix:
-        junit-version: [ '5.8.2' ]
+        junit-version: [ '5.8.2', '5.9.0' ]
```
beatngu13, line 150:
```suggestion
        junit-version: [ '5.9.0' ]
```

beatngu13 (changes requested):

beatngu13 (approved):

Bukama:
> Abstract entry-based extension > should not mix backups of different extensions on clear environment variable and set system property FAILED
    org.junit.jupiter.api.extension.ExtensionConfigurationException: Cannot access Java runtime internals to modify environment variables. Have a look at the documentation for possible solutions: https://junit-pioneer.org/docs/environment-variables/#warnings-for-reflective-access
        at app//org.junitpioneer.jupiter.EnvironmentVariableUtils.getFieldValue(EnvironmentVariableUtils.java:104)
        at app//org.junitpioneer.jupiter.EnvironmentVariableUtils.setInProcessEnvironmentClass(EnvironmentVariableUtils.java:78)
        at app//org.junitpioneer.jupiter.EnvironmentVariableUtils.modifyEnvironmentVariables(EnvironmentVariableUtils.java:50)
        at app//org.junitpioneer.jupiter.EnvironmentVariableUtils.clear(EnvironmentVariableUtils.java:45)
        at app//org.junitpioneer.jupiter.AbstractEntryBasedExtensionTests.setUp(AbstractEntryBasedExtensionTests.java:36)

        Caused by:
        java.lang.reflect.InaccessibleObjectException: Unable to make field private static final java.util.HashMap java.lang.ProcessEnvironment.theEnvironment accessible: module java.base does not "opens java.lang" to unnamed module @59d016c9
            at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:354)
            at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:297)
            at java.base/java.lang.reflect.Field.checkCanSetAccessible(Field.java:180)
            at java.base/java.lang.reflect.Field.setAccessible(Field.java:174)
            at org.junitpioneer.jupiter.EnvironmentVariableUtils.getFieldValue(EnvironmentVariableU

Bukama:
@beatngu13  Do you have an idea how to implement the set environment thing without reflection so I can get rid of the error?

beatngu13:
These tests are already failing on `main`:

https://github.com/junit-pioneer/junit-pioneer/runs/7782024824?check_suite_focus=true#step:8:61

So either you ignore it, as you didn't change anything in this regard, or you disable the test on Java 17+ (preferred by me):

```java
@EnabledForJreRange(max = JRE.JAVA_16, disabledReason = "See: https://github.com/junit-pioneer/junit-pioneer/issues/509")
```

In #651 / #614, we try to (re)enable them with `--adds-open`, but we are still struggling with the Gradle setup.

Bukama:
Disabled `AbstractEntryBasedExtensionTests`. The two `EnvironmentVariable...Tests` were already disabled (what's strange as the `AbstractEntryBasedExtensionTests` also belongs to these env thing. Let's see.

Bukama:
Blocked by #651 (Gradle 7.5) due Java 18 is not supported with 7.4:
Gradle-issue https://github.com/gradle/gradle/issues/20344 is fixed with 7.5

beatngu13:
@Bukama not blocked, all builds passed (now). Feel free to merge. ✌️

Issue #658, opened 2022-08-17
Avoid use of Locale constructor in default locale extension


beatngu13:
The `Locale` constructors have been [deprecated](https://download.java.net/java/early_access/loom/docs/api/java.base/java/util/Locale.html#%3Cinit%3E(java.lang.String)) in Java 19. From the constructor implementation notes:

>* Obsolete ISO 639 codes ("iw", "ji", and "in") are mapped to their current forms. See [Legacy language codes](https://download.java.net/java/early_access/loom/docs/api/java.base/java/util/Locale.html#legacy_language_codes) for more information.
>* For backward compatibility reasons, this constructor does not make any syntactic checks on the input.

Several [alternatives](https://download.java.net/java/early_access/loom/docs/api/java.base/java/util/Locale.html#ObtainingLocale) exist to obtain a `Locale`, but, e.g., [`Locale.Builder`](https://download.java.net/java/early_access/loom/docs/api/java.base/java/util/Locale.Builder.html) does syntax checks, which leads to the following test failures:

```text
DefaultLocaleTests
  applied on the method level
    sets the default locale using a language, a country and a variant
      java.util.IllformedLocaleException: Ill-formed variant: gb [at index 0]
    sets the default locale using a language
      java.util.IllformedLocaleException: Ill-formed language: en_EN [at index 0]
```

We should avoid the use of the deprecated `Locale` constructors and switch to one of the aforementioned alternatives. Note that this is a _breaking change_, as we change the behavior of the default locale extension.

Bukama:
Locale is the same for 8 and 11 with
* Constructors (with language/country/variant)
* Factory-method "forLanguage(language-String), which takes IETF BCP 47 language strings, but does not makes syntax checks on the strings (as far as I understand the docs)
* Constants, e.g. Locale.US
* Builder which allows setting language/country/variant but with checking the " syntax requirements defined by the Locale class"

In 19 (didn't check version between) there's also a factory method "of" for language/country/variant. But the constructors are removed (as @beatngu13  already mentioned).

So question is now: which way do we use?
I personally like constants most as I don't have to care about the string values. But I think for our test cases the builder way makes more sense - meaning we have to change some cases to match valid IETF BCP 47 strings.

beatngu13:
Closed via #662.

PR #662, opened 2022-08-30
Replace Locale constructor by Locale builder (#658 / #662)

Bukama:
Proposed commit message:

```
Replace Locale constructor by Locale builder (#658 / #662)

The Locale constructor gets deprecated in newer Java version.
Therefore we replace it by the Locale builder.

closed: #658
PR: #662
```
linked issue: #658

### DIFF
--- a/docs/default-locale-timezone.adoc
+++ b/docs/default-locale-timezone.adoc
@@ -16,11 +16,13 @@ The default `Locale` can be specified using an https://docs.oracle.com/javase/8/
 include::{demo}[tag=default_locale_language]
 ----
 
-Alternatively the default `Locale` can be specified according to https://docs.oracle.com/javase/8/docs/api/java/util/Locale.html#constructor.summary[its constructors] - it needs either:
+Alternatively the default `Locale` can be created using the following attributes of which a https://docs.oracle.com/javase/8/docs/api/java/util/Locale.Builder.html[Locale Builder] can create an instance with:
 
 * `language` or
 * `language` and `country` or
 * `language`, `country`, and `variant`
+
+Note:: The variant needs to be a string which follows the https://www.rfc-editor.org/rfc/rfc5646.html[IETF BCP 47 / RFC 5646] syntax!
 
 [source,java,indent=0]
 ----
@@ -28,7 +30,7 @@ include::{demo}[tag=default_locale_language_alternatives]
 ----
 
 Note that mixing language tag configuration and constructor based configuration will cause an `ExtensionConfigurationException` to be thrown.
-Furthermore a `variant` can only be specified if `country` is also specified.
+Furthermore, a `variant` can only be specified if `country` is also specified.
 If `variant` is specified without `country`, an `ExtensionConfigurationException` will be thrown.
 
 Any method level `@DefaultLocale` configurations will override class level configurations.
--- a/docs/default-locale-timezone.adoc
+++ b/docs/default-locale-timezone.adoc
@@ -22,7 +22,7 @@ Alternatively the default `Locale` can be created using the following attributes
 * `language` and `country` or
 * `language`, `country`, and `variant`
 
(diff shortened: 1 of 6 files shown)


Bukama:
Made a new branch/PR with clean base
Difference from last
* Remove PioneerUtils from Democode
* Added PioneerUtils to extension class
* Added test with invalid BCP 47 variant

Michael1993 (approved):
Please fix the build 😇

Bukama:
Adding merge-ready for full build

Michael1993 (changes requested):
I think the documentation might need an update to reflect these changes.

Bukama:
> I think the documentation might need an update to reflect these changes.

What do you miss? The examples in `DefaultLocaleTimezoneExtensionDemo.java` are updated and in the other parts of the documentation I don't see the need for an update.

beatngu13:
@Bukama Michael is right, I have overlooked this, too. For example:

https://github.com/junit-pioneer/junit-pioneer/blob/8e9bfe345ace6c057b6fecc5059004d59b492fd4/src/main/java/org/junitpioneer/jupiter/DefaultLocale.java#L25-L33

Here we refer to the deprecated constructors, we no longer use under the hood. It can be seen as an implementation detail, but I believe it is important. Maybe refer to [`Locale.Builder`](https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/util/Locale.Builder.html)?

>`Builder` is used to build instances of `Locale` from values configured by the setters. Unlike the `Locale` constructors, the `Builder` checks if a value configured by a setter satisfies the syntax requirements defined by the `Locale` class. A `Locale` object created by a `Builder` is well-formed and can be transformed to a well-formed IETF BCP 47 language tag without losing information.

---

We should also note that once this is merged, `main` is essentially on 2.0.0-SNAPSHOT. We can longer create a 1.x.y (hotfix) release, at least not with our current pipelines.

Bukama:
Ah thanks for the hint, with documentation I checked the documentation subpage :D

Bukama:
Docs updated.

> We should also note that once this is merged, main is essentially on 2.0.0-SNAPSHOT. We can longer create a 1.x.y (hotfix) release, at least not with our current pipelines.

This is not a thing of this PR. The changes in this PR are totally fine with the 1.x versions as it's still Java 8. The fact that the extension behaves a little bit different is no argument for a major change. In my eyes it's more a bugfix, as invalid variants are not allowed anymore.

If the want/need a way to create a 1.x release after we have set up `main` for 2.x one has to create such a Github workflow. But again, this is not part of this issue/PR.

beatngu13:
> This is not a thing of this PR. The changes in this PR are totally fine with the 1.x versions as it's still Java 8. The fact that the extension behaves a little bit different is no argument for a major change. In my eyes it's more a bugfix, as invalid variants are not allowed anymore.

We went down this path with https://github.com/junit-pioneer/junit-pioneer/issues/479#issuecomment-1078361099, which gave us #623. 😉 Not sure if we can or want to "sell" this change as a bugfix. Just as we did in our own tests, users may create `Locale`s that are invalid IETF BCP 47 strings. This would be a breaking change for them.

> If the want/need a way to create a 1.x release after we have set up `main` for 2.x one has to create such a Github workflow. But again, this is not part of this issue/PR.

Alternatively, we could also start integrating 2.x changes into a dedicated branch. But yeah, here is indeed the wrong place to discuss this.

Bukama:
Well it's the same discussion in the linked issues/PR: Is a bug fix, which does not accept past inputs a breaking change and therefor a major release? For me - absolutely not. The only thing I can get in mind is that you rewrite a whole functionality because you can't fix the bug, but then I would not call that a bugfix, but a rewrite of the functionality where the bug was only the point where the stone got rolling.

So what about a mid was? We meet inbetween and call this change worth a minor release?

nipafx:
If be7295d gets merged in #651, it needs to be reverted in this PR.

beatngu13 (approved):

PR #664, opened 2022-09-01
Fix JSON sources to convert to Lists and other types

### LINKED ISSUE #654: Convert ArrayNode to List
I'd like to convert a json array into a java List, but I got this error:

```
Error converting parameter at index 0: No implicit conversion to convert object of type com.fasterxml.jackson.databind.node.ArrayNode to type java.util.List
org.junit.jupiter.api.extension.ParameterResolutionException: Error converting parameter at index 0: No implicit conversion to convert object of type com.fasterxml.jackson.databind.node.ArrayNode to type java.util.List

```
Here my test:
```
@ParameterizedTest
@JsonClasspathSource("items.json")
void test1(@Property("items") List<Integer> items) {
    assertFalse(items.isEmpty());
    assertEquals(2, items.size());
}

@ParameterizedTest
@JsonClasspathSource("items.json")
void test2(@Property("items") Integer[] items) {
    assertEquals(2, items.length);
}
```

items.json:

```
[
  {
    "items": [ 1, 3 ]
  },
  {
    "items": [ 2, 4 ]
  }
]
```

what am I missing?

Michael1993:
Closes #654 

Proposed commit message:

```
Fix JSON sources to convert to Java types (#654 / #664)

The initial implementation of the JSON sources for ParameterizedTest
was missing the functionality to map from JSON to more complex Java
types like List. This commit fixes that by using Jackson's
ObjectMapper and JavaType.

Closes: #654 
PR: #664
```

---
**PR checklist**

Documentation (general)

Documentation (new extension)

Code

Contributing


filiphr (commented):
Sorry for intruding, but since I was working on the Json support I wanted to chime in.

I believe that this approach is a bit strict and it has a dependency on Jackson. 

---

Jackson has support for converting to a generic type with the help of its `JavaType`.  This `JavaType` can be constructed from a `java.lang.reflect.Type`.

This means that new methods (or the existing methods) of the `Node` can look like:

```java
<T> T toType(Type type);
Object value(Type typeHint);
```

Everything in the end delegates to `toType(Type)` which is implemented in the following way:

```java
@Override
public <T> T toType(Type type) {
	try {
		return objectMapper.treeToValue(node, objectMapper.constructType(type));
	}
	catch (JsonProcessingException e) {
		throw new UncheckedIOException("Failed to convert to type " + type, e);
	}
}
```

As a final step all the places that were invoking the `toType` and `value` methods can be changed from doing `Parameter#getType` to `Parameter#getParameterizedType`.

There won't be a need to have a custom argument converter and it would be possible to map into things like `Map<String, Poem>` or `List<Map<String, Poem>>` etc.

src/main/java/org/junitpioneer/jupiter/json/AbstractJsonArgumentsProvider.java
```diff
@@ -55,15 +55,14 @@ private static Object createArgumentForCartesianProvider(Parameter parameter, No
...
-		if (singleParameter) {
+		if (method.getParameterCount() == 1) {
 			Parameter onlyParameter = method.getParameters()[0];
 			// When there is a single parameter, the user might want to extract a single value or an entire type.
 			// When the parameter has the `@Property` annotation, then a single value needs to be extracted.
 			Property property = onlyParameter.getAnnotation(Property.class);
 			if (property == null) {
 				// no property specified -> the node should be converted in the parameter type
-				return Arguments.arguments(node.toType(onlyParameter.getType()));
+				return () -> new Object[] { node.toType(onlyParameter.getParameterizedType()) };
```
filiphr, line 65: What is wrong with using `Arguments.arguments`?

src/test/java/org/junitpioneer/jupiter/json/ArrayNodeToListTests.java
```diff
@@ -0,0 +1,216 @@
...
+			then(poems).extracting(Poem::getRelease).noneMatch(i -> i < 1900);
+		} else {
+			then(poems).extracting(Poem::getRelease).noneMatch(i -> i > 1900);
+		}
+	}
+
+	@ParameterizedTest
+	@ValueSource(strings = { "12", "23" })
+	@ReportEntry("{0}")
+	void doesNotConvertNonNodesSimple(@Property("nope") String i) {
```
filiphr, line 161: Is this test needed? It only tests `@ValueSource` and I don't think that this was the intention

src/test/java/org/junitpioneer/jupiter/json/ArrayNodeToListTests.java
```diff
@@ -0,0 +1,216 @@
...
+		then(i).isNotNull();
+	}
+
+	@ParameterizedTest
+	@MethodSource("letters")
+	@ReportEntry("{0}")
+	void doesNotConvertNonNodesList(@Property("nope") List<String> letters) {
```
filiphr, line 168: Is this test needed? It only tests `@ValueSource` and I don't think that this was the intention

src/test/java/org/junitpioneer/jupiter/json/ArrayNodeToListTests.java
```diff
@@ -0,0 +1,216 @@
...
+
+	public static Stream<Arguments> letters() {
+		return Stream.of(arguments(Arrays.asList("a", "b", "c")));
+	}
+
+	static class BadConfigurationTestCase {
+
+		@ParameterizedTest
+		@JsonClasspathSource(BAD_POEMS)
+		void conversionException(@Property("poems") List<Poem> poems) {
```
filiphr, line 180: Why does this test case not work? The only difference that I can spot between `bad_poems.json` and `poets.json` is the name of the property `poems` vs `works`. However, this shouldn't be causing a problem

filiphr (commented):
The implementation looks as I would expect now @Michael1993.

I've left some comments about some more general confusions I have.

One thing that you should decide within the team is the breaking change in `Node.java`. From my point of view it is fine to do that change, since this is a new interface and I really doubt that someone has implemented an entire new JSON parsing on top of JUnit Pioneer without contributing it back.

src/main/java/org/junitpioneer/jupiter/json/AbstractJsonArgumentsProvider.java
scordio: This reminded me of assertj/assertj@6cab52be094ece12818cc41f82028ccf833af201 and junit-team/junit5#2708, not sure if they are related to this one.

Michael1993: It's a weird generics issue as far as I understood - we call a method that returns a `T`, pass it to a method with varargs and it gets interpreted as an array. But we return a single object, not an array and it throws a `ClassCastException`.

src/test/java/org/junitpioneer/jupiter/json/ArrayNodeToListTests.java
Michael1993: `release` in one case has "abcd" which is not a number.

Michael1993:
> One thing that you should decide within the team is the breaking change in `Node.java`.
This PR contains no breaking changes as all classes/interfaces modified are package-private.

src/test/java/org/junitpioneer/jupiter/json/ArrayNodeToListTests.java
Michael1993: Not needed. Deleted. Thanks!

src/test/java/org/junitpioneer/jupiter/json/ArrayNodeToListTests.java
Michael1993: Not needed. Deleted. Thanks!

src/main/java/org/junitpioneer/jupiter/json/AbstractJsonArgumentsProvider.java
Michael1993: It's fairly trivial to reproduce the problem:
```java
public static void main(String[] args) {
	methodWithVarargs(getGenericThing());
}

public static void methodWithVarargs(Object... arguments) {
	System.out.println(Arrays.toString(arguments));
}

public static <T> T getGenericThing() {
	return (T) new Object();
}
```

This throws:
```
Exception in thread "main" java.lang.ClassCastException: class java.lang.Object cannot be cast to class [Ljava.lang.Object; (java.lang.Object and [Ljava.lang.Object; are in module java.base of loader 'bootstrap')
```

Michael1993: I will add a comment explaining it to the code, to avoid future confusion.

filiphr:
>This PR contains no breaking changes as all classes/interfaces modified are package-private.

I didn't notice that. They used to be public when I wrote them initially

src/main/java/org/junitpioneer/jupiter/json/AbstractJsonArgumentsProvider.java
filiphr: Got it. If you cat to `Object` you should not have the problem. In any case, thanks for the explanation

src/test/java/org/junitpioneer/jupiter/json/ArrayNodeToListTests.java
filiphr: Got it, didn't saw that. Thanks

src/main/java/org/junitpioneer/jupiter/json/AbstractJsonArgumentsProvider.java
Michael1993: I'm actually not sure about this line - would it look nicer with an `(Object)` cast?

Bukama (approved):
As far as I understand all those reflection things it looks good to me, but I mostly trust @filiphr. When I have a look at the documentation I think there's no need in updating it. But please have a second look at it @Michael1993

Bukama:
Adding merge-ready to trigger full build

beatngu13 (approved):
Thanks a lot Michael for working on this!

I only have two "take it or leave it" comments. In addition, do you think it make sense to add an example to the docs/demos?

PR #666, opened 2022-09-14
Improve Cartesian test documentation

nipafx:
Proposed commit message:

```
Improve Cartesian test documentation (#665 / #666)

`@CartesianTest.MethodFactory`-annotated methods can be non-static
under certain circumstances, but the demo snippets showed non-static
factory methods that didn't match these and were thus wrong. This
change fixes that, which required moving the factory methods from
non-static inner classes into the outer class. For conistency and to
make things easier in the future, the same was done for all other
snippets as well.

The snippets were prepended with imports, which was incoherent when
the snippets turned from classes to just methods. Hence the imports
were removed as well, which has the eadded benefit of not having to
maintain the imports and making the snippets shorter.

Closes: #665
PR: #666
```

---
**PR checklist**

Documentation (general)

Documentation (new extension)

Code

Contributing


nipafx:
Funny story: The failing demo demonstrated something, which I think we just don't do. 😳

> The `value` attribute is required in the following example because the method parameter is declared as `TemporalUnit`, i.e. the interface implemented by `ChronoUnit`, which isn't an enum type.
>
> ```java
> @CartesianTest
> void testExplicitEnum(@Enum(ChronoUnit.class) TemporalUnit unit) { ... }
```

But that fails because of the check whether the parameter type is actually an enum:

```java
class CartesianEnumArgumentsProvider<E extends Enum<E>> implements CartesianParameterArgumentsProvider<E> {

	@Override
	public Stream<E> provideArguments(ExtensionContext context, Parameter parameter) {
		Class<?> parameterType = parameter.getType();
		if (!Enum.class.isAssignableFrom(parameterType))
			throw new PreconditionViolationException(
				String.format("Parameter of type %s must reference an Enum type", parameterType));
		// ...
	}
```

What's up with that?

docs/cartesian-product.adoc
```diff
@@ -27,23 +27,21 @@ NOTE: The CartesianTest extension has undergone significant changes in 1.6.0.
...
-== Basic Use
+== Basic use
 
 `@CartesianTest` is used _instead_ of `@Test` or other such annotations (e.g. `@RepeatedTest`).
 
 You can supply test parameters to `@CartesianTest` in two ways:
 
- - You can annotate your test method, providing all parameter values in a single annotation.
- - The test parameters can be annotated with `@CartesianTest.Values`, `@CartesianTest.Enum`, or range source annotations (see <<Annotating your test parameters>>)
+ - The test parameters can be annotated directly with `@CartesianTest.Values`, `@CartesianTest.Enum`, or range source annotations (see <<Defining arguments on parameters>>)
```
Michael1993, line 36: I thought we wanted to use active voice, wherever possible?

Michael1993:
> Funny story: The failing demo demonstrated something, which I think we just don't do. 😳
> 
> > The `value` attribute is required in the following example because the method parameter is declared as `TemporalUnit`, i.e. the interface implemented by `ChronoUnit`, which isn't an enum type.
> > ```java
> > @CartesianTest
> > void testExplicitEnum(@Enum(ChronoUnit.class) TemporalUnit unit) { ... }
> > ```
> 
> But that fails because of the check whether the parameter type is actually an enum:
> 
> ```java
> class CartesianEnumArgumentsProvider<E extends Enum<E>> implements CartesianParameterArgumentsProvider<E> {
> 
> 	@Override
> 	public Stream<E> provideArguments(ExtensionContext context, Parameter parameter) {
> 		Class<?> parameterType = parameter.getType();
> 		if (!Enum.class.isAssignableFrom(parameterType))
> 			throw new PreconditionViolationException(
> 				String.format("Parameter of type %s must reference an Enum type", parameterType));
> 		// ...
> 	}
> ```
> 
> What's up with that?

Happens too early - that check is only supposed to happen if the annotation has no parameters.

nipafx:
Do you have time to fix this? For simplicity's sake, we can do it as part of this PR.

docs/cartesian-product.adoc
nipafx: Good point! Fixed in d3cb494.

Michael1993:
> In Discord (❓) @Michael1993 pointed out:
> 
> image
> 
> > The documentation on the website is missing this part:
> > (unless the test class is annotated with `@TestInstance(Lifecycle.PER_CLASS)` and the factory method is defined in the test class)
> 
> That's true for [the documentation of the old Cartesian test extension](https://junit-pioneer.org/docs/cartesian-product-v1/#conditions-for-the-static-factory-method). Is the addition correct there as well?

No, in the old version it always had to be static, then someone suggested that it could be non-static when we were doing the new version.

nipafx:
Ok, then the documentation is correct as proposed in this PR. ✅

Michael1993 (approved):

Issue #551, opened 2021-12-04
Provide annotation that a test currently fails by given circumstances


Bukama:
Creating an issue as base for the PR for a `@NotWorking` annotation. Text comes from the PR by @Marcono1234

--- 

Often tests fail due to a bug in the tested application or in used dependencies.
Traditionally such a test method would be annotated with JUnit's `@Disabled`.
However, this has the following disadvantages when the bug causing the test failure
is fixed:

* the developer might not notice the existing test method and create a new one
* the existing test method might not be noticed and remains disabled for a long
  time after the bug has been fixed, adding no value for the project
`@NotWorking` tries to solve these issues. Unlike `@Disabled` it still executes the
annotated test method but aborts the test if a test failure or error occurs.
However, if the test is executed successfully the `@NotWorking` annotation will cause
a test failure because the test _is working_.
This lets the developer know that they have fixed a bug (possibly by accident) and that
they can now remove the `@NotWorking` annotation from the test method.

It is therefore recommended to prefer using `@NotWorking` over `@Disabled`
in most cases; exceptions for this are for example 'flaky' tests.

*Important*: This annotation is *not* intended as a way to mark test methods
which intentionally cause exceptions.
Such test methods should use JUnit's `assertThrows` or similar means to explicitly
test for a specific exception class being thrown by a specific action.

nipafx:
I really like this idea, it just makes sense to have the option to bake the assumption that the test fails into the enable/disable logic. Straight-up winner on that front.

But I'm wondering, whether it isn't still too close to `@Disabled`. All the requirements that exist for (disable on OS, on JDK version, until a date, for a specific parameter, etc.) could apply here as well and reimplementing all of that functionality for a new variant of that enable/disable behavior seems so wasteful. The question is whether there's a way around that.

Is there a way to implement this as a kind of "modifier" on `@Dsiabled`, e.g.:

```java
@NotWorking
@DisabledOnOs(MAC)
void test() {
	// ...
}
```

Then `@NotWorking` would have to detect `@DisabledOnOs`'s decision and, if it wants to disable, override that and wrap it into the behavior described above. It's very unclear to me whether that can work given that some of the `@Disabled...` annotations come from Jupiter Proper.

One weird trick that could work would be:

* copy all Jupiter `@Disabled...` annotations and extensions to Pioneer (maybe under a different class name?)
* edit them and all of our annotations to interact with `@NotWorking` as described above

[...]

PR #676, opened 2022-10-02
Add ExpectedToFail test extension (#551 / #676)

Marcono1234:
Proposed commit message:

```
Add ExpectedToFail test extension (#551 / #676)

PR: #676
```

Supersedes #546 (based on the same commits of that PR):
- I did not override `InvocationInterceptor.interceptTestFactoryMethod` as originally proposed in https://github.com/junit-pioneer/junit-pioneer/pull/546#discussion_r775149283 because that appears to be only the call to the factory method but not the individual test executions (?) so covering that probably does not make much sense.
- As already mentioned in the comments of #546, handling for exceptions which occur in `@AfterEach` could not be kept with the `InvocationInterceptor` approach.
- Due to using `InvocationInterceptor` this probably only supports `@Test` methods now, but not any dynamic tests (have not investigated that in detail though).

---
**PR checklist**

Documentation (general)

Documentation (new extension)

Code

Contributing


src/main/java/org/junitpioneer/jupiter/ExpectedToFailExtension.java
```diff
@@ -0,0 +1,82 @@
...
+	 * <p>This method is used for exceptions which abort test execution and should
+	 * have higher precedence than aborted exceptions thrown by this extension.
+	 */
+	private static boolean shouldPreserveException(Throwable t) {
+		// Note: Ideally would use the same logic JUnit uses to determine if exception is aborting
+		// execution, see its class OpenTest4JAndJUnit4AwareThrowableCollector
+		return TestAbortedException.class.isInstance(t);
+	}
+
+	private static <T> T invokeAndInvertResult(Invocation<T> invocation, ExtensionContext extensionContext)
```
Marcono1234, line 52: Have implemented this as separate method in case more `InvocationInterceptor` methods will be overridden in the future.

src/test/java/org/junitpioneer/testkit/assertion/PioneerAssert.java
```diff
@@ -88,86 +88,110 @@ public void hasNoReportEntries() {
...
-		assertSingleTest(actual.testEvents().skipped().count());
+		assertSingleTest(actual.testEvents().skipped());
 	}
 
 	@Override
 	public TestCaseStartedAssert hasSingleDynamicallyRegisteredTest() {
-		return assertSingleTest(actual.testEvents().dynamicallyRegistered().count());
+		Events testEvents = actual.testEvents();
+		assertSingleTest(testEvents.dynamicallyRegistered().count());
+		// Retain all events for returned Assert because TestCaseStartedAssert allows further filtering
```
Michael1993, line 122: The double filtering was intentional. Could you provide your rationale for removing it?

Marcono1234: The comment was actually a bit misleading; I have now adjusted it.

The issue was that previously (in the current version of this project) the `TestCaseAssertBase` was always created with all test / container events, without being restricted to the type of the `hasSingle...()` method:
https://github.com/junit-pioneer/junit-pioneer/blob/f36ee4c351d043e59878fe829303848f5a85e67c/src/test/java/org/junitpioneer/testkit/assertion/PioneerAssert.java#L119-L122

That happened to work because `TestCaseAssertBase.throwable()` previously always restricted the events to `failed()`, however this caused issues for this extension when I implemented `TestCaseAbortedAssert`.

Therefore most `hasSingle...()` now properly filter the events, but for the methods where this is not possible (and which therefore behave the same way they did before) I added those comments.

nipafx:
Thanks for your work on this, @Marcono1234. I made some small changes (see 9731701, 11e03c2) and I think it's ready to be merged and released in version 1.8.0. 👍🏾

nipafx:
Proposed commit message:

```
Add `@ExpectedToFail` test extension (#551 / #676)

As opposed to JUnit Jupiter's `@Disabled` annotation, this extension
executes the annotated test and then effectively flips the result:

* failed ~> ignored
* passed ~> failed

This ensures that a temporarily disabled test will get discovered
once the reason for being disabled no longer holds, so it can/must
be re-enabled by removing `@ExpectedToFail`.

PR: #676
```

nipafx:
@Marcono1234 Forgot to write: Can you let us know whether the changes I made are ok? I'll wait with the merge until then.

docs/expected-to-fail-tests.adoc
```diff
@@ -3,16 +3,16 @@
...
+However, this has disadvantages when the bug that causes the test failure is fixed:
 
 * the developer might not notice the existing test method and create a new one
 * the existing test method might not be noticed and remains disabled for a long time after the bug has been fixed, adding no value for the project
 
-`@ExpectedToFail` tries to solve these issues.
+`@ExpectedToFail` solves these issues.
 Unlike `@Disabled` it still executes the annotated test method but aborts the test if a test failure or error occurs.
-However, if the test is executed successfully the `@ExpectedToFail` annotation will cause a test failure because the test _is working_.
+However, if the test is executed successfully the it will cause a test failure because the test _is working_.
```
Marcono1234, line 15:
```suggestion
However, if the test is executed successfully it will cause a test failure because the test _is working_.
```

Marcono1234 (commented):
@nipafx, thanks, the changes look fine to me! There only seems to be one small typo.

Feel free to merge this afterwards.

nipafx:
Thank you, @Marcono1234! 👍🏾

nipafx:
Thanks again, for your high-quality work and patience. I just released 1.8.0 with [your extensions](https://junit-pioneer.org/docs/expected-to-fail-tests/). 🙌🏾

PR #698, opened 2022-11-28
Fix StdIo and ReportEntry to target ANNOTATION_TYPE

### LINKED ISSUE #696: Add support for composed annotations to all annotations
Looking through the source code of the annotations used to register extensions, it looks like none of them (or none that I could find) support [composed annotations](https://junit.org/junit5/docs/current/user-guide/#writing-tests-meta-annotations), that is, reusing the annotation in another annotation.

JUnit's user manual has a few examples of this in the [declarative extension registration section](https://junit.org/junit5/docs/current/user-guide/#extensions-registration-declarative).  It would allow users who repeatedly use several extensions together to bundle those into a single annotation like this:
```java
@Retention(RetentionPolicy.RUNTIME)
@Target({ TYPE, METHOD, ANNOTATION_TYPE })
@Stopwatch  @DefaultLocal("Local.UK") @SetSystemProperty(key="BROILER" value="ON")
public @interface StandardTestSetup { }
```

To enable that, two things would be needed:
1. The `ANNOTATION_TYPE` would need to be added to the `@Target` of the annotations
2. The `PioneerAnnotationUtils` would need to be able to find them.

Currently the PioneerAnnotationUtils class only looks for the specified annotation - It doesn't check to see what meta-annotations an annotation might have.  I think that happens (or doesn't happen) [right here](https://github.com/junit-pioneer/junit-pioneer/blob/1330e8bd50a5088c1ebe833967d045fcd2d9ee74/src/main/java/org/junitpioneer/internal/PioneerAnnotationUtils.java#L210).

Michael1993:
Proposed commit message:

```
 Fix StdIo and ReportEntry not targeting ANNOTATION_TYPE (#696 / #698)

This commit fixes the annotations @ReportEntry and @StdIo missing
the ANNOTATION_TYPE or TYPE from their @Target. They can now be used
on composed annotations.

Closes: #696
PR: #698
```

Closes #696

### DIFF
--- a/src/main/java/org/junitpioneer/jupiter/ReportEntry.java
+++ b/src/main/java/org/junitpioneer/jupiter/ReportEntry.java
@@ -39,7 +39,7 @@ import org.junit.jupiter.api.extension.ExtendWith;
  * @since 0.5.6
  */
 @Repeatable(ReportEntry.ReportEntries.class)
-@Target(ElementType.METHOD)
+@Target({ ElementType.METHOD, ElementType.ANNOTATION_TYPE })
 @Retention(RetentionPolicy.RUNTIME)
 @ExtendWith(ReportEntryExtension.class)
 public @interface ReportEntry {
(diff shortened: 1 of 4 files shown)


nipafx:
Good work @eeverman for finding this and reporting and @Michael1993 for fixing! 👍🏾

PR #699, opened 2022-12-07
WIP Issue/574 sys props restore

### LINKED ISSUE #574: Add extension / annotation for restoring system properties
Based on https://github.com/junit-pioneer/junit-pioneer/issues/381#issuecomment-1001221833

In some situations the existing annotations for clearing and setting system properties are not sufficient:
- When the property name is dynamic
- When the property value is dynamic and the currently described workaround of using `@ClearSystemProperty` is considered too error-prone (especially when setting multiple system properties)
- When system properties are set outside the test method, e.g. inside the tested code

It would therefore be useful to have an annotation which captures the current system properties before the test execution (obtained via `System.getProperties()`), copies them and then restores them to this state after the test execution.
The name of the annotation could for example be `@RestoreSystemProperties`. This is the same name used by [System Rules](https://stefanbirkner.github.io/system-rules/#RestoreSystemProperties), but it is quite accurate.

Considerations / open questions:
- Should usage of this annotation be allowed together with `@ClearSystemProperty` and `@SetSystemProperty`?
If so, it should be checked if there are any corner cases where this could cause issues.
- System properties can, despite being discouraged, contain non-`String` keys and values. It is probably not worth it add any special handling for them (such as performing deep copies), however the documentation should then document this behavior.
- Should the annotation be allowed on class level?
Probably, would definitely be useful.
- When used to annotate a test class, should the annotation apply to before-all & after-all, before-each & after-each or only the test method itself (see also #479)?

eeverman:
WIP:  I intend to add documentation and much more testing (especially the interaction of set/clear/restore).
I put this PR in for initial feedback / sanity check.
All comments welcome - I already threw out one implementation that I decided was too ugly :-)

---

Add restore functionality to SystemProps & EnvironmentVars Extensions (#574 / ${pull-request})

To allow users to modify System.Properties and Env Vars directly in their test code outside of the existing
`set` and `clear` annotations, new `restore` annotations revert these resources to their pre-test state.
The new `restore` annotations can be used at the class or method level and the functionality was
added to the existing extension classes.  

Closes #574
PR: ${pull-request}

---
**PR checklist**

Documentation (general)

Documentation (new extension)

Code

Contributing

### DIFF
--- a/src/main/java/org/junitpioneer/jupiter/AbstractEntryBasedExtension.java
+++ b/src/main/java/org/junitpioneer/jupiter/AbstractEntryBasedExtension.java
@@ -36,6 +36,7 @@ import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.jupiter.api.extension.ExtensionContext.Namespace;
 import org.junit.jupiter.api.extension.ExtensionContext.Store;
 import org.junit.platform.commons.support.AnnotationSupport;
+import org.junitpioneer.internal.PioneerAnnotationUtils;
 import org.junitpioneer.internal.PioneerUtils;
 
 /**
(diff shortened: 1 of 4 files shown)


eeverman:
I had two versions of this solution and created the merge request from the wrong one.  Closing this PR - Please see the v2.

Issue #706, opened 2023-01-12
Update code to Java 11


nipafx:
Now that we baseline against Java 11, we can use newer language features and APIs. In some places, the improvements are marginal (e.g. most `var`s), but in others the code becomes noticeably more readable (e.g. handling of `InaccessibleObjectException` in `EnvironmentVariableUtils`). Making a dedicated effort to find the latter pieces of code and refactoring them will improve the code base.

talios:
If the target bytecode is now 11 - can we have a major version bump?

Michael1993:
> If the target bytecode is now 11 - can we have a major version bump?

You mean to, like, 2.0.0?

talios:
> You mean to, like, 2.0.0?

Ewps *hangs head in shame and hides* - um, yes - Just noticed that bump's been made already.  Move along - nothing to see here :)

beatngu13:
Closed via #707.

PR #707, opened 2023-01-12
Update code to Java 11

nipafx:
A first stab ab #706.

---
**PR checklist**

Documentation (general)

Documentation (new extension)

Code

Contributing


src/main/java/org/junitpioneer/internal/PioneerAnnotationUtils.java
```diff
@@ -10,6 +10,8 @@
 
 package org.junitpioneer.internal;
 
+import static java.util.stream.Collectors.toUnmodifiableList;
```
beatngu13, line 13: Given your [recent tweet](https://twitter.com/nipafx/status/1613812216419028993), and the changes in this commit, I assume you prefer to statically import methods from `Collectors`?! 😉 Then, we should probably enforce this via Spotless.

src/main/java/org/junitpioneer/jupiter/EnvironmentVariableUtils.java
```diff
@@ -98,15 +99,12 @@ private static Map<String, String> getFieldValue(Class<?> clazz, Object object,
...
-		catch (RuntimeException ex) {
-			// Java 9 added InaccessibleObjectException, but we compile against Java 8.
-			if (ex.getClass().getName().equals("java.lang.reflect.InaccessibleObjectException"))
-				throw new ExtensionConfigurationException(
-					"Cannot access Java runtime internals to modify environment variables. "
-							+ "Have a look at the documentation for possible solutions: "
-							+ "https://junit-pioneer.org/docs/environment-variables/#warnings-for-reflective-access",
-					ex);
-			throw ex;
+		catch (InaccessibleObjectException ex) {
```
beatngu13, line 102: Good catch (literally)!

nipafx: dad

src/main/java/org/junitpioneer/internal/PioneerAnnotationUtils.java
nipafx: I'm not sure we need to enforce such details, but we _can_. What would that rule look like?

jbduncan: I'm guessing [OpenRewrite](https://github.com/openrewrite/rewrite) or [Refaster](https://errorprone.info/docs/refaster) could be used to auto-format code to use this style.

beatngu13: @jbduncan we already have Checkstyle and Spotless. I couldn't find a corresponding functionality in Spotless (which I prefer, because it not just detects violations, but also fixes them), but Checkstyle can do this this:

https://stackoverflow.com/a/47649243

So I guess we could extend our [`checkstyle.xml`](https://github.com/junit-pioneer/junit-pioneer/blob/main/.infra/checkstyle/checkstyle.xml) as follows:

```xml
<module name="RegexpSinglelineJava">
	<property name="format" value="import java.util.stream.Collectors;"/>
	<property name="message" value="Collectors must be statically imported"/>
</module>

Comment (jbduncan) [reply src/main/java/org/junitpioneer/internal/PioneerAnnotationUtils.java:13]: @beatngu13 Okay, if there's no desire to include another tool, then I totally understand!

But for reference, here is OpenRewrite's doc on auto-fixing Checkstyle issues: https://docs.openrewrite.org/running-recipes/popular-recipe-guides/automatically-fix-checkstyle-violations

Comment (jbduncan) [reply src/main/java/org/junitpioneer/internal/PioneerAnnotationUtils.java:13]: A potential con for OpenRewrite: In my experience, its Gradle plugin runs a fair bit slower than Spotless, but I only have experience with it on a trivially small project, so YMMV.

Comment (beatngu13) [reply src/main/java/org/junitpioneer/internal/PioneerAnnotationUtils.java:13]: @jbduncan oh, nice – didn't know that, thank you!

File: src/main/java/org/junitpioneer/internal/PioneerAnnotationUtils.java [hunk 3]
```diff
@@ -220,18 +217,15 @@ private static <A extends Annotation> Stream<A> findOnOuterClasses(Optional<Clas
...
-			return Collections.emptyList();
+			return List.of();
 		if (findRepeated)
 			return AnnotationSupport.findRepeatableAnnotations(element, annotationType);
 
-		List<A> onElement = AnnotationSupport
-				.findAnnotation(element, annotationType)
-				.map(Collections::singletonList)
-				.orElse(Collections.emptyList());
+		List<A> onElement = AnnotationSupport.findAnnotation(element, annotationType).map(List::of).orElse(List.of());
```
Michael1993, line 224:
```suggestion
		List<A> onElement = AnnotationSupport.findAnnotation(element, annotationType).stream().collect(toUnmodifiableList());
```

src/main/java/org/junitpioneer/internal/PioneerAnnotationUtils.java
Michael1993: I have updated the checkstyle ruleset but I'm not sure about this approach - we _could_ use the [IllegalImport](https://checkstyle.sourceforge.io/config_imports.html#IllegalImport) check.

beatngu13: I'm fine with both, but `IllegalImport` seems more specific, so feel free to use it.

build.gradle.kts
```diff
@@ -294,7 +292,6 @@ tasks {
 							filter {
 								includeTestsMatching("*Demo")
 							}
-							jvmArgs(testJvmArgs)
```
beatngu13, line 297: Aren't our demo tests running on the module path, why is that obsolete? 😳

beatngu13 (approved):

src/main/java/org/junitpioneer/internal/PioneerAnnotationUtils.java
Bukama: I like the `IllegalImport` rule too as it's to covered the `sun.` packages

build.gradle.kts
Michael1993: Are we? That folder doesn't have a `module-info.java` and gradle build has this warning:
```
> Task :demoTests
WARNING: Unknown module: org.junitpioneer specified to --patch-module
WARNING: Unknown module: org.junitpioneer specified to --add-opens
WARNING: Unknown module: org.junitpioneer specified to --add-opens
```

src/main/java/org/junitpioneer/internal/PioneerAnnotationUtils.java
Michael1993: I think I actually prefer this rule after thinking about it. `IllegalImport` is specifically about _not using_ certain classes/packages. This is not the case here - we _want to_ use `Collectors` but only with static imports.

build.gradle.kts
beatngu13: Indeed. 👍

src/main/java/org/junitpioneer/jupiter/DisableIfTestFails.java
```diff
@@ -20,18 +20,18 @@
 
 /**
  * Disables all remaining tests in a container if one of them failed.
- *
+ * <p>
```
Bukama, line 23: You always open a paragraph, but you don't close them. That's invalid HTML https://www.w3schools.com/html/html_paragraphs.asp

Note: Don't use <br>-Tag to create empty lines.

Bukama (changes requested):
I would like to see that we write valid HTML (see comment on "p"-Tag, applies on many changes in the PR)

src/main/java/org/junitpioneer/jupiter/StdIo.java
```diff
@@ -21,7 +21,7 @@
  * Allows specifying the input that's read from {@code System.in} as well as capturing
  * lines read from {@code System.in} (with parameter {@link StdIn}) or
  * written to {@code System.out} (with parameter {@link StdOut StdOut}).
- *
+ * <p>
```
Bukama, line 24: Here a closing tag is missing in line 28 (can't comment the not changed line)

src/main/java/org/junitpioneer/jupiter/cartesian/CartesianArgumentsSource.java
```diff
@@ -25,7 +25,7 @@
  * <p>{@code @CartesianArgumentsSource} may also be used as a meta-annotation in order to
  * create a custom <em>composed annotation</em> that inherits the semantics
  * of {@code @CartesianArgumentsSource}.
- *
+ * <p>
```
Bukama, line 28: I think this should be a closing, not an opening tag

src/main/java/org/junitpioneer/jupiter/cartesian/CartesianMethodArgumentsProvider.java
```diff
@@ -14,7 +14,7 @@
 
 /**
  * Provides arguments for all parameters of a {@link CartesianTest} method.
- *
+ * <p>
  * For more information, see
  * <a href="https://junit-pioneer.org/docs/cartesian-product/" target="_top">the Cartesian product documentation</a>.
  */
```
Bukama, line 20: Missing closing tag

src/main/java/org/junitpioneer/jupiter/cartesian/CartesianParameterArgumentsProvider.java
```diff
@@ -17,7 +17,7 @@
 
 /**
  * Provides arguments for a single parameter of a {@link CartesianTest} method.
- *
+ * <p>
```
Bukama, line 20: Missing closing tag

src/main/java/org/junitpioneer/jupiter/issue/IssueExtensionExecutionListener.java
```diff
@@ -93,23 +94,22 @@ public void testPlanExecutionFinished(TestPlan testPlan) {
...
+				.stream()
+				.collect(toMap(IssueTestCaseBuilder::getIssueId, this::getIssueTestCases, this::mergeIssueTestCases))
+				.entrySet()
+				.stream()
+				.map(entry -> new IssueTestSuite(entry.getKey(), entry.getValue()))
 				.collect(toUnmodifiableList());
-		//@formatter:on
+	}
+
+	private List<IssueTestCase> getIssueTestCases(IssueTestCaseBuilder t) {
```
beatngu13, line 107: Why `t`, for "test" / "test case"? If you want to stick to a single character, maybe `b`? (I think I would use `builder`.)

Bukama (commented):
One more closing paragraph tag and I'm fine with it (if you added it, see this an approval and feel free to merge, as @beatngu13 already accepted)

Michael1993:
Proposed commit message:
```
Update code to Java 11 (#706 / #707)

This commit updates the code base to use the language features now
available to us (as we baseline against Java 11).

PR: #707
```

Bukama (approved):
Thank you for your patience :)

PR #541, opened 2021-11-15
Adds FreePort test extension

p1729:
Proposed commit message:

```
Adds extension for providing free port (#141 / #541)

This parameter resolution extension provides with a free port to test.

Closes: #141
PR: #541
```

---
**PR checklist**

The following checklist shall help the PR's author, the reviewers and maintainers to ensure the quality of this project.
It is based on our contributors guidelines, especially the ["writing code" section](https://github.com/junit-pioneer/junit-pioneer/blob/main/CONTRIBUTING.md#writing-code).
It shall help to check for completion of the listed points.
If a point does not apply to the given PR's changes, the corresponding entry can be simply marked as done. 

Documentation (general)
* [x] There is documentation (Javadoc and site documentation; added or updated)
* [ ] There is implementation information to describe _why_ a non-obvious source code / solution got implemented
* [x] Site documentation has its own `.adoc` file in the `docs` folder, e.g. `docs/report-entries.adoc`
* [x] Only one sentence per line (especially in `.adoc` files)
* [ ] Javadoc uses formal style, while sites documentation may use informal style

Documentation (new extension)
* [x] The `docs/docs-nav.yml` navigation has an entry for the new extension
* [x] The `package-info.java` contains information about the new extension

Code
* [x] Code adheres to code style, naming conventions etc.
* [x] Successful tests cover all changes
* [x] There are checks which validate correct / false usage / configuration of a functionality and there are tests to verify those checks
* [x] Tests use [AssertJ](https://assertj.github.io/doc/) or our own [PioneerAssert](https://github.com/junit-pioneer/junit-pioneer/blob/main/CONTRIBUTING.md#assertions) (which are based on AssertJ)

Contributing
* [x] A prepared commit message exists
* [ ] The list of contributions inside `README.md` mentions the new contribution (real name optional) 

---

I hereby agree to the terms of the [JUnit Pioneer Contributor License Agreement](https://github.com/junit-pioneer/junit-pioneer/blob/main/CONTRIBUTING.md#junit-pioneer-contributor-license-agreement).

Issue #141, opened 2019-08-14
Add a FreePort extension


krzyk:
This is quite similar to `TempDirectory`.

There are frequently tests that depend on network layer that need to use a port that is not already used by any other service (either to start a server on it, or to test a failure scenario where there is no service at given port).

Proposed usage:
```
@Test
@ExtendWith(FreePort.class)
void test(@FreePrt int port) {
    // do something with the `port`
}
```

marcphilipp:
How do you guarantee the port stays free between the time you've checked it and the test actually using it? Usually you can pass 0 to servers to assign a randomly allocated port and then query the port that's been allocated after they're started. This approach doesn't have the race condition problem. Do you have use cases where this approach wouldn't work?

krzyk:
Yes, race condition is problematic, but in some cases it can't be avoided, and the proposition is still better than hardcoding port number.

1. I have a test that needs to have nothing listening at the port (to test the application behavior for invalid host+port).
2. Some Java libraries allow passing a `0` as a port, but don't have a way to retrieve the assigned port back, so I need to pass a known free port there (yes, with possibility that it will be taken).

As for 2, I don't recall the names, because I had those problems some time ago, but as for 1, I needed to create such extension to be able to test failure conditions. So maybe someone else would find it useful also.

sbrannen:
> How do you guarantee the port stays free between the time you've checked it and the test actually using it?

You of course cannot guarantee that the port remains free, but there are use cases for it.

For example, Spring has [`SocketUtils`](https://github.com/spring-projects/spring-framework/blob/master/spring-core/src/main/java/org/springframework/util/SocketUtils.java) exactly for such use cases.

nicolaiparlog:
I think this is an interesting idea, but I don't have enough experience with testing port-reliant code to add anything to the discussion except a few (possibly naive) ideas. Here's one:

Would it make sense to design the extension so that the strategy to find a free port can be plugged in? That way, a Spring app (to use a random example) can use Spring's facilities to discover free ports and other apps use other mechanism. Or users could ask for random port scan or for a linear port scan in a certain range, ...

What do you think, @krzyk?

Bukama:
I would guess that for such things testcontainers are the right thing to use as they handle the port things out of the box?

krzyk:
@Bukama Running docker just for a single test (or even for the whole suit) is a overkill in majority of projects.

nicolaiparlog:
[...]

PR #733, opened 2023-05-02
FreePort resource to create ServerSocket

Michael1993:
Another approach to #141. Based on #541


Michael1993:
A couple things that need to be updated still:
- documentation has to be updated
- add `@NewPort`/`@SharedPort` annotation (?)

src/main/java/org/junitpioneer/jupiter/resource/NewPort.java
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+package org.junitpioneer.jupiter.resource;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@New(FreePort.class)
```
Bukama, line 18: I think it means that a new instance of the `FreePort` class is created, but I would like to double check as I know `@New` only as a deprecated CDI annotation (and I'm not getting any other proper search result from google 🆘). This would mean that the user always get a Socket on port zero - which is hardly what they want I guess.

Bukama (commented):
Aside from my thought that I might have an understanding issue personally, please add some Javadoc for the class and interface (and maybe answer my personal questions :) )

src/main/java/org/junitpioneer/jupiter/resource/NewPort.java
```diff
@@ -0,0 +1,22 @@
...
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.PARAMETER)
+public @interface NewPort {
```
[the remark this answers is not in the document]
Michael1993: Decided against adding this - it's not really useful.

src/main/java/org/junitpioneer/jupiter/resource/FreePort.java
```diff
@@ -0,0 +1,48 @@
...
+ * http://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junitpioneer.jupiter.resource;
+
+import java.io.IOException;
+import java.net.ServerSocket;
+import java.util.List;
+
+public final class FreePort implements ResourceFactory<ServerSocket> {
```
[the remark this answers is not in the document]
Michael1993: Added documentation.

src/main/java/org/junitpioneer/jupiter/resource/NewPort.java
Michael1993: Again - NewPort was removed. The `@New` annotation is part of our resources abstract extension.

Michael1993:
I intend to tackle #808 in this one when I have the time, actually.

src/main/java/org/junitpioneer/jupiter/resource/NewPort.java
Bukama: Thanks for updating.

Bukama:
Looks good to me, @beatngu13 would you also take a look?

Bukama (approved):

Issue #734, opened 2023-05-09
Argument converter for converting integral numbers to byte arrays


scordio:
In some projects related to cryptography, I'm often dealing with bitwise operations and I would like to write tests like the following:

```java
@ParameterizedTest
@ValueSource(ints = {Integer.MIN_VALUE, 0, 0xFF, 0xFFFF, 0xFFFFFF, Integer.MAX_VALUE})
void test(byte[] byteArray) {
  assertThat(byteArray).hasSize(4);
  // do some real stuff
}
```

where I can leverage binary and hexadecimal literals which would be automatically converted to a byte array using under the hood a [`ByteBuffer`](https://docs.oracle.com/javase/8/docs/api/java/nio/ByteBuffer.html) with [`BIG_ENDIAN`](https://docs.oracle.com/javase/8/docs/api/java/nio/ByteOrder.html#BIG_ENDIAN) order.

However, JUnit doesn't offer an implicit conversion for this combination of types. I originally raised junit-team/junit5#2702 but the team [prefers not to support it](https://github.com/junit-team/junit5/issues/2702#issuecomment-955676577) and would rather suggest an opt-in annotation, similar to [`@JavaTimeConversionPattern`](https://junit.org/junit5/docs/current/api/org.junit.jupiter.params/org/junit/jupiter/params/converter/JavaTimeConversionPattern.html).

Also, they suggested JUnit Pioneer could host it 🙂 

AFAIK, JUnit Pioneer doesn't have standalone argument converters. Is it an addition you would accept?

Written in a style similar to the [official JUnit docs](https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests-argument-conversion-implicit), this is what I'd like to have from the custom converter:

| Source            | Target      | Example                                                                            |
|-------------------|-------------|------------------------------------------------------------------------------------|
| `byte` / `Byte`   | `byte[]`    | `0xF` &#8594; `ByteBuffer.allocate(Byte.BYTES).put((byte) 0xF).array()`                     |
| `short` / `Short` | `byte[]`    | `0xFFF` &#8594; `ByteBuffer.allocate(Short.BYTES).putShort((short) 0xFFF).array()`           |
| `int` / `Integer` | `byte[]`    | `0xFFFFFFF` &#8594; `ByteBuffer.allocate(Integer.BYTES).putInt(0xFFFFFFF).array()`       |
[...]

PR #742, opened 2023-07-05
Add basic conversion from integral types to byte array

Michael1993:
Proposed commit message:

```
Adds number to byte array converter (#734 / #742)

Adds a new extension that converts numbers to their byte array
representation.

Closes: #734
PR: #742
```

---
**PR checklist**

Documentation (general)

Documentation (new extension)

Code

Contributing


src/main/java/org/junitpioneer/jupiter/params/IntegralToBytesConverter.java
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	@Override
+	public byte[] convert(Object source, ParameterContext context) throws ArgumentConversionException {
+		java.nio.ByteOrder byteOrder = AnnotationSupport
+				.findAnnotation(context.getParameter(), IntegralToBytesConversion.class)
+				.map(IntegralToBytesConversion::byteOrder)
+				.map(ByteOrder::getValue)
+				.orElseThrow(() -> new IllegalStateException(
+					format("Could not find %s annotation on the parameter", IntegralToBytesConversion.class)));
```
scordio, line 35: In my local attempt, I used [`AnnotationConsumer`](https://junit.org/junit5/docs/5.9.2/api/org.junit.jupiter.params/org/junit/jupiter/params/support/AnnotationConsumer.html) to capture the target annotation in a private field (basically taking inspiration from [`JavaTimeArgumentConverter`](https://github.com/junit-team/junit5/blob/bd7d03f517dfca3730fbd123c5fd4d4b70930129/junit-jupiter-params/src/main/java/org/junit/jupiter/params/converter/JavaTimeArgumentConverter.java#L35-L36) of JUnit).

Something like the following:

```java

public class ByteArrayArgumentConverter extends TypedArgumentConverter<Number, byte[]>
		implements AnnotationConsumer<ByteArrayConversion> {

	private java.nio.ByteOrder order;

	ByteArrayArgumentConverter() {
		super(Number.class, byte[].class);
	}

	@Override
	public void accept(ByteArrayConversion byteArrayConversion) {
		switch (byteArrayConversion.order()) {
			case BIG_ENDIAN:
				order = java.nio.ByteOrder.BIG_ENDIAN;
				break;
			case LITTLE_ENDIAN:
				order = java.nio.ByteOrder.LITTLE_ENDIAN;
				break;
			default:
				throw new IllegalArgumentException("Unsupported byte order: " + byteArrayConversion.order());
		}
	}

	@Override
	protected byte[] convert(Number source) throws ArgumentConversionException {
		...
	}

```

(`java.nio.ByteOrder` can actually be imported as it doesn't clash with the `ByteOrder` from this package)

src/main/java/org/junitpioneer/jupiter/params/IntegralToBytesConverter.java
```diff
@@ -0,0 +1,56 @@
...
+import static java.lang.String.format;
+
+import java.nio.ByteBuffer;
+
+import org.junit.jupiter.api.extension.ParameterContext;
+import org.junit.jupiter.params.converter.ArgumentConversionException;
+import org.junit.jupiter.params.converter.ArgumentConverter;
+import org.junit.platform.commons.support.AnnotationSupport;
+
+class IntegralToBytesConverter implements ArgumentConverter {
```
scordio, line 22: In my local attempt, I used a different name for the converter, `ByteArrayArgumentConverter`, which is closer to what I've used for the annotation (see other comment and, again, it's just a proposal and the final choice is yours, of course 🙂 )

Also, what if the converter would extend [`TypedArgumentConverter`](https://junit.org/junit5/docs/5.9.2/api/org.junit.jupiter.params/org/junit/jupiter/params/converter/TypedArgumentConverter.html)? It would get [some basic error handling](https://github.com/junit-team/junit5/blob/bd7d03f517dfca3730fbd123c5fd4d4b70930129/junit-jupiter-params/src/main/java/org/junit/jupiter/params/converter/TypedArgumentConverter.java#L51-L64) out of the box.

```suggestion
class ByteArrayArgumentConverter extends TypedArgumentConverter<Number, byte[]> {
```

AFAIR, JUnit Pioneer tries to stay compatible with older versions of JUnit, and using `TypedArgumentConverter` would require at least JUnit 5.7. I wasn't sure if the version [here](https://github.com/junit-pioneer/junit-pioneer/blob/main/gradle.properties#L1) is the current minimum required by Pioneer. If so, `TypedArgumentConverter` could be an option.

src/main/java/org/junitpioneer/jupiter/params/IntegralToBytesConversion.java
```diff
@@ -0,0 +1,38 @@
...
+import org.junit.jupiter.params.converter.ConvertWith;
+
+/**
+ * Annotation to convert an integral type (byte, short, int, long) to a byte array.
+ * The converter uses {@link java.nio.ByteBuffer} under the hood.
+ */
+@Target(ElementType.PARAMETER)
+@Retention(RetentionPolicy.RUNTIME)
+@ConvertWith(IntegralToBytesConverter.class)
+public @interface IntegralToBytesConversion {
```
scordio, line 27: As usual, naming is a challenge 😄 I wasn't happy with any of the proposals I made in #734 so I've introduced one more in my local attempt:

```suggestion
public @interface ByteArrayConversion {
```

Of course, please pick the one you like most.

src/main/java/org/junitpioneer/jupiter/params/IntegralToBytesConversion.java
```diff
@@ -0,0 +1,38 @@
...
+
+	/**
+	 * The byte order to use during the conversion.
+	 * Note that this is NOT {@link java.nio.ByteOrder}
+	 * because that is not an enum (and can not be used in an annotation).
+	 *
+	 * @return the byte order to use, either {@code BIG_ENDIAN} or {@code LITTLE_ENDIAN}.
+	 */
+	ByteOrder byteOrder() default ByteOrder.BIG_ENDIAN;
```
scordio, line 36: Going with the `ByteArrayConversion` name, I had the feeling `order` would be less verbose here:

```suggestion
	ByteOrder order() default ByteOrder.BIG_ENDIAN;
```

src/main/java/org/junitpioneer/jupiter/params/ByteOrder.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * http://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junitpioneer.jupiter.params;
+
+/**
+ * Utility enum because {@link java.nio.ByteOrder} is not an enum.
+ */
+public enum ByteOrder {
```
scordio, line 16: What about moving this enum into the annotation definition? In the style of [`@EnumSource`](https://github.com/junit-team/junit5/blob/r5.9.3/junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/EnumSource.java) and its [`Mode`](https://github.com/junit-team/junit5/blob/bd7d03f517dfca3730fbd123c5fd4d4b70930129/junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/EnumSource.java#L95-L181) enum.

src/main/java/org/junitpioneer/jupiter/params/ByteOrder.java
```diff
@@ -0,0 +1,42 @@
...
+
+	private final java.nio.ByteOrder value;
+
+	ByteOrder(java.nio.ByteOrder value) {
+		this.value = value;
+	}
+
+	public java.nio.ByteOrder getValue() {
+		return value;
+	}
```
scordio, line 40: Maybe this mapping doesn't need to be exposed as public, I had the feeling it's more like an implementation detail.

In the comment where I mentioned the `AnnotationConsumer` usage, I've kept the mapping to `java.nio.ByteOrder` inside the converter implementation.

src/main/java/org/junitpioneer/jupiter/params/ByteOrder.java
```diff
@@ -0,0 +1,42 @@
...
+ * Copyright 2016-2022 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junitpioneer.jupiter.params;
```
scordio, line 11: Still taking inspiration from `JavaTimeConversionPattern`, JUnit uses a dedicated package for converters:

```suggestion
package org.junitpioneer.jupiter.params.converter;
```

src/main/java/org/junitpioneer/jupiter/converter/ByteArrayArgumentConverter.java
```diff
@@ -0,0 +1,63 @@
...
+			var bytes = (short) source;
+			return ByteBuffer.allocate(Short.BYTES).order(order).putShort(bytes).array();
+		} else if (source instanceof Integer) {
+			var bytes = (int) source;
+			return ByteBuffer.allocate(Integer.BYTES).order(order).putInt(bytes).array();
+		} else if (source instanceof Long) {
+			var bytes = (long) source;
+			return ByteBuffer.allocate(Long.BYTES).order(order).putLong(bytes).array();
+		}
+		throw new ArgumentConversionException(format("Unsupported parameter type: %s", source.getClass()));
```
scordio, line 52: Would you like to add support also for  `Float` and `Double`? Personally, I have no concrete use case for them but it would be trivial since `ByteBuffer` already supports them.

scordio (approved):
LGTM and thanks again, @Michael1993!

src/test/java/org/junitpioneer/jupiter/converter/ByteArrayConversionTests.java
```diff
@@ -0,0 +1,99 @@
...
+import org.junitpioneer.jupiter.converter.ByteArrayConversion.ByteOrder;
+import org.junitpioneer.testkit.ExecutionResults;
+import org.junitpioneer.testkit.PioneerTestKit;
+import org.junitpioneer.testkit.assertion.PioneerAssert;
+
+public class ByteArrayConversionTests {
+
+	@ParameterizedTest
+	@ValueSource(ints = { Integer.MIN_VALUE, 0, 0xFF, 0xFFFF, 0xFFFFFF, Integer.MAX_VALUE })
+	void test(@ByteArrayConversion byte[] byteArray) {
```
scordio, line 35:
```suggestion
	void testInts(@ByteArrayConversion byte[] byteArray) {
```

src/main/java/org/junitpioneer/jupiter/converter/ByteArrayArgumentConverter.java
Bukama: If it's easy to implement we should support it right from the start  I guess.

scordio: Actually, I was also thinking about wheter this converter should be generalized further, adding the support from `String` to `byte[]` via [`getBytes(String)`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html#getBytes(java.lang.String)).

Right now, just from the `@ByteArrayConversion` name is not clear why the conversion is limited to numeric types (naming was my proposal so definitely my fault 😅).

I see two options:
* `@ByteArrayConversion` renamed to `@IntegralToByteArrayConversion` or `@NumberToByteArrayConversion` (in case float and double are added)
* Add support for string conversion with an additional `charset` attribute, which would be `""` by default (implying [`getBytes()`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html#getBytes()) usage).

In case of the latter, `byteOrder` would be ignored when there is a string input, while `charset` would be ignored for numeric ones. Maybe not ideal but somehow similar to what [`@ValueSource`](https://junit.org/junit5/docs/5.9.3/api/org.junit.jupiter.params/org/junit/jupiter/params/provider/ValueSource.html) does today.

WDYT?

Michael1993: Personally, I'd prefer not to include `getBytes` just to make sure that no one thinks that it's a number conversion, e.g.: `String` -> `BigDecimal` -> `byte[]`

docs/byte-array-converter.adoc
```diff
@@ -0,0 +1,34 @@
+:page-title: Convert Number Argument to Byte Array
+:page-description: The JUnit 5 (Jupiter) extension `@NumberToByteArrayConversion` convert the number value of an argument to its byte array representation
```
scordio, line 2: What about `argument converter` instead of `extension`?

My concern is the terminology clash with [`Extension`](https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/extension/Extension.html).

Applies also to other sentences in this file.

scordio (commented):
Hi @Michael1993, sorry, I'm a little late to provide feedback. I have only one cosmetic point for the user guide.

docs/byte-array-converter.adoc
scordio: Superseded by #751.

Issue #769, opened 2023-10-06
Support `@ExpectedToFail(onlyIf=MyException.class)`


knutwannheden:
I really loved using JUnit 4's `@Test(expected)` for the use case for which `@ExpectedToFail` was designed thanks for this!). The only thing I am missing is a way to restrict this feature to only apply when the test case failed with a given exception. Sometimes I would for example throw an `UnsupportedOperationException` in the production code and only want the corresponding tests not to fail, when that is the reason, not when there is an `AssertionError` (which may instead indicate a bug).

beatngu13:
I like that idea! With `onExceptions`, we already have something similar in `@RetryingTest`:

https://junit-pioneer.org/docs/retrying-test/#configuring-on-which-exceptions-to-retry

Not sure if I would go with `onlyIf`, but that's a detail. What does everybody else think?

If feedback is positive, @knutwannheden would you be interested in giving this a shot (i.e. provide a PR)?

nipafx:
Quoting from the `@ExpectedToFail` documentation:

> This annotation is not intended as a way to mark test methods which intentionally cause exceptions. Such test methods should use [JUnit’s assertThrows](https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/Assertions.html#assertThrows(java.lang.Class,org.junit.jupiter.api.function.Executable)) or similar means to explicitly test for a specific exception class being thrown by a specific action.

knutwannheden:
@nipafx Yes, I get that part and I don't see any real contradiction to my use case. The current annotation can be thought to have `onExceptions=Throwable.class`.

In the use case I outlined I have developed a set of tests (e.g. to match some specification) and in the implementation I throw `UnsupportedOperationException`s. Now as I make progress I fix the production code by replacing the exception with a proper implementation, at which point the `@ExpectedToFail` annotation can be removed again.

nipafx:
Yeah, I see your point. `@ExpectedToFail` essentially says "I know this code fails - let me know when that changes" and with your addition, it's "I know this code fails _for a specific reason_ - let me know when that changes". Makes sense, I'll have a look at the PR-

PR #774, opened 2023-10-07
Add `withExceptions` attribute to `@ExpectedToFail`

knutwannheden:
Adds a new `onExceptions` attribute to `@ExpectedToFail` using which the scope of the extension can be limited to only consider the test as successful if the thrown exception matches one of the given types.

Fixes: #769


knutwannheden:
This is still in draft state, as I will need to add more documentation and tests. I just wanted to make sure I am not pulling in the wrong direction with this and also check if you already have some input on this.

src/main/java/org/junitpioneer/jupiter/ExpectedToFailExtension.java
```diff
@@ -42,7 +42,12 @@ private static void invokeAndInvertResult(Invocation<Void> invocation, Extension
 				throw t;
 			}
 
-			String message = getExpectedToFailAnnotation(extensionContext).value();
+			ExpectedToFail expectedToFail = getExpectedToFailAnnotation(extensionContext);
+			if (!expectedToFail.onExceptions()[0].isInstance(t)) {
```
Michael1993, line 46: I thought this was an array but you only check the first element?

src/test/java/org/junitpioneer/jupiter/ExpectedToFailExtensionTests.java
```diff
@@ -101,6 +101,16 @@ void failsOnWorkingTest() {
...
 	}
 
+	@Test
+	void failsOnTestThrowingUnexpectedException() {
+		ExecutionResults results = PioneerTestKit.executeTestMethod(ExpectedToFailTestCases.class, "unexpectedException");
+		assertThat(results)
+				.hasSingleStartedTest()
+				.whichFailed()
+				.withExceptionInstanceOf(AssertionError.class)
+				.hasMessage("Test marked as 'expected to fail' failed with an unexpected class java.lang.IllegalArgumentException exception");
```
Michael1993, line 111: I would not bake in the classname to the test, but that's personal preference.

The exception is actually an [AssertionFailedError](https://ota4j-team.github.io/opentest4j/docs/1.0.0/api/org/opentest4j/AssertionFailedError.html?is-external=true) an extension of but not an `AssertionError` (exactly).

src/main/java/org/junitpioneer/jupiter/ExpectedToFailExtension.java
```diff
@@ -42,7 +42,12 @@ private static void invokeAndInvertResult(Invocation<Void> invocation, Extension
...
+				fail("Test marked as 'expected to fail' failed with an unexpected " + t.getClass() + " exception");
```
Michael1993, line 47: `Assertions::fail` can take a second parameter `Throwable` cause, which I think is applicable here?

Michael1993 (commented):
I am undecided if I like this proposal. I really appreciate you taking the initiative and I'm not outright opposed to it, it's just that I worry that despite this being something for a specific use-case, people will be tempted to use it as a "fix-all". Does that make sense?

knutwannheden:
@Michael1993 I definitely agree with your review comments. I just slapped this together to better convey the idea and to have something concrete to discuss.

I understand your concern but I honestly don't quite see the "fix-all" problem. If anything it is JUnit's own `@Disabled` I see as a fix-all, that gets used (and then forgotten!) far too often. Also `@ExpectedToFail` annotations can linger on for a long time, which I think is OK, since they already narrow down the scope a lot (the test actually must fail). Now, this `onExceptions` attribute narrows this scope down even more, because the test must fail with a specific exception.

My only "concern" as a maintainer would be that the use case is probably rather unusual. But as I think I already mentioned in the issue, JUnit 4's `@Test(expected)` actually allowed this (even though it was probably primarily designed for what today `assertThrows()` should be used for) and I found it extremely useful when for example working on a new implementation for some API spec.

Michael1993:
Okay, I see your point. Please go ahead with the PR. 😄

src/main/java/org/junitpioneer/jupiter/ExpectedToFailExtension.java
knutwannheden: So you mean that if the test throws an `AssertionFailedError` then we should check its cause (if any) against the `onExceptions` types?

Michael1993: Hm, no, I don't think that's right. Assertion errors are assertion failures which is not the use-case I had in mind... Unless you can justify it?

I just meant we should not be "swallowing" unexpected exceptions because it makes debugging harder.

src/main/java/org/junitpioneer/jupiter/ExpectedToFail.java
```diff
@@ -33,6 +33,13 @@
  * of disabled tests over time.
  * </p>
  *
+ * <p>Further, the {@link #onExceptions()} attribute can be used to restrict the extension's behavior
+ * to specific exceptions. I.e. only if the test method end up throwing one of the specified exceptions
+ * will the test be aborted. This can for example be used when the production code temporarily throws
+ * an {@link UnsupportedOperationException} because some feature has not been implemented yet, but the
+ * test method is already implemented and should not be aborted on a failing assertion.
```
Michael1993, line 40: Should not be aborted?

This doesn't really makes sense. I think this should be "should not fail"?

Michael1993 (changes requested):
Please apply spotless and add documentation, if you can. Thank you!

src/main/java/org/junitpioneer/jupiter/ExpectedToFailExtension.java
knutwannheden: OK, I see what you mean. That of course makes sense.

src/demo/java/org/junitpioneer/jupiter/ExpectedToFailExtensionDemo.java
```diff
@@ -18,7 +18,7 @@ public class ExpectedToFailExtensionDemo {
 
 	// tag::expected_to_fail[]
 	@Test
-	@ExpectedToFail
+	@ExpectedToFail(onExceptions = Exception.class)
```
Michael1993, line 21: Sorry, why did you change this?

src/main/java/org/junitpioneer/jupiter/ExpectedToFailExtension.java
```diff
@@ -42,7 +43,12 @@ private static void invokeAndInvertResult(Invocation<Void> invocation, Extension
 				throw t;
 			}
 
-			String message = getExpectedToFailAnnotation(extensionContext).value();
+			ExpectedToFail expectedToFail = getExpectedToFailAnnotation(extensionContext);
+			if (Stream.of(expectedToFail.onExceptions()).noneMatch(c -> c.isInstance(t))) {
```
Michael1993, line 47: Personally, I am not a fan of single letter lambda parameters but that is just personal preference. 🐸

src/demo/java/org/junitpioneer/jupiter/ExpectedToFailExtensionDemo.java
knutwannheden: Hmm... That looks like a mistake. I think I probably did that when I first started looking into this and mistakenly edited this file. Sorry about that!

src/main/java/org/junitpioneer/jupiter/ExpectedToFailExtension.java
knutwannheden: For me it is either way. Paramount is of course the overall code consistency.

src/demo/java/org/junitpioneer/jupiter/ExpectedToFailExtensionDemo.java
```diff
@@ -32,10 +32,23 @@ void doSomething() {
 		int actual = brokenMethod();
 		assertEquals(10, actual);
 	}
-	// end::expected_to_fail_message[]
 
 	private int brokenMethod() {
 		return 0;
 	}
+	// end::expected_to_fail_message[]
```
knutwannheden, line 39: I hope you agree with me moving this end marker and doing the same for my new example.

Michael1993: I don't think the two are equivalent. Your example showcases a planned feature with no body (which could be an actual example of someone using this) but the previous example doesn't actually showcase what the example is talking about... The `brokenMethod` is not actually "broken", and seeing it **not** returning 10 doesn't help understanding. Does this make sense?

I'd move back the tag here but leave the one in your example where it is.

docs/expected-to-fail-tests.adoc
```diff
@@ -44,6 +44,18 @@ A custom message can be provided, explaining why the tested code is not working
 include::{demo}[tag=expected_to_fail_message]
 ----
 
+In some cases it may be desirable to only have the test be aborted on certain exceptions.
```
Michael1993, line 47: I like the documentation change but the project adheres to the "one-sentence-per-line" rule. Could you format it that way, please?

src/demo/java/org/junitpioneer/jupiter/ExpectedToFailExtensionDemo.java
knutwannheden: Yes, that makes sense.

docs/expected-to-fail-tests.adoc
knutwannheden: I should have read that more carefully. Sorry about that!

docs/expected-to-fail-tests.adoc
```diff
@@ -45,11 +45,9 @@ include::{demo}[tag=expected_to_fail_message]
...
 
 In some cases it may be desirable to only have the test be aborted on certain exceptions.
-In the following example a test case for `productionFeature()` has been implemented. The test
-has been fully implemented, but the production code has not been implemented yet and throws an
-`UnsupportedOperationException` to indicate this. Once the production code has been implemented
-the test method will fail if the code were to return the wrong result and fail due to a failing
-assertion.
+In the following example a test case for `productionFeature()` has been implemented.
+The test has been fully implemented, but the production code has not been implemented yet and throws an `UnsupportedOperationException` to indicate this.
+Once the production code has been implemented the test method will fail if the code were to return the wrong result and fail due to a failing assertion.
```
Michael1993, line 50: Wait, is this correct? I thought it would always fail because the expectation is not fulfilled so it doesn't matter what's in the test.

knutwannheden: Maybe you can propose a better way to phrase this. By "Once the production code has been implemented" I mean that `productionFeature()` is now no longer throwing an `UnsupportedOperationException`, but instead maybe the wrong answer (e.g. `11`). Without `onExceptions` the test would continue to be skipped. With `onExceptions` the failing assertion will now cause the test to fail.

Perhaps it would make it clearer if I spell it out like that.

Michael1993:
```
Once the feature has been implemented `@ExpectedToFail` will fail the test as it no longer throws `UnsupportedOperationException`. If you don't use `onExceptions` in `@ExpectedToFail` you risk "masking" your implementation being faulty - which `@ExpectedToFail` would mark as passing.
```

What do you think?

I would also put a warning here to NOT use `onExceptions` for assertion - you can look at Nicolai's comment on the linked issue.

knutwannheden: I've rephrased the documentation yet yet again (and added a subsection header). I hope this will clarify things.

Please note that I am of course also happy with you changing the documentation to whatever you like. I am obviously not a technical writer 😄

Michael1993 (approved):
I'm happy with current state of the PR. I'd ask you to add yourself to the README as a contributor (but that's not mandatory, only if you want to). Thank you for your patience and hard work!

knutwannheden:
> Thank you for your patience and hard work!

Thanks a lot for your thoroughness and patience!

Bukama:
Running full run

src/main/java/org/junitpioneer/jupiter/ExpectedToFail.java
```diff
@@ -33,6 +33,13 @@
  * of disabled tests over time.
  * </p>
  *
+ * <p>Further, the {@link #onExceptions()} attribute can be used to restrict the extension's behavior
+ * to specific exceptions. I.e. only if the test method end up throwing one of the specified exceptions
```
beatngu13, line 37:
```suggestion
 * to specific exceptions. That is, only if the test method ends up throwing one of the specified exceptions
```

src/main/java/org/junitpioneer/jupiter/ExpectedToFailExtension.java
```diff
@@ -42,7 +43,12 @@ private static void invokeAndInvertResult(Invocation<Void> invocation, Extension
 				throw t;
 			}
 
-			String message = getExpectedToFailAnnotation(extensionContext).value();
+			ExpectedToFail expectedToFail = getExpectedToFailAnnotation(extensionContext);
+			if (Stream.of(expectedToFail.onExceptions()).noneMatch(clazz -> clazz.isInstance(t))) {
+				fail("Test marked as 'expected to fail' failed with an unexpected exception", t);
```
beatngu13, line 48:
```suggestion
				fail("Test marked as temporarily 'expected to fail' failed with an unexpected exception", t);
```

nipafx (changes requested):
Hey @knutwannheden, thank you for opening the PR - it looks really good. 😃  A few things could be changed - see below.

But, importantly, can you please open another PR [from a feature branch](https://github.com/junit-pioneer/junit-pioneer/blob/main/CONTRIBUTING.adoc#branching-strategy) and not `main`?  Thanks!

nipafx:
Oh, and if you rebase it on the current `main`, you can fix the merge conflict - it comes from the `README`, which now `.adoc` not `.md`.

knutwannheden:
@nipafx Thanks for your review comments. I assume I should wait until the other maintainers have chimed in. In the meantime I've updated the PR with the latest changes from `main`.

Bukama:
> @nipafx Thanks for your review comments. I assume I should wait until the other maintainers have chimed in. In the meantime I've updated the PR with the latest changes from `main`.

Well what @nipafx didn't wrote: We literally had a team meeting online stream about all Pioneer things yesterday :)

nipafx:
I managed to push a small fix on the README order to your `main` branch, but it's a bit cumbersome. We can keep this PR as is, but please keep in mind for the next one, not to open it from your `main` branch.

Other than that, I will let you take a look at the two remaining comments I made above.

src/main/java/org/junitpioneer/jupiter/ExpectedToFail.java
```diff
@@ -72,4 +78,9 @@
 	 */
 	String value() default "";
 
+	/**
+	 * Specifies which exceptions are expected to be thrown and will cause the test to be aborted rather than fail.
+	 */
+	Class<? extends Throwable>[] onExceptions() default { Throwable.class };
```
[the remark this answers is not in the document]
knutwannheden: I renamed it to `withExceptions` now (notice the plural form). Only supporting a single exception type would however also work well for me.

docs/expected-to-fail-tests.adoc
```diff
@@ -44,6 +44,25 @@ A custom message can be provided, explaining why the tested code is not working
 include::{demo}[tag=expected_to_fail_message]
 ----
 
+=== Only Abort on Expected Exceptions
+
+A test that is `@ExpectedToFail` will change its behavior by starting to actually fail, once the code under test behaves correctly.
+If the underlying failure changes, though, for example from an assertion error to an exception or from an `UnsupportedOperationException` of a formerly missing implementation to a runtime exception of buggy implementation, the `@ExpectedToFail`-test will keep passing.
+
+To better react to such changes, `@ExpectedToFail` has an attribute `onExceptions` that can be used to enumerate the exceptions which when thrown will result in an aborted (and thus passing) test.
```
Michael1993, line 52:
```suggestion
To better react to such changes, `@ExpectedToFail` has an attribute `withExceptions` that can be used to enumerate the exceptions which when thrown will result in an aborted (and thus passing) test.
```

src/main/java/org/junitpioneer/jupiter/ExpectedToFailExtension.java
Michael1993: @knutwannheden You marked this as resolved but I don't see the change?

Michael1993 (changes requested):
If you have the time, please write a few more tests to cover the lines marked in the Sonar analysis.

knutwannheden:
@Michael1993 Sorry about the long delays!

src/main/java/org/junitpioneer/jupiter/ExpectedToFailExtension.java
```diff
@@ -31,6 +32,11 @@ public void interceptTestMethod(Invocation<Void> invocation, ReflectiveInvocatio
 
 	private static void invokeAndInvertResult(Invocation<Void> invocation, ExtensionContext extensionContext)
 			throws Throwable {
+		ExpectedToFail expectedToFail = getExpectedToFailAnnotation(extensionContext);
+		if (expectedToFail.withExceptions().length == 0) {
+			fail("@ExpectedToFail withExceptions must not be empty");
```
Michael1993, line 37: This is not an `AssertionFailedError`, since it does not originate from the test itself - should be an `ExtensionConfigurationException`.

Michael1993 (approved):

docs/expected-to-fail-tests.adoc
beatngu13: @Michael1993 given your approval on December 2, 2023, I mark this as resolved.

beatngu13 (approved):

beatngu13:
Proposed commit message:

```
Add withExceptions to ExpectedToFail (#769 / #774)

Adds a new `withExceptions` attribute to `@ExpectedToFail`, which
allows to limit the scope of the extension to only consider the test
as successful if the thrown exception matches one of the given types.

Closes: #769
PR: #774
```

beatngu13:
@knutwannheden thanks again! 🙏

Issue #735, opened 2023-05-19
@RetryingTest wraps AssertionFailedError, which prevents IntelliJ's "<Click to see difference>" from functioning


liberaj:
When using the `@RetryingTest` functionality of Junit Pioneer, an `AssertionFailedError` will be thrown on the final invokation of the retrying test, with the previously failing tests throwing a `TestAbortedException`. 

This approach works well, except for when the cause of test failure is an `AssertionFailedError`. In this scenario, the original `AssertionFailedError` will be wrapped inside of Pioneer's `AssertionFailedError`, as the `cause`. On paper this sounds like a sensible approach to ensuring that the original assertion details are not lost, but whilst still providing additional details of the status of the `RetryingTest` to the user.

Practically, for users of IntelliJ, this nesting of `AssertionFailedError` will interfere with the ability to view assertion differences.

### Output from Regular `@Test`

```java
    @Test
    void testAssertionFailsNormal() {
        assertEquals("expected", "actual");
    }
```

```text
org.opentest4j.AssertionFailedError: 
Expected :expected
Actual   :actual
<Click to see difference>

<6 internal lines>
	at AlwaysFailingTest.testAssertionFailsNormal(AlwaysFailingTest.java:14) <31 internal lines>
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511) <9 internal lines>
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511) <28 internal lines>
```

### Output from `@RetryingTest`

```java
    @RetryingTest(maxAttempts = 3)
    void testAssertionFailsRetryable() {
        assertEquals("expected", "actual");
    }
```

```text

org.opentest4j.AssertionFailedError: Test execution #3 (of up to 3 with at least 1 successes) failed ~> test fails - see cause for details

	at org.junitpioneer.jupiter.RetryingTestExtension$FailedTestRetrier.failed(RetryingTestExtension.java:158)
	at org.junitpioneer.jupiter.RetryingTestExtension.handleTestExecutionException(RetryingTestExtension.java:62) <26 internal lines>
	at java.base/java.util.Iterator.forEachRemaining(Iterator.java:133) <20 internal lines>
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511) <9 internal lines>
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511) <28 internal lines>
[...]

PR #788, opened 2023-11-07
#735 Removed throwing new AssertionFailedError when test failed

Edmnikishin:
Suggestion for: #735

My point is just throw an exception of what happend, because this annotation (in my opinion) is used to get rid of flaky tests (in api/e2e test especially). Or if it too flaky/broken, so we should fix it.

Thing what annoyed me in my practice, that when you try to make an allure report you'll have same fail issue (Which is not true) like "Test execution #3 (of up to 3 with at least 1 successes) failed ~ test fails - see cause for details"
And second is that you need to dive to previous exception/allure step to know what happened

So what do i mean. Here is how it works right now (without my PR)

Here how it'll work with mine PR

And that's how it'll work if we throw MultipleFailuresError instead AssertionFailedError. In my opinion it bloats fail message and when test failed you want to know cause. And if you want to know what retry it is you can just look at previous retry attempts

---
**PR checklist**

Documentation (general)

Documentation (new extension)

Code (general)
* [ ] Tests use [AssertJ](https://assertj.github.io/doc/) or our own [PioneerAssert](https://github.com/junit-pioneer/junit-pioneer/blob/main/CONTRIBUTING.adoc#assertions) (which are based on AssertJ)

Code (new package)

Contributing
* [ ] The list of contributions inside `README.adoc` mentions the new contribution (real name optional)

### DIFF
--- a/src/main/java/org/junitpioneer/jupiter/RetryingTestExtension.java
+++ b/src/main/java/org/junitpioneer/jupiter/RetryingTestExtension.java
@@ -16,8 +16,10 @@ import static java.util.Spliterators.spliteratorUnknownSize;
 import static java.util.stream.StreamSupport.stream;
 
 import java.lang.reflect.Method;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Iterator;
+import java.util.List;
 import java.util.NoSuchElementException;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Stream;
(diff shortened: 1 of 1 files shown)


Edmnikishin:
> Has the same problem as discussed in #735 - it hides why previous executions failed.

Not sure is that it really needed to be stored at last retry. For me personally it would be better to have it separate and see what happened in each retry. Maybe if someone had 20 retries it could be an issue... But error message would be so huge

Edmnikishin:
Okay, so i've tested it with adding previous errors and it doesn't look bad, actually nice :)
Here what i've got in allure 
image

And here is console
image

nipafx:
Thank you for the contribution, @Edmnikishin. I ended up creating my own change that does indeed throw the `MultipleFailuresError` mentioned in #735. I also changed how exceptions for aborted tests are rethrown and made sure the original exception's message comes first to ease parsing for tools (like you did [here](https://github.com/junit-pioneer/junit-pioneer/pull/788/files#diff-6e9689f1f446ad193b6c6ee231a1eaca2fe7da4a86b390297e226e24343192a0R163).

I'll close this PR.

PR #798, opened 2023-12-19
Fix jacoco output file location

jpenilla:
Before this change:

Screenshot 2023-12-18 at 5 11 36 PM

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -342,7 +342,7 @@ tasks {
 		enabled = !experimentalBuild
 		reports {
 			xml.required.set(true)
-			xml.outputLocation.set(file("${layout.buildDirectory}/reports/jacoco/report.xml"))
+			xml.outputLocation.set(layout.buildDirectory.file("reports/jacoco/report.xml"))
 		}
 	}
 


beatngu13:
@jpenilla thx for the heads-up!

I'm a bit confused, how is [Sonar](https://sonarcloud.io/project/activity?id=junit-pioneer_junit-pioneer&graph=custom&custom_metrics=coverage) still able to determine the coverage, given the report is not where it should be? 🤔

jpenilla:
I'm not sure, but I would guess that they scan for reports recursively to support projects with multiple modules. Something behaving like `**/reports/jacoco/reports.xml`. So it would still end up finding the `reports` directory even though it's nested inside a strange directory.

Michael1993:
This is a nice catch, thank you! That being said, I'd actually prefer completely removing this task as it currently does nothing.

jpenilla:
Do you mean removing the `jacoco` plugin entirely, or just removing the lines where we change its configuration from the defaults? Even if it's not used by CI anymore, it's still useful to run the report task locally when working on improving test coverage.

Bukama:
As @Michael1993 figured out in the meantime:

> 1 XML report (unused) and a jacoco report (used by Sonar)
> Wait, nevermind
> It just doesn't take the output folder into consideration I guess
> But sonar does use the task
> I thought we ran sonar during the check step automatically

So we need the report and decided to keep it:

> Sure. We should fix it then

Bukama:
@beatngu13 Could you please have a look at this as we can merge it or if the path has to be changed again for Sonar? I'm confused here :/

beatngu13 (approved):

Issue #809, opened 2024-04-01
Mention naming convention in contributing


Bukama:
We have a lots of coding convention in our contributing file (https://github.com/junit-pioneer/junit-pioneer/blob/main/CONTRIBUTING.adoc#code-style).

But we don't mentioned that we write test cases in camelCase.

Question came up in Discord

PR #812, opened 2024-04-20
Mention naming convention in CONTRIBUTING.adoc (#809 / #812)

Bukama:
The naming convention for classes and methods was not yet mentioned in our contributing guideline.

closes: #809
PR: #812

### DIFF
--- a/CONTRIBUTING.adoc
+++ b/CONTRIBUTING.adoc
@@ -75,6 +75,12 @@ Everything related to branches, commits, and more is described #fixing-bugs-deve
 
 Where to put types and how to name them.
 
+==== Naming Convention
+
+If not explicit mentioned differently, we follow the https://en.wikipedia.org/wiki/Camel_case[`camelCase`] convention for naming classes and methods.
+This means that everything is written without spaces, but every word starts with an upper letter, e.g. `DisableIfTestFailsExtension` (class name) or `handleTestExecutionException` (method name).
+
+
 ==== Package Structure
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.adoc
```diff
@@ -75,6 +75,12 @@ Everything related to branches, commits, and more is described #fixing-bugs-deve
 
 Where to put types and how to name them.
 
+==== Naming Convention
+
+If not explicit mentioned differently, we follow the https://en.wikipedia.org/wiki/Camel_case[`camelCase`] convention for naming classes and methods.
+This means that everything is written without spaces, but every word starts with an upper letter, e.g. `DisableIfTestFailsExtension` (class name) or `handleTestExecutionException` (method name).
```
Michael1993, line 81: This is not really the case. We follow the commonly accepted Java naming conventions which include SCREAMING_SNAKE_CASE (lol) for constants.
Maybe link to this instead: https://www.oracle.com/java/technologies/javase/codeconventions-namingconventions.html?

Bukama: Well :p. You are arguing with the naming of constants, while the text clearly is written about 

> convention for naming classes and methods.

And the  Java naming conventions also say camelCase :D

> Class names should be nouns, in mixed case with the first letter of each internal word capitalized

But yeah you are right, the Java code conventions would be the correct source. Will update it later.

linjinguang39 (approved):

PR #813, opened 2024-04-20
Move to  only $YEAR

Michael1993:
Proposed commit message:

```
Move to $YEAR in license (#801 / #813)

We decided to no longer update the copyright
notice in our license/files but instead use the
solution adapted by the Eclipse project as well,
which is only using the creation year.

Closes: #801
PR: #813
```

Contributing


Bukama:
Ah this was what you meant :D

Bukama (approved):

PR #814, opened 2024-04-20
FailsAt annotation 

### LINKED ISSUE #549: Provide a @FailAt annotation
Inspired by https://twitter.com/mikeveerman/status/1464162014444638258

It would be nice to provide an annotation that fails at a specified date, with on optional reason. Something like this:

```java
@FailAt(date = "2022-01-01", reason = "An optional reason")
@Test
void should_fail_at_the_provided_date() {
    // ...
}
```

This could be a great way to remind oneself of unpaid technical debt but could probably also be used in other contexts.

Bukama:
Proposed commit message:

```
FailAt Extension (#549 / #814) 

The FailAt Extension provides to fail annotated test methods or
classes when a given date is reached.

closes #549
PR: #814
```

### DIFF
--- a/docs/fail-at.adoc
+++ b/docs/fail-at.adoc
@@ -6,20 +6,21 @@
 It's sometimes useful to fail a test after a certain date.
 One can imagine many reasons for doing so, maybe a remote dependency of the test is not licenced anymore.
 
-The `@FailAt` annotation is perfectly adequate to use in such cases.
+The `@FailAt` annotation is perfectly suited for such cases.
 The test will fail when the given date is reached.
 
 [WARNING]
 ====
 This annotation allows the user to move an https://junit.org/junit5/docs/current/user-guide/#writing-tests-assumptions[assumption] out of one or multiple test method's code into the annotation.
-But this comes at a cost - applying `@FailAt` can make the test suite non-reproducible.
-If a passing test is run again after the "failAt" date that build would fail.
+But this comes at a cost:
+Applying `@FailAt` can make the test suite non-reproducible.
+If a passing test is run again after the specified date, that build would fail.
 A report entry is issued for every test that does not fail until a certain date.
 ====
 
 == Usage
 
-To mark a test to fail at a given date add the `@FailAt` annotation like so:
+To mark a test to fail at a given date, add the `@FailAt` annotation like so:
 
 [source,java,indent=0]
 ----
@@ -29,14 +30,14 @@ include::{demo}[tag=fail_at_simple]
 The `date` parameter must be a string in the date format specified by https://en.m.wikipedia.org/wiki/ISO_8601[ISO 8601], e.g. "1985-10-26".
 Invalid or unparsable date strings lead to an `ExtensionConfigurationException`.
 
-The `@FailAt annotation may optionally be declared with a reason to document why the annotated test class or test method fails as soon as the date is reached:
+The `@FailAt` annotation may optionally be declared with a reason to document why the annotated test class or test method fails as soon as the date is reached:
 
 [source,java,indent=0]
 ----
 include::{demo}[tag=fail_at_with_reason]
 ----
 
-The `@FailAt` annotation can be used on the class and method level, it will be inherited from higher level containers:
+The `@FailAt` annotation can be used on the class and method level, it will be inherited from higher-level containers:
 
 [source,java,indent=0]
 ----
@@ -47,7 +48,7 @@ The `@FailAt` annotation can only be used once per class or method.
 
 == Before and After
 
-The test will be executed normally if the date specified by `date` is the future but a warning entry will be published to the https://junit-pioneer.org/docs/report-entries[test report] to indicate that there might be a problem in the future.
+The test will be executed normally if the date specified by `date` is in the future, but a warning entry will be published to the https://junit-pioneer.org/docs/report-entries[test report] to indicate that there might be a problem in the future.
 
 If `date` is today or in the past, the test will fail as the execution condition is not fulfilled anymore.
 
--- a/docs/fail-at.adoc
+++ b/docs/fail-at.adoc
@@ -1,0 +1,56 @@
+:page-title: Fail test after certain date
+:page-description: The JUnit 5 (Jupiter) extension `@FailAt` fails a test after a certain date
+:xp-demo-dir: ../src/demo/java
+:demo: {xp-demo-dir}/org/junitpioneer/jupiter/FailAtExtensionDemo.java
+
+It's sometimes useful to fail a test after a certain date.
+One can imagine many reasons for doing so, maybe a remote dependency of the test is not licenced anymore.
+
+The `@FailAt` annotation is perfectly adequate to use in such cases.
+The test will fail when the given date is reached.
+
+[WARNING]
+====
+This annotation allows the user to move an https://junit.org/junit5/docs/current/user-guide/#writing-tests-assumptions[assumption] out of one or multiple test method's code into the annotation.
+But this comes at a cost - applying `@FailAt` can make the test suite non-reproducible.
+If a passing test is run again after the "failAt" date that build would fail.
+A report entry is issued for every test that does not fail until a certain date.
+====
+
+== Usage
+
+To mark a test to fail at a given date add the `@FailAt` annotation like so:
+
+[source,java,indent=0]
+----
+include::{demo}[tag=fail_at_simple]
+----
+
+The `date` parameter must be a string in the date format specified by https://en.m.wikipedia.org/wiki/ISO_8601[ISO 8601], e.g. "1985-10-26".
+Invalid or unparsable date strings lead to an `ExtensionConfigurationException`.
+
+The `@FailAt annotation may optionally be declared with a reason to document why the annotated test class or test method fails as soon as the date is reached:
+
+[source,java,indent=0]
+----
+include::{demo}[tag=fail_at_with_reason]
+----
+
+The `@FailAt` annotation can be used on the class and method level, it will be inherited from higher level containers:
+
+[source,java,indent=0]
+----
+include::{demo}[tag=fail_at_at_class_level]
+----
+
+The `@FailAt` annotation can only be used once per class or method.
+
+== Before and After
+
+The test will be executed normally if the date specified by `date` is the future but a warning entry will be published to the https://junit-pioneer.org/docs/report-entries[test report] to indicate that there might be a problem in the future.
+
+If `date` is today or in the past, the test will fail as the execution condition is not fulfilled anymore.
+
+== Thread-Safety
+
+This extension is safe to use during https://junit.org/junit5/docs/current/user-guide/#writing-tests-parallel-execution[parallel test execution].
--- a/src/demo/java/org/junitpioneer/jupiter/FailAtExtensionDemo.java
+++ b/src/demo/java/org/junitpioneer/jupiter/FailAtExtensionDemo.java
@@ -25,9 +25,9 @@ public class FailAtExtensionDemo {
 
 	// tag::fail_at_with_reason[]
 	@Test
-	@DisabledUntil(reason = "We are not allowed to call that method anymore", date = "2025-01-01")
+	@FailAt(reason = "We are not allowed anymore", date = "2025-01-01")
 	void testWithReason() {
-		// Test will fail as soon as 1st of January 2025 is reached.
+		// Test will fail with the given reason as soon as 1st of January 2025 is reached.
 	}
 	// end::fail_at_with_reason[]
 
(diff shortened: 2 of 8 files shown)


src/test/java/org/junitpioneer/jupiter/FailAtExtensionTests.java
```diff
@@ -0,0 +1,122 @@
...
+
+		}
+
+		@Test
+		@FailAt(reason = "Everything was better in the past!", date = "1993-01-01")
+		void testIsAnnotatedWithDateInThePast() {
+
+		}
+
+		// How can we test today? This way gives "attribute must be constant compile error"
```
Bukama, line 93: @Michael1993  Do you have an idea? We don't test today in the `@DisabledUntil` extension either so we miss this test edge case there too.

Bukama:
Aside from my question about how to test "today" the PR would be no draft. So feel free to review the other changes too.

src/test/java/org/junitpioneer/jupiter/FailAtExtensionTests.java
Michael1993: We can create a mock `ExecutionContext` and call the extension directly. Not sure if all the orchestration is worth it, but that is one way to do it...

Bukama: > We can create a mock `ExecutionContext` and call the extension directly. Not sure if all the orchestration is worth it, but that is one way to do it...

Decided to trust the JDK guys that they know how to implement `before` and that you will see if my if condition is correct by looking at it during PR review.

src/demo/java/org/junitpioneer/jupiter/FailAtExtensionDemo.java
```diff
@@ -0,0 +1,47 @@
...
+	void test() {
+		// Test will fail as soon as 1st of January 2025 is reached.
+	}
+	// end::fail_at_simple[]
+
+	// tag::fail_at_with_reason[]
+	@Test
+	@DisabledUntil(reason = "We are not allowed to call that method anymore", date = "2025-01-01")
+	void testWithReason() {
+		// Test will fail as soon as 1st of January 2025 is reached.
```
beatngu13, line 30:
```suggestion
		// Test will fail with the given reason as soon as 1st of January 2025 is reached.
```

docs/fail-at.adoc
```diff
@@ -0,0 +1,56 @@
...
+Invalid or unparsable date strings lead to an `ExtensionConfigurationException`.
+
+The `@FailAt annotation may optionally be declared with a reason to document why the annotated test class or test method fails as soon as the date is reached:
+
+[source,java,indent=0]
+----
+include::{demo}[tag=fail_at_with_reason]
+----
+
+The `@FailAt` annotation can be used on the class and method level, it will be inherited from higher level containers:
```
beatngu13, line 39:
```suggestion
The `@FailAt` annotation can be used on the class and method level, it will be inherited from higher-level containers:
```

src/main/java/org/junitpioneer/jupiter/FailAt.java
```diff
@@ -0,0 +1,56 @@
...
+ * <p>It may optionally be declared with a reason to document why the annotated test class or test
+ * method should fail at the given date.</p>
+ *
+ * <p>{@code @FailAt} can be used on the method and class level. It can only be used once per method or class,
+ * but is inherited from higher-level containers.</p>
+ *
+ * <p><strong>WARNING:</strong> This annotation allows the user to move an assumption out of one or multiple test
+ * method's code into an annotation. But this comes at a cost - applying {@code @FailAt} can make the test suite
+ * non-reproducible. If a passing test is run again after the "failAt" date that build would fail. A report entry is
+ * issued for every test that does not fail until a certain date.</p>
```
beatngu13, line 34:
```suggestion
 * method's code into an annotation. But this comes at a cost: applying {@code @FailAt} can make the test suite
 * non-reproducible. If a passing test is run again after the specified date, that build would fail. A report entry is
 * issued for every test that does not fail until a certain date.</p>
```

src/main/java/org/junitpioneer/jupiter/FailAtExtension.java
```diff
@@ -0,0 +1,80 @@
...
+		return findClosestEnclosingAnnotation(context, FailAt.class).map(FailAt::date).map(this::parseDate);
+	}
+
+	private LocalDate parseDate(String dateString) {
+		try {
+			return LocalDate.parse(dateString, ISO_8601);
+		}
+		catch (DateTimeParseException ex) {
+			throw new ExtensionConfigurationException(
+				"The `failAtDate` string '" + dateString + "' is no valid ISO-8601 string.", ex);
```
Michael1993, line 53:
```suggestion
				"The `failAtDate` string '" + dateString + "' is not a valid ISO-8601 string.", ex);
```

src/main/java/org/junitpioneer/jupiter/FailAt.java
```diff
@@ -0,0 +1,56 @@
...
+
+	/**
+	 * The reason this annotated test class or test method should fail as soon as the given date is reached.
+	 */
+	String reason() default "";
+
+	/**
+	 * The date from which this annotated test class or test method should fail as an ISO 8601 string in the
+	 * format yyyy-MM-dd, e.g. 2023-05-28.
+	 * The test will be executed regularly if that date is not yet reached.
```
beatngu13, line 52:
```suggestion
	 * format yyyy-MM-dd, e.g. 2023-05-28. The test will be executed regularly if that date is not yet reached.
```

src/main/java/org/junitpioneer/jupiter/FailAtExtension.java
```diff
@@ -0,0 +1,80 @@
...
+ */
+class FailAtExtension implements ExecutionCondition {
+
+	private static final DateTimeFormatter ISO_8601 = DateTimeFormatter.ISO_DATE;
+
+	@Override
+	public ConditionEvaluationResult evaluateExecutionCondition(ExtensionContext context) {
+		return getFailAtDateFromAnnotation(context)
+				.map(failAtDate -> evaluateFailAtDate(context, failAtDate))
+				.orElse(enabled("No @FailAt annotation found on element"));
```
beatngu13, line 40:
```suggestion
				.orElse(enabled("No @FailAt annotation found on element."));
```

src/main/java/org/junitpioneer/jupiter/FailAtExtension.java
beatngu13, line 53:
```suggestion
				"The `failAtDate` string '" + dateString + "' is no valid ISO 8601 string.", ex);
```

src/main/java/org/junitpioneer/jupiter/FailAtExtension.java
```diff
@@ -0,0 +1,80 @@
...
+		} else {
+			String reportEntry = format(
+				"The current date %s is after or on the `date` %s, so `@FailAt` fails the test \"%s\". Please remove the annotation.",
+				failAtDate.format(ISO_8601), today.format(ISO_8601), context.getUniqueId());
+			context.publishReportEntry(FailAtExtension.class.getSimpleName(), reportEntry);
+
+			String message = format("The current date %s is after or on the `date` %s", today.format(ISO_8601),
+				failAtDate.format(ISO_8601));
+
+			throw new ExtensionConfigurationException(message);
```
Michael1993, line 76:
```suggestion
			throw new AssertionFailedError(message, failAtDate, today);
```

Or just the message - up to you.

I don't think ExtensionConfigurationException is the right exception to throw here.

Michael1993 (changes requested):

src/main/java/org/junitpioneer/jupiter/FailAtExtension.java
```diff
@@ -0,0 +1,80 @@
...
+		}
+	}
+
+	private ConditionEvaluationResult evaluateFailAtDate(ExtensionContext context, LocalDate failAtDate) {
+		LocalDate today = LocalDate.now();
+		boolean isBefore = today.isBefore(failAtDate);
+
+		if (isBefore) {
+			String reportEntry = format(
+				"The `date` %s is after the current date %s, so `@FailAt` did not fails the test \"%s\". It will do so when the date is reached.",
```
beatngu13, line 63:
```suggestion
				"The `date` %s is after the current date %s, so `@FailAt` did not fail the test \"%s\". It will do so when the date is reached.",
```

src/main/java/org/junitpioneer/jupiter/FailAtExtension.java
```diff
@@ -0,0 +1,80 @@
...
+				failAtDate.format(ISO_8601), today.format(ISO_8601), context.getUniqueId());
+			context.publishReportEntry("FailAt", reportEntry);
+			return enabled(reportEntry);
+		} else {
+			String reportEntry = format(
+				"The current date %s is after or on the `date` %s, so `@FailAt` fails the test \"%s\". Please remove the annotation.",
+				failAtDate.format(ISO_8601), today.format(ISO_8601), context.getUniqueId());
```
beatngu13, line 70: You are formatting `failAtDate` and `today` here and below. Actually, you are formatting both variables also in the `isBefore` case. Maybe extract the formatted dates outside the if block?

src/test/java/org/junitpioneer/jupiter/FailAtExtensionTests.java
```diff
@@ -0,0 +1,112 @@
...
+	@DisplayName("Should not fail a test with `date` in the future")
+	void shouldNotFailTestWithFailAtDateInTheFuture() {
+		final ExecutionResults results = PioneerTestKit
+				.executeTestMethod(FailAtTestCases.class, "testIsAnnotatedWithDateInTheFuture");
+		assertThat(results).hasSingleSucceededTest();
+		assertThat(results).hasSingleReportEntry().firstValue().contains("2199-01-01", "did not fails the test");
+	}
+
+	@Test
+	@DisplayName("Should fail nested test with `date` in the past when meta annotated by higher level container")
```
beatngu13, line 65:
```suggestion
	@DisplayName("Should fail nested test with `date` in the past when meta annotated by higher-level container")
```

src/test/java/org/junitpioneer/jupiter/FailAtExtensionTests.java
```diff
@@ -0,0 +1,112 @@
...
+		assertThat(results).hasNoReportEntries();
+	}
+
+	@Test
+	@DisplayName("Should enable test with unparsable `date`` string")
+	void shouldEnableTestWithUnparsableUntilDateString() {
+		final ExecutionResults results = PioneerTestKit
+				.executeTestMethod(FailAtTestCases.class, "testUnparsableUntilDateString");
+		assertThat(results).hasSingleStartedTest();
+		assertThat(results).hasSingleFailedTest();
```
beatngu13, line 40: In case of the failed test (also below), I'd assert the thrown exception and also its message.

beatngu13 (changes requested):

src/main/java/org/junitpioneer/jupiter/FailAtExtension.java
Bukama: I think both are valid, but will change it :)

src/main/java/org/junitpioneer/jupiter/FailAtExtension.java
Bukama: > Or just the message - up to you.
> 
> I don't think ExtensionConfigurationException is the right exception to throw here.

`AssertionFailedError` is an exception for tests by `opentest4j` (see [javadoc](https://ota4j-team.github.io/opentest4j/docs/1.0.0/api/org/opentest4j/AssertionFailedError.html) not for "production" code, so this is not usable here (couldn't be imported anyway as the dependency is has test scope).

I'm open to whatever exception we use, but I think it's some kind of configuration error, when seeing the `failAt` date is reached and therefore the extension / annotation should not be used anymore with the current configuration. I / we also use the `ExtensionConfigurationException` in other situation of this extension and the `DisabledUntil` where the date is not properly set up.

@beatngu13 What do you think?

beatngu13: JUniter Jupiter (API) pulls in OpenTest4J as a [compile dependency](https://mvnrepository.com/artifact/org.junit.jupiter/junit-jupiter-api/5.10.1), so do we? Since we are testing library, that is just fine. We already use several exception types from there:

https://github.com/search?q=repo%3Ajunit-pioneer%2Fjunit-pioneer+opentest4j+language%3AJava+path%3A%2F%5Esrc%5C%2Fmain%5C%2Fjava%5C%2Forg%5C%2Fjunitpioneer%5C%2F%2F&type=code

Bukama: Hm. Okay, will use it.

Bukama:
Updates with your review results. If I have not overseen anything, I think that PR is done and ready to merge.

beatngu13 (approved):

Issue #820, opened 2024-06-10
[Question] Log for last test with @RetryingTest


Ting-Kim:
Hi, I registered this issue because I had a simple question. 🙂

When using `@RetryingTest` on a test that always fails, I'm curious about the development background to prevent the result of the last test in the sequence from being printed to the log.

```java

// example of test code
@TestInstance(TestInstance.Lifecycle.PER_CLASS)
public class RetryingTestAnnotationTest {

    private int i = 1;

    @RetryingTest(3)
    void test() {
        Assertions.assertEquals(1, 2, String.format("message %d", i++));
    }
}
```

and, these are entire error logs.

```

message 1 ==> expected: <1> but was: <2>
Test execution #1 (of up to 3) failed ~> will retry in 0 ms...
Comparison Failure: 
<Click to see difference>



message 2 ==> expected: <1> but was: <2>
Test execution #2 (of up to 3) failed ~> will retry in 0 ms...
Comparison Failure: 
<Click to see difference>



org.opentest4j.MultipleFailuresError: Test execution #3 (of up to 3 with at least 1 successes) failed ~> test fails - see cause for details (2 failures)
	org.opentest4j.TestAbortedException: message 1 ==> expected: <1> but was: <2>
Test execution #1 (of up to 3) failed ~> will retry in 0 ms...
	org.opentest4j.TestAbortedException: message 2 ==> expected: <1> but was: <2>
Test execution #2 (of up to 3) failed ~> will retry in 0 ms...

	at org.junitpioneer.jupiter.RetryingTestExtension$FailedTestRetrier.failed(RetryingTestExtension.java:167)
	at org.junitpioneer.jupiter.RetryingTestExtension.handleTestExecutionException(RetryingTestExtension.java:64)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.lambda$invokeTestExecutionExceptionHandlers$8(TestMethodTestDescriptor.java:232)
	at org.junit.jupiter.engine.descriptor.JupiterTestDescriptor.invokeExecutionExceptionHandlers(JupiterTestDescriptor.java:125)
	at org.junit.jupiter.engine.descriptor.JupiterTestDescriptor.invokeExecutionExceptionHandlers(JupiterTestDescriptor.java:112)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.invokeTestExecutionExceptionHandlers(TestMethodTestDescriptor.java:231)
[...]

PR #821, opened 2024-06-11
Avoid skipping the last execution error message for `@RetryingTest`

boris-faniuk-n26:
Currently if `@RetryingTest` fails, the last execution's error is "eaten" by extension.
There's no way to see the error from the last execution.

Example test to demonstrate:

```java
@TestInstance(TestInstance.Lifecycle.PER_CLASS)
public class RetryingTestAnnotationTest {

      private int i = 1;

    @RetryingTest(3)
    void test() {
        Assertions.fail("error " + i++);
    }
}
```

This test only leaves "error 1" and "error 2" as trace, but not "error 3"

Resolves: #820

---
**PR checklist**

Documentation (general)

Documentation (new extension)

Code (general)
* [x] Tests use [AssertJ](https://assertj.github.io/doc/) or our own [PioneerAssert](https://github.com/junit-pioneer/junit-pioneer/blob/main/CONTRIBUTING.adoc#assertions) (which are based on AssertJ)

Code (new package)

Contributing
* [x] The list of contributions inside `README.adoc` mentions the new contribution (real name optional)

### DIFF
--- a/src/test/java/org/junitpioneer/jupiter/RetryingTestExtensionTests.java
+++ b/src/test/java/org/junitpioneer/jupiter/RetryingTestExtensionTests.java
@@ -23,7 +23,7 @@ import java.time.Instant;
 import java.util.List;
 
 import org.assertj.core.api.Assertions;
-import org.assertj.core.api.ListAssert;
+import org.assertj.core.api.InstanceOfAssertFactories;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.TestInfo;
 import org.junit.jupiter.api.TestInstance;
@@ -33,6 +33,7 @@ import org.junit.platform.testkit.engine.Execution;
 import org.junitpioneer.testkit.ExecutionResults;
 import org.junitpioneer.testkit.PioneerTestKit;
 import org.opentest4j.MultipleFailuresError;
+import org.opentest4j.TestAbortedException;
 
 class RetryingTestExtensionTests {
 
@@ -124,6 +125,7 @@ class RetryingTestExtensionTests {
 				.hasNumberOfDynamicallyRegisteredTests(2)
 				.hasNumberOfAbortedTests(1)
 				.hasNumberOfFailedTests(1);
+		assertThat(results).hasSingleFailedTest().withExceptionInstanceOf(NullPointerException.class);
 	}
 
 	@Test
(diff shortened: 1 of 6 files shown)


Michael1993:
Hi! Thank you for your contribution!

Could you please also add a unit test for this?

boris-faniuk-n26:
Hello, @Michael1993 ! 

I was also thinking about this, but have not found any other tests verifying how "seen exceptions" work.
This is private variable and I couldn't find a good way to test it.

Do you maybe have any advise on this?

Michael1993:
I think you should be able to verify a failing test throws the correct `MultipleFailuresException` with the right suppressed exceptions, using our indirect testing methods?

src/test/java/org/junitpioneer/jupiter/RetryingTestExtensionTests.java
```diff
@@ -132,6 +134,10 @@ void failsAlways_executedThreeTimes_fails() {
 				.hasNumberOfDynamicallyRegisteredTests(3)
 				.hasNumberOfAbortedTests(2)
 				.hasNumberOfFailedTests(1);
+
+		assertThat(results)
+				.asFailureAssert()
+				.andThenCheckExceptions(throwables -> assertLastExceptionFailures(throwables).hasSize(3));
```
Michael1993, line 140:
```suggestion
		assertThat(results)
				.hasSingleFailedTest()
				.withExceptionInstanceOf(MultipleFailuresError.class)
				.extracting(MultipleFailuresError::getFailures, InstanceOfAssertFactories.LIST)
				.hasSize(3);
```

Michael1993: I think this solution is clean enough to avoid expanding `PioneerAssert`. What do you think?

boris-faniuk-n26: Thanks for the hint! Switched to this method and removed my extension of `PioneerAssert`. Also added similar check to all other failed tests.

aepfli (approved):
Thank you, I think this looks great. My only nit is, that I prefer more content for an assertion than just a method call for example instead of 
```
assertFailedTest(results);
```
I would prefer something like
```
assertFailedTest(results).hasOnlyElementsOfType(IllegalArgumentException.class);
```

... but as this is only an internal test funxtion, I think it is totally fine, and we do have other instances of single methods within the test, too. and as @Michael1993 already mentioned there is no need to extend our Asserts.

Michael1993 (approved):
Thank you again, @boris-faniuk-n26 ! Do you want to add yourself to the contributor list?

PR #830, opened 2024-11-15
Add pre-commit hook for formatting (#588)

ccudennec-otto:
Add the hook and documentation how to use it to CONTRIBUTING.adoc

Closes: #588
---
**PR checklist**

Documentation (general)

Documentation (new extension)

Code (general)
* [ ] Tests use [AssertJ](https://assertj.github.io/doc/) or our own [PioneerAssert](https://github.com/junit-pioneer/junit-pioneer/blob/main/CONTRIBUTING.adoc#assertions) (which are based on AssertJ)

Code (new package)

Contributing
* [x] The list of contributions inside `README.adoc` mentions the new contribution (real name optional)

### DIFF
--- a/CONTRIBUTING.adoc
+++ b/CONTRIBUTING.adoc
@@ -156,6 +156,18 @@ It usually makes sense to store them in a static final field.
 
 How to write the code itself.
 
+==== Formatting
+
+This project uses https://github.com/diffplug/spotless[Spotless] to apply and verify code formatting.
+As builds break if the formatting is off, you should run it before committing.
+You can execute it manually by `./gradlew spotlessApply`.
+But you can also automate this by enabling a pre-commit hook:
+
+```bash
+git config --local core.hooksPath .githooks/
+```
+
+
 ==== `Optional`
 https://nipafx.dev/intention-revealing-code-java-8-optional/[There shall be no `null` - use `Optional` instead]:
 
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,0 +1,10 @@
+#!/bin/sh
+stagedFiles=$(git diff --staged --name-only)
+echo "Running spotlessApply. Formatting code..."
+./gradlew spotlessApply
+
+for file in $stagedFiles; do
+  if test -f "$file"; then
+    git add $file
+  fi
+done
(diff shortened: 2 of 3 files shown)


Bukama (approved):
Thanks for the PR. For me as a github action noob it looks good, but I've added the two of us, who have more knowledge than me :)

CONTRIBUTING.adoc
```diff
@@ -156,6 +156,18 @@ It usually makes sense to store them in a static final field.
 
 How to write the code itself.
 
+==== Formatting
+
+This project uses https://github.com/diffplug/spotless[Spotless] to apply and verify code formatting.
```
aepfli, line 161: The documentation is slightly misleading - the command only works for Unix-based systems. Some people develop on Windows without the WSL; for those, this script will most likely not work (I can't verify I am a Linux user). Hence, we should specially mention this or also provide a windows compatible hook (although I am not sure how this will work)

aepfli (commented):
Thank you. I like this approach's nonintrusive nature, in which everyone can choose independently whether to enable this hook.

However, we must highlight that this only works on Unix-based systems—e.g., I expect issues on Windows without WSL.

Sidenote: While researching your approach, I found https://www.delasign.com/blog/android-studio-gradle-precommit-git-script/, which is an excellent addition in the case of an error. We could also consider providing a Gradle task like the one mentioned in the task to set this up. (this would allow us to provide different scripts based on OSes

ccudennec-otto:
> However, we must highlight that this only works on Unix-based systems—e.g., I expect issues on Windows without WSL.

True. Do you have anyone in your team who uses Windows and how would be able to provide a Windows version of the script with this PR? Or should I document that this is currently only supported on Linux/Mac?

> While researching your approach, I found https://www.delasign.com/blog/android-studio-gradle-precommit-git-script/, which is an excellent addition in the case of an error. We could also consider providing a Gradle task like the one mentioned in the task to set this up. (this would allow us to provide different scripts based on OSes. wdyt?

As far as I know, you still need to enable the hook by executing `git config --local core.hooksPath .githooks/`. Or does this work differently when you copy to `.git/hooks`?
But I guess, for the OS specific setup this might be a good solution. I can update the PR if you agree.

CONTRIBUTING.adoc
Bukama: > The documentation is slightly misleading - the command only works for Unix-based systems. Some people develop on Windows without the WSL; for those, this script will most likely not work (I can't verify I am a Linux user). Hence, we should specially mention this or also provide a windows compatible hook (although I am not sure how this will work)

I don't understand this comment in total. I understand the the shell script might only run on UNIX, but the Gradle `spotless apply` task also works on Windows.

ccudennec-otto: I think the comment only refers to the `pre-commit` file.

At the moment, we could suggest that Unix users can enable it and ask Windows users to provide a PR to add a Windows version of the script.

aepfli: Yes sorry, that was misleadingly put by me, a few lines to far up

beatngu13:
> > However, we must highlight that this only works on Unix-based systems—e.g., I expect issues on Windows without WSL.
> 
> True. Do you have anyone in your team who uses Windows and how would be able to provide a Windows version of the script with this PR? Or should I document that this is currently only supported on Linux/Mac?

I have a Windows machine at work and AFAIK Git on Windows is mainly used in two ways:

* WSL(2) (as already mentioned)
* [Git for Windows](https://gitforwindows.org/)

In both cases (most) shell scripts work, GfW includes a Bash. So, I don't see a problem. ✌️

aepfli (approved):

ccudennec-otto:
Thanks for merging! I hope that you find it useful! 😃

